PR #1, opened 2023-04-07
Create base codeql config

timothyfroehlich:
This was the default one that was generated by the "advanced setup" option. The "basic" option wasn't abailable.


iamralpht (approved):
Thanks for adding this -- will need to do something similar for the Rust components.

PR #3, opened 2023-04-28
Add presubmits for the SDK

timothyfroehlich:
I've brought over all of the main presubmits from Kokoro, with the exception of AAOS Unbundled related builds. Those will be handled in a separate CL later.

The presubmits have had caching enabled which should speed up things nicely.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -17,9 +17,9 @@ jobs:
         cache: "gradle"
 
 
-    - name: ktlintCheck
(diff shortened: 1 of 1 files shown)


rylin8 (approved):

iamralpht:
@timothyfroehlich In the future it would be good to squash on merge. This PR has a lot of wip commits that aren't especially useful in main (same for Umesh's current change); I think we got used to Gerrit's model which more-or-less forces squashing.

PR #6, opened 2023-05-01
Update the readme to reflect the published Widget and Plugin

timothyfroehlich:


iamralpht (approved):
LGTM, and nice to keep the instructions for when folks are extending the plugins themselves. @rylin8 also please take a look.

rylin8 (approved):

PR #7, opened 2023-05-01
Add support for injecting fake key events to trigger Figma interactions

rylin8:
Figma's interaction tools give us key input as a way to trigger an action. This commit adds a function to inject one of these key events, which consists of a key and a list of optional metakeys (shift, alt, control, meta). There are two ways to use this key injection function.

1. Call it directly with the DesignInjectKey() function.
2. Create a new function annotated with the @DesignKeyAction() annotation. This generates a function that calls the DesignInjectKey() function with the parameters specified in the annotation.

An example of this has been added to the Interaction validation document.

The way this works is that when a node with key reactions is rendered, we register a function with the node as a listener for the key events it is interested in. When the key injection happens, listeners are notified and their interactions are dispatched.

Issue https://github.com/google/automotive-design-compose/issues/8

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/DesignView.kt
+++ b/designcompose/src/main/java/com/android/designcompose/DesignView.kt
@@ -262,30 +262,39 @@ data class DesignKeyEvent(val key: Char, val metaKeys: List<DesignMetaKey>) {
     }
 }
 
+private data class KeyAction(
+    val interactionState: InteractionState,
+    val action: Action,
+    val targetInstanceId: String?,
+    val key: String?,
+    val undoInstanceId: String?
+)
+
 // Manager to handle key event injects and listeners of key events
 private object KeyInjectManager {
-    private val keyListenerMap: HashMap<DesignKeyEvent, HashSet<(DesignKeyEvent) -> Unit>> =
-        HashMap()
+    private val keyListenerMap: HashMap<DesignKeyEvent, HashSet<KeyAction>> = HashMap()
 
-    // Inject a key event and notify listeners of the key event
+    // Inject a key event and dispatch any interactions on listeners of the key event
     fun injectKey(key: Char, metaKeys: List<DesignMetaKey>) {
         val keyEvent = DesignKeyEvent(key, metaKeys)
         val listeners = keyListenerMap[keyEvent]
-        listeners?.forEach { it(keyEvent) }
+        listeners?.forEach {
+            it.interactionState.dispatch(it.action, it.targetInstanceId, it.key, it.undoInstanceId)
+        }
     }
 
     // Register a listener for a specific key event. This happens when a view with a key event
     // trigger is composed.
-    fun addListener(keyEvent: DesignKeyEvent, setTriggeredKeyEvent: (DesignKeyEvent) -> Unit) {
+    fun addListener(keyEvent: DesignKeyEvent, keyAction: KeyAction) {
         if (keyListenerMap[keyEvent].isNullOrEmpty()) keyListenerMap[keyEvent] = HashSet()
-        keyListenerMap[keyEvent]?.add(setTriggeredKeyEvent)
+        keyListenerMap[keyEvent]?.add(keyAction)
     }
 
     // Remove a listener for a specific key event. This happens when a view with a key event trigger
     // is removed from composition.
-    fun removeListener(keyEvent: DesignKeyEvent, setTriggeredKeyEvent: (DesignKeyEvent) -> Unit) {
+    fun removeListener(keyEvent: DesignKeyEvent, keyAction: KeyAction) {
         val listeners = keyListenerMap[keyEvent]
-        listeners?.remove(setTriggeredKeyEvent)
+        listeners?.remove(keyAction)
     }
 }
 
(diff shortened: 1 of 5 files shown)


iamralpht:
Created #8 for this work (copied from the internal bug)

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -435,6 +496,40 @@ internal fun DesignView(
             )
     }
 
+    // triggeredKeyEvent starts out as null and gets populated if any key events that we are
+    // interested in get triggered. We just need to check if it's not null, find the reaction that
+    // the event corresponds with, and execute it.
+    val (triggeredKeyEvent, setTriggeredKeyEvent) =
```
iamralpht, line 502: It seems a bit funny to do this via a value that gets set and triggers a re-compose to dispatch the side effect. What if the application calls two key injections fast enough that a re-compose doesn't happen until after the second (and the same node is listening for both)? I think that the first one will get squashed.

Does the design break if the `KeyInjectManager` listener directly calls `InteractionState.dispatch`? (I think somehow that ought to cause a re-compose, too, but I've forgotten how).

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/MainActivity.kt
```diff
@@ -437,13 +442,27 @@ fun VConstraintsTest() {
...
+    @DesignKeyAction(key = 'C', metaKeys = [DesignMetaKey.MetaMeta]) fun clickedC()
 }
 
 @Preview
 @Composable
 fun InteractionTest() {
-    InteractionTestDoc.MainFrame()
+    val rootView = LocalView.current
+    InteractionTestDoc.MainFrame(
+        onTapKeyButtonB = { InteractionTestDoc.clickedB() },
```
iamralpht, line 463: Yep, this is about as nice as we can make this without being able to register custom events for the Figma UI to show...

If we have multiple instances of a component, then I think they're all sharing the same `InteractionState` instance, so that means that the generated methods can be static, but it also means that we can't have two independent instances of a component. This isn't a new issue, and it hasn't hurt us much before (does it break pressed state for list items?).

If we want to fix this global state issue, then we'll need to pass in `InteractionState` (or some "context"/"identity" object that just contains a unique key) as an argument to the component and as an argument to the generated key action functions; maybe make a follow-up ticket on this?

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -234,6 +235,65 @@ internal object DebugNodeManager {
...
+    // Remove a listener for a specific key event. This happens when a view with a key event trigger
+    // is removed from composition.
+    fun removeListener(keyEvent: DesignKeyEvent, setTriggeredKeyEvent: (DesignKeyEvent) -> Unit) {
+        val listeners = keyListenerMap[keyEvent]
+        listeners?.remove(setTriggeredKeyEvent)
+    }
+}
+
+// Public function to inject a key event
+fun DesignInjectKey(key: Char, metaKeys: List<DesignMetaKey>) {
```
iamralpht, line 293: I'm also wondering if we want to expose this API. I think we're much better off if the connection between actions and the "keys" is explicit and can be in the JSON doc. If it's implicit like this, then we can't surface the events to the designer easily (because it could be any events coming in, not just the ones in the annotation).

Are you OK with not adding this to the API for now, and we'll revisit if we find out that we need it?

rylin8: I'm ok with it being internal, but this is called by the generated code, which so far calls public functions in DesignCompose.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
rylin8: I simplified this by registering the listener with the dispatch data, and then having `KeyInjectManager` call the `dispatch()` function on all listeners directly. This worked, and I believe it's because `InteractionState` changes state internally, triggering a recompose.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -331,7 +401,7 @@ internal fun DesignView(
                         currentTimeout = (reaction.trigger as Trigger.AfterTimeout).timeout
                     }
                 }
-            // XXX: keypress
+                is Trigger.OnKeyDown -> onKeyReactions.add(reaction)
```
iamralpht, line 404: hurray, one less XXX!

iamralpht (approved):
The latest iteration addresses the "two event injections to the same DesignView without a recomposition in between" issue, and you're correct that the InteractionState mutation triggers a re-compose all by itself. I think this change is good to merge.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
iamralpht: Oh, right, at some point we might want to name things like that differently (public symbol but should only be called by generated code, so we don't consider it a supported API), but since this is in-keeping with the current approach we're good.

PR #11, opened 2023-05-01
Replace the existing Rust Android Gradle Plugin with our own

timothyfroehlich:
NOTE: This requires a rust upgrade to 1.68. Run `rustup update`

The new plugin includes a few new things vs the previous one:
- Stable interface with Android Gradle Plugin
The plugin is properly integrated with AGP's build lifecycle and
uses Gradle's and AGP's current APIs for creating task dependencies
- Configured for Gradle Build Caching
Gradle is aware of whether the sources for the build have changed and
whether it needs to run the task and trigger Cargo
- Handles Debug and Release builds properly
Creates individual tasks for compiling the Rust code with and without
the --release flag, using the right build profile for the right
Gradle artifact
- Stable with AGP 8+
The previous plugin, starting with AGP 7.4, would sometimes
build an AAR that was missing the JNI libraries.
- Removes the workarounds that were required to compile Rust against
the Android NDK
See https://blog.rust-lang.org/2023/01/09/android-ndk-update-r25.html


rylin8:
Tried rebasing this on top of main and got conflicts, but as is I was able to build successfully.

timothyfroehlich:
> Tried rebasing this on top of main and got conflicts, but as is I was able to build successfully.

I've been rebasing it fairly often but it's been in review for a week now and it touches a couple places that I've been trying to improve and submit other PRs for.

timothyfroehlich:
Also, personally I've swiitched back to merging main back into the branch instead of rebasing (I need to stop using the word "rebase" when I mean "merged main in") the whole branch on top of main. It leads to much fewer conflicts. Github says there's no merge conflicts so you shouldn't need to do any kind of rebasing.

README.md
```diff
@@ -92,14 +91,6 @@ The DesignCompose library includes a Rust JNI library, which requires Rust build
 ./install-rust-toolchains.sh
 ```
 
-### Python
```
iamralpht, line 95: super nice to reduce build dependencies!

rylin8 (approved):

PR #12, opened 2023-05-02
Don't display Toasts due to crash in AAOS on secondary displays.

urpatil:
AAOS bug has been filed.

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/DocServer.kt
+++ b/designcompose/src/main/java/com/android/designcompose/DocServer.kt
@@ -66,6 +66,10 @@ internal class LiveDocSubscriptions(
 
 object DesignSettings {
     internal var liveUpdatesEnabled = false
+    // Toast.makeText causes a crash in AAOS on secondary displays with a
+    // "display must not be null" error.  This flag is used to disable
+    // Toasts in affected apps till that issue is resolved.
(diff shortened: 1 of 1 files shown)


designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -135,8 +135,12 @@ object DesignSettings {
 
     fun showMessageInToast(msg: String, duration: Int) {
         Log.i(TAG, "Raising toast: $msg")
+        // Toast.makeText dies in AAOS with "display must not be null" on secondary displays.
+        // Just log the message till that issue is resolved.
```
timothyfroehlich, line 139: Is there a system flag that we could put this behind? Like, is there a flag that marks the system as AAOS instead of regular Android?

Also, is issue present on all versions of AAOS and will previous versions be patched?

urpatil: One option is to add a DesignSettings flag - disableToasts (default false) and is set to true when called from apps that use a secondary screen in AAOS (like the cluster).

I have seen this crash is in S+.

urpatil (commented):
Added the flag.

rylin8 (approved):
Please squash when merging

PR #13, opened 2023-05-02
Make the Tutorial app a standalone build

timothyfroehlich:


reference-apps/tutorial/app/build.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+
+    buildTypes {
+        getByName("release") {
+            isMinifyEnabled = true
+            proguardFiles(
+                getDefaultProguardFile("proguard-android-optimize.txt"),
+                "proguard-rules.pro"
+            )
+        }
+    }
```
timothyfroehlich, line 53: I cleaned up some old config that we weren't using any more.

dev-scripts/format-all.sh
```diff
@@ -26,6 +26,7 @@ if ! which "$GOPATH"/addlicense; then
     exit 1
 fi
 
+"$GOPATH"/addlicense -c "Google LLC" -l apache "$GIT_ROOT"
```
timothyfroehlich, line 29: Moving this earlier so that it still adds the license even if a later format command fails (like if I don't have an AAOS Unbundled checkout handy)

.github/workflows/main.yml
```diff
@@ -123,6 +123,34 @@ jobs:
...
+      with:
+        distribution: "temurin"
+        java-version: "11"
+    - uses: actions/download-artifact@v3
+      with:
+        name: designcompose-m2-repo-zip
+
+    - run: unzip designcompose_m2repo.zip
+
+    - name: "Set environmental variables"
```
iamralpht, line 142: super minor nit, should be "environment".

iamralpht (approved):

timothyfroehlich:
Will resolve the Mac build and merge

PR #16, opened 2023-05-02
Call design doc ready callback with the ID of the Figma document

rylin8:
Issue https://github.com/google/automotive-design-compose/issues/14


urpatil (approved):
LGTM!

timothyfroehlich (approved):
Seems like it'll be nice to do that log message by default, but this is good as is.

iamralpht (approved):
Looks good, and that this will re-issue the notification when the doc is switched, too. So this change supports keeping other clients in sync with the current doc.

PR #17, opened 2023-05-02
Set GitHub actions to use specific version refs

timothyfroehlich:
Recommended by Google OSS and the Scorecard action


.github/workflows/main.yml
```diff
@@ -29,31 +29,34 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 
+permissions:  # added using https://github.com/step-security/secure-repo
```
timothyfroehlich, line 32: This bit restricts the permissions of the action to just reading the PR.

rylin8 (approved):

PR #19, opened 2023-05-02
Remove Standalone Apps

timothyfroehlich:
PR #13 makes the Tutorial app into it's own project. I used a new method that makes things a bit cleaner and avoids some (but not all) of the symlinking.

The standalone Validation app basically existed to make sure the published repo worked, but the new Tutorial app handles that better. So I'm removing the standalone version of the Validation app.

Meanwhile, MediaCompose  (which is now the only app in the AAOS Unbundled project) works the same way now, basically. So I can remove it's "Standalone" version.


iamralpht:
Do we still have a buildable validation as part of the designcompose target?

timothyfroehlich:
Yep, this was just a separate project .

iamralpht (approved):

PR #20, opened 2023-05-03
Make sure everything builds on both Linux and MacOS

timothyfroehlich:
I'm adding the default MacOS target to our CI builds to make sure everything builds fine. I'm only adding it for the basic builds of each language. This is important for the NDK builds especially

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,7 +54,10 @@ jobs:
         run: cargo-fmt --all --check
 
   rust-test:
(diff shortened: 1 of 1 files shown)


timothyfroehlich:
> Looks good, and important to have coverage on macOS. Do you know why the sdk-build status hasn't been reported?

It was looking for the original version of the check, but now there's Ubuntu and MacOS versions

iamralpht (approved):

PR #24, opened 2023-05-05
Upgrade to Gradle 8.1.1 and AGP 8.0.1

timothyfroehlich:
Depends on #21

PR #25, opened 2023-05-05
Upgrade Kotlin to 1.8.21 and all other Android dependencies

timothyfroehlich:
Depends on #24


designcompose/build.gradle.kts
```diff
@@ -74,11 +75,8 @@ dependencies {
     api(libs.androidx.lifecycle.runtime)
     api(libs.androidx.lifecycle.livedata.core)
     implementation(libs.accompanist.flowlayout)
-    implementation(libs.grpc.auth)
-    implementation(libs.grpc.stub)
-    implementation(libs.grpc.okhttp)
-    implementation(libs.grpc.protobufLite)
```
rylin8, line 80: Do we not need these grpc dependencies anymore, but still need grpc.stub?

rylin8:
Why do we update versions so frequently?

designcompose/build.gradle.kts
timothyfroehlich: Yes. We haven't fully cleaned up the code that used GRPC yet. I was removing those dependencies, then realized that and added Stub back.

rylin8 (approved):

PR #26, opened 2023-05-05
Enable the configuration cache

timothyfroehlich:
This will speed up rebuilds of the Gradle project.

Resolving the following problems that prevent the cache from working

- Remove OSS Licenses plugin: It's not compatible with the cache and we weren't using it - we're not shipping full apps.

- Remove usages of project from the Cargo task

Stacked on #32


timothyfroehlich:
I'm having some issues with how the ktfmt formatter is working, so I'm pulling this back to draft for now.

timothyfroehlich:
Should be good to go, though I'll need either #33 or #32 to merge so that I can run presubmit tests.

rylin8 (approved):

Issue #27, opened 2023-05-08
Investigate possible caching issue with Rust serialization


timothyfroehlich:
I was able to reproduce the issue on main

1. Build and run Validation, select the Dials and Gauges panel, confirm that it renders
1.  Modify crates/figma_import/src/toolkit_style.rs, add a new `bool` field to pub struct ProgressBarMeterData {}
2. Rebuild and run Validation, select Dials and Guages, it should encounter a deserialization error
3. Clean and rebuild the app with --no-build-cache, the Dials and Guages panel should render fine

It looks like the generated Java code includes the new boolean field at least.

Will reproduce on top of #24 and figure out the problem.

PR #29, opened 2023-05-08
Removing build results tracking from the CargoBuild Task 

timothyfroehlich:
Stacked on #24

Attempting to have it be cacheable and attempting to determine if we can skip running Cargo was pretty naive. Trying to get it right will be super error prone.

Cargo is already pretty quick at returning if it doesn't need to build. Gradle adds to that time, but once we have configuration caching (#26) enabled, the tasks will run simultaneously and won't really add to the build time.

Fixes Investigate possible caching issue with Rust serialization #27


timothyfroehlich:
Forgot to mention, I added a basic test to Validation to help me debug this. Not the most useful but it's something.

rylin8 (approved):

Issue #28, opened 2023-05-08
Tutorial App may be using incorrect DesignCompose version


timothyfroehlich:
When building in Android Studio the Tutorial app may be using the gMaven published version instead of DesignCompose's source. I think I saw it happen on one build and need to confirm that that's what happened.

PR #30, opened 2023-05-08
Fix Tutorial App may be using incorrect DesignCompose version #28

timothyfroehlich:
Re-add it as a regular included build. Technically this could cause issues if you tried to build both the regular designcompose project and tutorial app project at the same time, but I don't know why you would. Each project is meant for different audiences.


timothyfroehlich:
Wow, that commit history is way longer than it needs to be... I've got too many branches on branches on branches going. Fortunately the squash and merge will clean all that up.

iamralpht (approved):

iamralpht:
> Wow, that commit history is way longer than it needs to be... I've got too many branches on branches on branches going. Fortunately the squash and merge will clean all that up.

I still like rebase, but I've had enough rebases go horribly wrong that I know how to back up and try a different strategy ;)

PR #2, opened 2023-04-28
Rollup of changes made since the initial commit

timothyfroehlich:
Beginning the transition over to GitHub, migrating the changes over from the old repo

PR #31, opened 2023-05-08
Support progress bars where an element just moves along the bar

### LINKED ISSUE #58: Support progress bar indicator mode
Support another type of progress bar where a visual element, like a tick mark, just moves along the bar instead of growing as a normal progress bar does.

rylin8:
Add a new plugin option to the dials and gauges plugin to move a frame or rect along a progress bar instead of adjusting the size like a normal progress bar. Write this new boolean into the plugin data, parse it, and render it in FrameRender.

Fix #58

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/FrameRender.kt
+++ b/designcompose/src/main/java/com/android/designcompose/FrameRender.kt
@@ -38,9 +38,13 @@ import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.unit.IntSize
 import androidx.core.graphics.minus
 import androidx.core.graphics.plus
+import com.android.designcompose.serdegen.ArcMeterData
 import com.android.designcompose.serdegen.BoxShadow
 import com.android.designcompose.serdegen.MeterData
 import com.android.designcompose.serdegen.Overflow
+import com.android.designcompose.serdegen.ProgressBarMeterData
+import com.android.designcompose.serdegen.ProgressMarkerMeterData
+import com.android.designcompose.serdegen.RotationMeterData
 import com.android.designcompose.serdegen.StrokeAlign
 import com.android.designcompose.serdegen.StrokeCap
 import com.android.designcompose.serdegen.ViewShape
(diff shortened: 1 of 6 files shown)


support-figma/extended-layout-plugin/ui.html
```diff
@@ -156,6 +156,12 @@
           <label for="barVerticalSwitch" class="switch__label" style="font-weight:normal">Vertical</label>
         </div>
       </div>
+      <div style="display: flex">
+        <div class="switch">
+          <input id="barMoveOnlySwitch" type="checkbox" class="switch__toggle">
+          <label for="barMoveOnlySwitch" class="switch__label" style="font-weight:normal">Move Only (Scrubber)</label>
```
iamralpht, line 162: "Move Only" sounds a bit funny, but "Translate" sounds quite technical. I guess we can figure out the right way to name this once Aubrey gets into designing the plugin panel (but then we will need to update JS, Rust, and Kotlin...).

support-figma/extended-layout-plugin/code.ts
```diff
@@ -741,10 +741,19 @@ else if (figma.command === "move-plugin-data") {
     rotation = (node as FrameNode).rotation;
 
     if (node.type == "FRAME" || node.type == "RECTANGLE") {
-      if (meterData && meterData.progressBarData && meterData.progressBarData.vertical)
-        progress = node.height;
-      else
-        progress = node.width;
+      if (meterData && meterData.progressBarData) {
```
iamralpht, line 744: I wonder if there's some destructuring sugar like `if let Some(ProgressBarData::Vertical { end } = ...` in TypeScript so these can be less nesty. (Maybe ["Default Values" in this guide?](https://medium.com/@satyamsaluja04/destructuring-in-javascript-aliases-nested-objects-and-default-values-8fc1b378cf6d)). Maybe that makes it less readable?

iamralpht:
In your test doc, it would be nice to try out blend modes to carve a space around the "move only" scrubber thing. Can you see if that works?

[This page in the Compose documentation](https://developer.android.com/jetpack/compose/graphics/draw/modifiers) uses the "Clear" blend mode to make a space around a circle, and the parent object uses the "Normal" blend mode instead of the "Passthrough" blend mode. I think this kind of "cut out around a triangle" is fairly common for gas gauges.

support-figma/extended-layout-plugin/ui.html
rylin8: I really struggled to figure out what to call this. I was thinking of this as a tick mark, so maybe `tickMode` but that didn't sound right.

rylin8:
> In your test doc, it would be nice to try out blend modes to carve a space around the "move only" scrubber thing. Can you see if that works?
> 
> [This page in the Compose documentation](https://developer.android.com/jetpack/compose/graphics/draw/modifiers) uses the "Clear" blend mode to make a space around a circle, and the parent object uses the "Normal" blend mode instead of the "Passthrough" blend mode. I think this kind of "cut out around a triangle" is fairly common for gas gauges.

Hmm I am not familiar enough with Figma to know how to do this. Looks like you have to change the compositing strategy in Android to do this... don't know how to do that in Figma either. Maybe you can do a quick demo of it for me? My doc is https://www.figma.com/file/lZj6E9GtIQQE4HNLpzgETw/Dials-and-Gauges?node-id=72%3A32&t=3FvKMfq1qriFaJLP-1

iamralpht:
> Hmm I am not familiar enough with Figma to know how to do this. Looks like you have to change the compositing strategy in Android to do this... don't know how to do that in Figma either. Maybe you can do a quick demo of it for me? My doc is https://www.figma.com/file/lZj6E9GtIQQE4HNLpzgETw/Dials-and-Gauges?node-id=72%3A32&t=3FvKMfq1qriFaJLP-1

Ok, I tried with `#ProgressFrameMoveOnlyIndicator`, but Figma doesn't have a "Clear" blend mode, and without that I can't make it work. I think we will need to implement masking (or native boolean ops!! yikes!!) to do this correctly.

iamralpht:
@timothyfroehlich I kicked the CI job for MediaCompose again on this -- it failed to checkout the ktfmt prebuilt (missing commit? but might just be a weird timeout error?)

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -370,12 +370,38 @@ internal fun ContentDrawScope.render(
                                 progressBarData.discrete,
                                 progressBarData.discreteValue
                             )
-                        if (progressBarData.vertical) {
-                            frameSize = Size(size.width, progress)
-                            val topOffset = style.top.pointsAsDp().value
-                            overrideTransform = androidx.compose.ui.graphics.Matrix()
-                            overrideTransform.translate(0F, -topOffset + end - progress)
-                        } else frameSize = Size(progress, size.height)
+                        if (progressBarData.moveOnly) {
```
timothyfroehlich, line 373: The code for if and else are very simmilar, and it's indented seven times. I have a few suggestions:

#1, refactor the code in the if and else blocks into a function.
#2, move the contents of the whole "is MeterData.progressBarData" block into a separate function. 
#3, can you add a unit test for the progress bar? Not a visual test, just a test of the math.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -370,12 +370,38 @@ internal fun ContentDrawScope.render(
...
+                                        ?: androidx.compose.ui.graphics.Matrix()
+                                val decomposed = style.transform.decompose(density)
+                                val topOffset = end + start - style.top.pointsAsDp().value
+                                overrideTransform.translate(
+                                    0F,
+                                    topOffset - progress + decomposed.translateY
+                                )
+                            } else {
+                                overrideTransform =
+                                    style.transform.asComposeTransform(density)
```
timothyfroehlich, line 386: Why isn't "start" and "end" used in the else block?

support-figma/extended-layout-plugin/ui.html
timothyfroehlich: What about reframing it from a boolean to an option named "progressIndicatorMode" or "indicatorMode", with the options being "expanding" or "positionOnly".

Or allow both options to exist at the same time? Have an "expanding" element and a "position" element, and just render both if frames are given for both.

timothyfroehlich (changes requested):

support-figma/extended-layout-plugin/ui.html
iamralpht: Heh, my feedback on the first round was to only allow one option at a time because it was hard for me to think of valid use cases for combinations, and because it multiplies the validation effort.

Fundamentally, we're either "transforming" (rotate or translate) or "resizing" (along an arc, or along a rect/frame's axis); that's a very low level description, but perhaps that's OK given all of the different ways of composing the drawing capabilities with this feature result in many use cases (progress, dials, gauges, indicators...) getting tackled.

@miriamvu wdyt?

rylin8: It would be harder to support having both since this is on one frame or one rectangle vector.

support-figma/extended-layout-plugin/code.ts
rylin8: Yea I can make this a little better, will clean up.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: They are already taken into account with the calculation of `progress`. The vertical format needs additional calculations because positive y goes down, but we want the progress bar to go up.

support-figma/extended-layout-plugin/ui.html
miriamvu: I agree that we should allow one option at a time to keep things simpler and easier to understand/use.  As for terminology, it seems that the there are three possible operations:  translation, rotation, and scaling (Figma uses this term too for expanding and shrinking an item:  https://help.figma.com/hc/en-us/articles/360040451453-Resize-layers-with-the-scale-tool).   Hope this helps!

rylin8:
After further testing with various matrix transformations on progress bars, I found some bugs that I am in the process of fixing. The "vertical" option complicates things as I need a separate and more complicated code path for them, so I'm thinking of removing this option completely. A vertical progress bar should still be able to be created by simply rotating a horizontal progress bar.

rylin8:
After testing this with some more interesting progress bars with various transforms applied to them, I found some problems. I spent a lot of time trying to get this working for all transformations but have unable to so far. Since I think it will be rare to do these transformations, the current implementation mostly supports skew and rotation transformations, but some odd configurations won't work properly and will take a lot more work to get it right.

The latest changes are functional, but I have a couple cleanup changes incoming still.

rylin8:
Discussed with @miriamvu and am considering making the "indicator mode" a separate section, like below. Not sure what to call it still -- how about "Progress Marker"?
Screenshot 2023-05-16 at 4 27 21 PM

rylin8:
> Discussed with @miriamvu and am considering making the "indicator mode" a separate section, like below. Not sure what to call it still -- how about "Progress Marker"? Screenshot 2023-05-16 at 4 27 21 PM

Updated this PR with this change.

rylin8:
Rebased and removed `customProgressBar` since it is no longer used.

rylin8:
Fixed a few small issues and fixed typescript warnings. Please review!

rylin8:
> Fixed a few small issues and fixed typescript warnings. Please review!

Ping

support-figma/extended-layout-plugin/code.ts
```diff
@@ -707,12 +718,93 @@ else if (figma.command === "move-plugin-data") {
...
+    result.m11 = row0x
+    result.m12 = row0y
+    result.m21 = row1x
+    result.m22 = row1y
+
+    // Convert into degrees because our rotation functions expect it.
+    result.angle = radiansToDegrees(result.angle);
+
+		// calculate delta transform point
+		const px = deltaTransformPoint(t, { x: 0, y: 1 });
```
iamralpht, line 792: funny indent here -- wonder if we can get eslint going on the plugin sources at some point...

iamralpht (approved):
nice to fix the indent, but the logic is clear, and there's less repetition now (I'm still noodling on how we can share more logic between plugins/widgets and the DC runtime implementations...).

support-figma/extended-layout-plugin/code.ts
rylin8: Oops those got inserted as tabs, probably from copying some code. I'll fix in another commit as I have another change incoming soon.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: This has been heavily refactored.

PR #48, opened 2023-05-10
Propagate better errors from figma_import JNI

timothyfroehlich:
- I saw the JNI crate using the "thiserror" to help with error handling, so I added it here to make it easier to display errors that we aren't specifically handling

- Most of the errors from Rust should be handled by doing more than just printing an error, so I'm setting up a mapping between Rust errors and Exceptions that'll be thrown. I think that interpreting the errors into simple exceptions should be handled on the Rust side, partly to reduce the amount of data that needs to be copied over.

- I cleaned up the error display for Feedback a bit, making it more concise.

- I've also updated the rust-JNI library to 0.21, which improves safety especially related to reference lifetimes.

### DIFF
--- a/designcompose/src/androidTest/kotlin/com/android/designcompose/JniNoNetworkTests.kt
+++ b/designcompose/src/androidTest/kotlin/com/android/designcompose/JniNoNetworkTests.kt
@@ -16,6 +16,8 @@
 
 package com.android.designcompose
 
+import android.net.ConnectivityManager
+import androidx.core.content.ContextCompat.getSystemService
 import androidx.test.platform.app.InstrumentationRegistry
 import java.io.IOException
 import java.net.ConnectException
@@ -27,7 +29,6 @@ import org.hamcrest.core.IsInstanceOf
 import org.junit.After
 import org.junit.Before
 import org.junit.Test
-import java.lang.Thread.sleep
 
 class JniNoNetworkTests {
     @Before
@@ -36,7 +37,15 @@ class JniNoNetworkTests {
             .uiAutomation
             .executeShellCommand("cmd connectivity airplane-mode enable")
         // Add a sleep to make sure the connection goes down.
-        sleep(250)
+        val connectivityManager =
+            getSystemService(
+                InstrumentationRegistry.getInstrumentation().context,
+                ConnectivityManager::class.java
+            )
+        println(connectivityManager?.isDefaultNetworkActive)
+        println("Hii")
+
+        wait
     }
     @Test
     fun networkFailure() {


timothyfroehlich:
I could use feedback on what I have so far, and especially the Rust syntax that I'm using.

crates/figma_import/Cargo.toml
```diff
@@ -19,6 +19,7 @@ fetch = ["clap"]
 "serde_json" = "1.0"
 "serde_bytes" = "0.11"
 "bincode" = "1.3"
+"thiserror" = "1.0"
```
iamralpht, line 22: This is good -- `thiserror` and `anyhow` are the crates for syntax sugaring error handling in Rust that are recommended by Core Android.

crates/figma_import/src/error.rs
```diff
@@ -12,55 +12,19 @@
...
-        }
-    }
-}
-impl Display for Error {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{:?}", self)
-    }
+    #[error("IO Error")]
+    IoError(#[from] std::io::Error),
+    #[error("HTTP Error")]
```
iamralpht, line 22: Maybe "Network Error"?

designcompose/src/androidTest/kotlin/com/android/designcompose/JniFetchTests.kt
```diff
@@ -38,9 +38,9 @@ class JniFetchTests {
         try {
             LiveUpdateJni.jniFetchDoc("DummyDocId", requestJson)
             Assert.fail("Should have thrown an exception")
-        } catch (exception: Exception) {
+        } catch (exception: AccessDeniedException) {
```
rylin8, line 41: What happens if a different exception is thrown here? Will the test fail?

designcompose/src/main/java/com/android/designcompose/LiveUpdateExceptions.kt
```diff
@@ -0,0 +1,3 @@
+package com.android.designcompose
+
+class AccessDeniedException: RuntimeException()
```
rylin8, line 3: What does this do?

crates/live_update/src/jni.rs
```diff
@@ -15,60 +15,73 @@
...
-fn jni_fetch_doc(
-    env: JNIEnv,
+fn jni_fetch_doc<'local>(
+    mut env: JNIEnv<'local>,
     _class: JClass,
     jdoc_id: JString,
     jrequest_json: JString,
-) -> jbyteArray {
-    let doc_id: String = match env.get_string(jdoc_id) {
+) -> JByteArray<'local> {
```
rylin8, line 39: Is `jbyteArray` different than `JByteArray`?

crates/live_update/src/jni.rs
```diff
@@ -15,60 +15,73 @@
...
+use crate::error_map::map_err_to_exception;
+
+fn throw_basic_exception(env: &mut JNIEnv, msg: String) {
     //An error occurring while trying to throw an exception shouldn't happen, but let's at least panic with a decent error message
     env.throw(msg).expect("Error while trying to throw Exception");
 }
 
-fn jni_fetch_doc(
-    env: JNIEnv,
+fn jni_fetch_doc<'local>(
```
rylin8, line 34: What changed such that you need this lifetime specifier now? I don't get how that works since this is called from Kotlin isn't it?

designcompose/src/androidTest/kotlin/com/android/designcompose/JniFetchTests.kt
timothyfroehlich: Not sure, I should find the right "assert that exception occurred" assertion.

designcompose/src/main/java/com/android/designcompose/LiveUpdateExceptions.kt
timothyfroehlich: Just declares an exception class so we can catch and handle specific ones.

crates/live_update/src/jni.rs
timothyfroehlich: Yes, it includes lifetime data

crates/live_update/src/jni.rs
timothyfroehlich: Part of the upgrade to jni-rs 0.21: https://github.com/jni-rs/jni-rs/blob/master/docs/0.21-MIGRATION.md

"Most of these changes are needed to ensure that all local references (JObject and the like) have the correct lifetime and can't be used after they're deleted, which would cause undefined behavior. See https://github.com/jni-rs/jni-rs/issues/392 for a discussion of the problem these changes solve."

designcompose/src/androidTest/kotlin/com/android/designcompose/JniFetchTests.kt
iamralpht: I think that the test will fail because any other kind of exception won't be caught. Tim tells me that Kotlin makes all exceptions unchecked.

designcompose/src/androidTest/kotlin/com/android/designcompose/JniNoNetworkTests.kt
```diff
@@ -0,0 +1,64 @@
...
+        InstrumentationRegistry.getInstrumentation()
+            .uiAutomation
+            .executeShellCommand("svc data disable")
+    }
+    @Test
+    fun networkFailure() {
+        val exception =
+            assertFailsWith<IOException> {
+                LiveUpdateJni.jniFetchDoc("DummyDocId", dummyFigmaTokenJson)
+            }
```
rylin8, line 46: Do you need to check for an auth token in setup() I tried running this test and got the following:

java.lang.AssertionError: Expected an exception of class java.io.IOException to be thrown, but was com.android.designcompose.AccessDeniedException: Invalid Authentication Token
at kotlin.test.AssertionsKt__AssertionsImplKt.AssertionErrorWithCause(AssertionsImpl.kt:117)

timothyfroehlich: Hm, no, this one is specifically meant to test a network connection error to Figma. The setup should turn off networking on your emulator, and the cleanup should turn it back on. Could you try it again?

rylin8 (approved):

PR #64, opened 2023-05-17
Add Figma Doc ID to the SerializableFigmaDoc struct.

srahman80:
This change will facilitate loading locally saved Figma docs for the renderer, or anything using the renderer library.

### DIFF
--- a/crates/figma_import/src/bin/fetch.rs
+++ b/crates/figma_import/src/bin/fetch.rs
@@ -76,6 +76,7 @@ fn fetch_impl(args: Args) -> Result<(), ConvertError> {
         last_modified: doc.last_modified().clone(),
         name: doc.get_name(),
         version: doc.get_version(),
+        id: doc.get_document_id(),
     };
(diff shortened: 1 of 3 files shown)


timothyfroehlich (approved):
LGTM

iamralpht (approved):
Thanks for putting this together, Sherief! CI looks happy, so please add this to the merge queue.

In a follow-up change, could you rename `SerializedFigmaDoc` to `SerializedDesignDoc`. FYI @rylin8 since the serialized doc uses the `toolkit_schema` (and no `figma_schema` objects) I think `SerializedDesignDoc` is a better name for it.

rylin8:
Is this going to break all our saved serialized files like the tutorial?

iamralpht:
> Is this going to break all our saved serialized files like the tutorial?

Oh right, yes, so we should bump the doc version, too...

srahman80:
@iamralpht there is also `SerializedFigmaDocHeader`, shouldn't that be changed as well?

PR #66, opened 2023-05-18
Rename SerializedFigmaDoc to SerializedDesignDoc

srahman80:
I mentioned this in the doc_id PR I created so I'll mention it here too so it doesn't get lost.. @iamralpht should we also consier renaming `SerializedFigmaDocHeader`?

### DIFF
--- a/crates/figma_import/src/serialized_document.rs
+++ b/crates/figma_import/src/serialized_document.rs
@@ -18,16 +18,16 @@ use serde::{Deserialize, Serialize};
 
 use crate::{document::FigmaDocInfo, image_context::EncodedImageMap, toolkit_schema, NodeQuery};
 
-static CURRENT_VERSION: u32 = 12;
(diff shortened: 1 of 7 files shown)


iamralpht:
Yes, please update the header. We also need to bump the serialized doc version because of this and the last change, so please do that, too.

crates/figma_import/src/serialized_document.rs
```diff
@@ -18,22 +18,22 @@ use serde::{Deserialize, Serialize};
 
 use crate::{document::FigmaDocInfo, image_context::EncodedImageMap, toolkit_schema, NodeQuery};
 
-static CURRENT_VERSION: u32 = 12;
+static CURRENT_VERSION: u32 = 13;
```
rylin8, line 21: Need to change FSAAS_DOC_VERSION in FsaasSession.kt also.

rylin8 (approved):

PR #70, opened 2023-05-19
Run the "presubmit" actions on the stable branch

timothyfroehlich:
So we can get build results from it.  Should probably rename / refactor presubmit into continuous and presubmit only sometime soon.


timothyfroehlich:
Yep. Same way I've normally done it, cut a branch from main to be "stable" for a specific release, so we can continue making commits to main while testing that release

PR #85, opened 2023-05-26
Support masking layers from Figma

### LINKED ISSUE #63: Support masking layers from Figma, to enable better Dials & Gauges

rylin8:
On the Rust side, parse the "is_mask" parameter from Figma from frames and vectors that have this property, and export this into the corresponding ViewShape that we write into a serialized Figma file.

In Kotlin, reorder children when masks are involved and keep track of some new data to help render masks properly.

When a node has a child that is a mask, we give the node a MaskViewType of MaskParent. This causes two changes when we render the node. First, we save the size of the parent node's draw context so that when it's child mask renders, the mask can make sure it masks the entire parent. Second, we render the parent and all its children under a withSaveLayer() call so that they are rendered in a separate layer to ensure that the alpha channel of the mask is applied properly. When rendering the children of this parent node, we first iterate through the children and reorder them so that masks are rendered after the siblings that they affect.

When a node is a mask, we give the node a MaskViewType of MaskNode. This also causes two changes when we render the node. First, we get the saved size of the parent node and use it to create a rectangle that matches the parent's size. Second, we render the mask under a withSaveLayer() call with blendmode set to DSTIN. This causes the mask to be rendered with alpha blending only where the nodes that it masks were already rendered.

Fixes https://github.com/google/automotive-design-compose/issues/63


rylin8:
Putting this initial implementation up to get some feedback. I spent days trying different approaches and found one that worked, which is this. However, I feel that I got stuck in a particular frame of mind and missed something -- like perhaps there's a way to do this by rendering everything under a mask first and then rendering the mask on top of it.

iamralpht:
Rust change looks good. I'm trying to figure out if there's a simpler way to do the Kotlin side.

> Putting this initial implementation up to get some feedback. I spent days trying different approaches and found one that worked, which is this. However, I feel that I got stuck in a particular frame of mind and missed something -- like perhaps there's a way to do this by rendering everything under a mask first and then rendering the mask on top of it.

I think there might be a simpler rendering model available, that avoids some complexity:
 1. On the parent of the mask (and masked siblings), call saveLayer
 2. Draw the masked siblings
 3. Call saveLayer with blend mode DstIn
 4. Draw the mask layer
 5. Call restore
 6. Call restore

If the mask layer is the same size as the parent layer then the DstIn blend mode ought to result in the parent layer getting cleared everywhere the mask has zero alpha. So this should avoid the need to use clipPath or track individual paths (at the cost of a second temporary layer buffer).

I'm not sure how to map this to Compose best. I wonder if that might be do-able with a `Modifier.drawWithContent` and `compositingStrategy = CompositingStrategy.Offscreen`? Somehow the Composable(s) that make up the mask need to not be rendered with the regular content (via `drawContent`) and only rendered afterwards in a layer with the `DstIn` blend mode... BUT still should work with layout & constraints.

I don't know if it's possible to alter the drawing behavior of something with `is_mask == true` to only draw when we want to do the DstIn layer?

Hopefully that's helpful; I'm not sure if this approach will pan out either. I saw that `react-native-skia` uses a similar approach to doing masking, but they don't have the extra constraints of managing layout of mask layers in-flow with other elements... so it's simpler for them to isolate the mask things from the content things.

rylin8:
> 

So I tried something like this. In DesignView when iterating through children, I tried reordering the children to render the mask at the end, and then using DstIn blendmode. It didn't work -- rendered as if the mask was not there. I think because it was not rendered to a separate buffer? I didn't call saveLayer() and restore() in your step 1 and 6 though. Maybe if we render all non-masked siblings and then the mask to a separate bitmap? I tried `CompositingStrategy.Offscreen` also but it didn't seem to make any difference -- I read that if you are rendering with alpha < 1 it will use `CompositingStrategy.Offscreen` anyway.

iamralpht:
> So I tried something like this. In DesignView when iterating through children, I tried reordering the children to render the mask at the end, and then using DstIn blendmode. It didn't work -- rendered as if the mask was not there. I think because it was not rendered to a separate buffer? I didn't call saveLayer() and restore() in your step 1 and 6 though. Maybe if we render all non-masked siblings and then the mask to a separate bitmap? I tried `CompositingStrategy.Offscreen` also but it didn't seem to make any difference -- I read that if you are rendering with alpha < 1 it will use `CompositingStrategy.Offscreen` anyway.

Hmm, `saveLayer` ought to be the same as rendering to a separate bitmap, with `restore` blending it using whatever was in the `Paint`. I'm not really sure how the Compose CompositingStrategy works with the canvas layer stack...

rylin8:
> Hmm, `saveLayer` ought to be the same as rendering to a separate bitmap, with `restore` blending it using whatever was in the `Paint`. I'm not really sure how the Compose CompositingStrategy works with the canvas layer stack...

Oops I just realized I have been using `save()` this whole time, not `saveLayer()`, which does something different. Let me experiment a bit, I think this might work.

rylin8:
> Oops I just realized I have been using `save()` this whole time, not `saveLayer()`, which does something different. Let me experiment a bit, I think this might work.

No luck so far. I've been trying variations of this, where I call `saveLayer()`, render all masked nodes, then call `saveLayer(BlendMode=DstIn)`, render the mask, then `restore()` twice.

This is what it should render:
Screenshot 2023-06-01 at 5 28 43 PM

This is what I get:
Screenshot 2023-06-01 at 5 28 26 PM

rylin8:
Ready for another review. This latest change does what @iamralpht suggested that simplifies the logic, but does not support multiple masks under a parent or unmasked nodes under a parent that has a child mask.

crates/figma_import/src/toolkit_schema.rs
```diff
@@ -28,14 +28,18 @@ pub use crate::figma_schema::StrokeCap;
 /// Shape of a view, either a rect or a path of some kind.
 #[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
 pub enum ViewShape {
-    Rect,
+    Rect {
+        is_mask: bool,
```
iamralpht, line 32: pity we have to replicate this common field; but there also isn't a convenient parent container to place it in. (no action needed)

designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
```diff
@@ -560,17 +565,44 @@ internal fun Modifier.frameRender(
...
+                    style,
+                    frameShape,
+                    customImageWithContext,
+                    document,
+                    name,
+                    customizations,
+                )
+            when (maskInfo?.type ?: MaskViewType.None) {
+                MaskViewType.MaskNode -> {
+                    // When rendering a mask, call saveLayer with blendmode DSTIN so that we blend
```
iamralpht, line 584: oh, so we can't just make the mask the size of the mask node without having incorrect output, and there's no way to pass parentSize from parent to child because it isn't known until layout, which necessarily happens after the view tree composition...

I think it'd be possible to make a layout modifier that stuffs the parent size value into a more local mutableStateOf (so only the mask view needs the extra state & modifier, and the parent DesignView can be unaware), but I don't think it's worth the effort unless we find out that the arrangement here makes relayout slow.

iamralpht (approved):
This looks good -- like the most minimal change we could make to support masking. I don't think `DrawContext.size` gives us the size we need at the mask view level, so we have to ferry the parent size around in a mutable state.

designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
rylin8: > oh, so we can't just make the mask the size of the mask node without having incorrect output, and there's no way to pass parentSize from parent to child because it isn't known until layout, which necessarily happens after the view tree composition...
> 
> I think it'd be possible to make a layout modifier that stuffs the parent size value into a more local mutableStateOf (so only the mask view needs the extra state & modifier, and the parent DesignView can be unaware), but I don't think it's worth the effort unless we find out that the arrangement here makes relayout slow.

Yea this was the only bit of complexity I ran into. That's an interesting idea though, will definitely look into that for the future.

PR #105, opened 2023-06-02
Adding Emulator testing to CI

timothyfroehlich:
Adding Emulator tests to CI, which will essentially start up an [ATD](https://developer.android.com/studio/test/gradle-managed-devices#gmd-atd) and run `./gradlew connectedCheck` on it. 

As part of getting it running I needed to fix up some tests and update the Figma docs for them.

I've also refactored the start-up and cache handling code for the GitHub actions a bit, to make sure that we have a consistent environment with the different jobs and make the caching more consistent for Android Rust builds.

Also, to save some processor time and for a bit of simplicity, I'm removing the macos runs from the presubmits. It should be fine to only test them nightly.


timothyfroehlich:
@rylin8 Having some trouble with the nightly build so I'm moving that to a separate change,

.github/actionlint-matcher.json
```diff
@@ -0,0 +1,17 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "actionlint",
+      "pattern": [
+        {
+          "regexp": "^(?:\\x1b\\[\\d+m)?(.+?)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*:(?:\\x1b\\[\\d+m)*(\\d+)(?:\\x1b\\[\\d+m)*: (?:\\x1b\\[\\d+m)*(.+?)(?:\\x1b\\[\\d+m)* \\[(.+?)\\]$",
```
rylin8, line 7: what is this voodoo?

rylin8 (commented):
Are the serialized Figma files checked in and updated just to be consistent and so we have it in the future? Most are not used yet, correct?

timothyfroehlich:
Yes, not all the committed Figma serialized files are used in tests right now. I took a bit of a shotgun approach to updating the ones we needed for the tests, so some were updated that didn't strictly need to be.

.github/actionlint-matcher.json
timothyfroehlich: A [GitHub Problem Matcher](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md) file for [ActionLint](https://github.com/rhysd/actionlint)

ActionLint lints our GitHub actions workflows, and the problem matcher highlights problems that were found in the "Files Changed" view of the PR, just like how the presubmits in Critique show where in the code the problem was.

integration-tests/validation/src/androidTest/java/com/android/designcompose/testapp/validation/InteractionTests.kt
```diff
@@ -45,36 +44,36 @@ class InteractionTests {
 
         // Navigate to the Triggers -> Timeouts menu
         composeTestRule.onNodeWithText("Triggers").performClick()
-        composeTestRule.onNodeWithText("While Pressed").assertIsDisplayed().performClick()
-        composeTestRule.onNodeWithText("Timeouts").assertIsDisplayed().performClick()
+        composeTestRule.onNodeWithText("While Pressed").assertExists().performClick()
+        composeTestRule.onNodeWithText("Timeouts").assertExists().performClick()
```
rylin8, line 48: Forgot to ask -- why was `assertIsDisplayed` replaced with `assertExists`?

timothyfroehlich: I was trying to figure out why the Tutorial app test always failed on Gradle Managed Device tests (like `./gradlew tabletAtdApi30Check`). I realized that the Gradle managed devices always start up in landscape mode, and for some reason Compose doesn't think the main screen of the Tutorial app is visible in landscape (it may have something to do with the main screen being taller than the entire screen when the device is in landscape). Switching things to "assertExists()" fixes it.

rylin8 (approved):

PR #113, opened 2023-06-07
Automate creation of release artifacts

timothyfroehlich:
This sets up a new workflow action that triggers when a new tag is created, building the Maven repo and the Figma resources and uploading them to the GitHub release for that tag. The Gradle build will have the version set to the tag's name, so if a tag named "v0.20.0-rc03", the DesignCompose libraries will be version "0.20.0-rc03". (No more making commits to set the version to rc02 or rc03 or whatever).


timothyfroehlich:
I did some test releases in a fork of the repo. Here's a test using the latest push: https://github.com/timothyfroehlich/automotive-design-compose/releases/tag/v0.2.test2

rylin8 (approved):

PR #117, opened 2023-06-09
Update the UI per latest designs

### LINKED ISSUE #116: Update dials and gauges plugin UI
Per Aubrey's design
Screenshot 2023-06-07 at 4 35 00 PM

rylin8:
Update Dials & Gauges plugin UI
    
Update the UI per latest designs. This adds images, tooltips, and incremental progress tick marks.
    
This commit slightly modifies how progress bar markers work. Progress bar marker positions are now calculated by simply moving them along the x-axis, with the range calculated using the parent frame's width and the marker's width. The marker now moves such that at the endpoints, the middle of the marker aligns with the edge of the parent.
    
This simplification relies on a change in how the parent frame of the marker is setup. Whereas previously the marker assumed that it had an axis aligned parent frame, it now assumes that the parent has the same skew and rotation transform as the marker. This means that in order to make a skewed or rotated progress bar, a designer would first create the marker and its parent as an axis aligned rectangle, then apply the skew/rotation to the parent, which affects the marker as well.
    
Fixes #116


rylin8:
@aubreydagoog has reviewed this and I've made a few minor tweaks. Ready for review.

rylin8:
Screenshot 2023-06-27 at 4 22 34 PM

urpatil (commented):
LGTM

rylin8:
Updated text and images.

rylin8:
Merging this since the deadline is coming up. There may be minor text changes after the release but it has already gone through a few revisions and is in a much better place.

PR #122, opened 2023-06-13
Add RenderMethod enum to View for use with telltales.

srahman80:
The RenderMethod enum will be used to indicate which nodes are telltales.  Setting this enum to "PixelPerfect" will indicate to the High Availability Renderer which nodes are telltale, in case we need to render them differently.

(Note, previously named "RenderStyle", which was colliding with vector_schema::RenderStyle in gradle.)


timothyfroehlich (approved):

johnoneil:
I'm good with this approach. Thanks for the heads up.

crates/figma_import/src/toolkit_schema.rs
```diff
@@ -105,6 +105,14 @@ impl Default for ScrollInfo {
     }
 }
 
+// This will be used to indicate which nodes are telltales, in the case
```
iamralpht, line 108: Can you make this a doc comment? I think it's also worth generalizing the language a bit to say that the DesignCompose renderer implementation can use this hint to determine if it's important for the content to be rendered identically on different platforms.

iamralpht:
This change modifies the serialized form, so you'll also need to regenerate the serialized docs that get checked in (like Design Switcher). It looks like we have a CI pass that detects this now (hooray @timothyfroehlich !)

crates/figma_import/src/toolkit_schema.rs
srahman80: Updated comment

srahman80:
> This change modifies the serialized form, so you'll also need to regenerate the serialized docs that get checked in (like Design Switcher). It looks like we have a CI pass that detects this now (hooray @timothyfroehlich !)

rylin8 (approved):

PR #124, opened 2023-06-14
Fix setting Figma Token on muti-user systems, add rough test script

timothyfroehlich:
This allows the token to be set on a multi-user system, which would previously fail.

I've included a very rough script for testing it. If needed I'll turn it into something more automatic, though testing behavior across multiple users requires more than just writing an AndroidInsturmentedTest. (https://source.android.com/docs/devices/admin/multi-user-testing)


urpatil (approved):
LGTM

PR #138, opened 2023-06-23
Add support for apps to provide a callback that gets called with the document byte data when the document is updated.

### LINKED ISSUE #135: Provide access to a Figma document's serialized content.
Provide a callback that Apps can use to access the Figma document's representation as a ByteArray.  This is the content that is serialzed to disk. 

- the document is loaded by the SDK from disk
- the document is fetched by the SDK from Figma during the initial fetch or due to design updates

rylin8:
The content provided is the content that is serialzed to disk. The callback is called when the document is loaded from disk and whenever the document is updated live from Figma changes.

Fixes #135

### DIFF
--- a/common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
+++ b/common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
@@ -43,17 +43,25 @@ class GenericDocContent(
     fun save(filepath: File, feedback: FeedbackImpl) {
         feedback.documentSaveTo(filepath.absolutePath, docId)
         try {
-            val serializer = BincodeSerializer()
-            header.serialize(serializer)
-            document.serialize(serializer)
+            val bytes = toSerializedBytes(feedback)
             val output = FileOutputStream(filepath)
-            output.write(serializer._bytes)
-            output.write(imageSessionData)
+            output.write(bytes)
             output.close()
             feedback.documentSaveSuccess(docId)
         } catch (error: Throwable) {
             feedback.documentSaveError(error.toString(), docId)
         }
+    }
+    fun toSerializedBytes(feedback: FeedbackImpl): ByteArray? {
+        try {
+            val serializer = BincodeSerializer()
+            header.serialize(serializer)
+            document.serialize(serializer)
+            return serializer._bytes.toUByteArray().toByteArray() + imageSessionData
+        } catch (error: Throwable) {
+            feedback.documentSaveError(error.toString(), docId)
+        }
+        return null
     }
 }
 
--- a/designcompose/src/main/java/com/android/designcompose/DesignView.kt
+++ b/designcompose/src/main/java/com/android/designcompose/DesignView.kt
@@ -817,6 +817,11 @@ enum class LiveUpdateMode {
     OFFLINE // Live updates off (load from serialized file)
 }
 
+class DesignComposeCallbacks(
+    val docReadyCallback: ((String) -> Unit)? = null,
+    val newDocDataCallback: ((ByteArray?) -> Unit)? = null,
+)
+
 @Composable
 fun DesignDoc(
     docName: String,
@@ -828,7 +833,7 @@ fun DesignDoc(
     serverParams: DocumentServerParams = DocumentServerParams(),
     setDocId: (String) -> Unit = {},
     designSwitcherPolicy: DesignSwitcherPolicy = DesignSwitcherPolicy.SHOW_IF_ROOT,
-    designDocReadyCallback: ((String) -> Unit)? = null,
+    designComposeCallbacks: DesignComposeCallbacks? = null,
     parentComponents: List<ParentComponentInfo> = listOf(),
 ) =
     DesignDocInternal(
@@ -841,7 +846,7 @@ fun DesignDoc(
         serverParams = serverParams,
         setDocId = setDocId,
         designSwitcherPolicy = designSwitcherPolicy,
-        designDocReadyCallback = designDocReadyCallback,
+        designComposeCallbacks = designComposeCallbacks,
         parentComponents = parentComponents,
     )
 
@@ -857,11 +862,18 @@ internal fun DesignDocInternal(
     setDocId: (String) -> Unit = {},
     designSwitcherPolicy: DesignSwitcherPolicy = DesignSwitcherPolicy.SHOW_IF_ROOT,
     liveUpdateMode: LiveUpdateMode = LiveUpdateMode.LIVE,
-    designDocReadyCallback: ((String) -> Unit)? = null,
+    designComposeCallbacks: DesignComposeCallbacks? = null,
     parentComponents: List<ParentComponentInfo> = listOf(),
 ) {
     val docId = DocumentSwitcher.getSwitchedDocId(incomingDocId)
-    val doc = DocServer.doc(docName, docId, serverParams, liveUpdateMode == LiveUpdateMode.OFFLINE)
+    val doc =
+        DocServer.doc(
+            docName,
+            docId,
+            serverParams,
+            designComposeCallbacks?.newDocDataCallback,
+            liveUpdateMode == LiveUpdateMode.OFFLINE
+        )
     val interactionState = InteractionStateManager.stateForDoc(docId)
     val interactionScope = rememberCoroutineScope()
     val isRoot = LocalDesignIsRootContext.current.isRoot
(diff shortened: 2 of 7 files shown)


common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
```diff
@@ -43,18 +43,26 @@ class GenericDocContent(
...
         } catch (error: Throwable) {
             feedback.documentSaveError(error.toString(), docId)
         }
     }
+    fun toSerializedBytes(feedback: FeedbackImpl): ByteArray? {
+        try {
+            val serializer = BincodeSerializer()
+            header.serialize(serializer)
+            document.serialize(serializer)
+            return serializer._bytes.toUByteArray().toByteArray() + imageSessionData
```
urpatil, line 60: Can a copy be avoided if you were to use asByteArray instead of toByteArray?

urpatil (commented):
Looking good.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -817,6 +817,11 @@ enum class LiveUpdateMode {
     OFFLINE // Live updates off (load from serialized file)
 }
 
+class DesignDocCallbacks(
```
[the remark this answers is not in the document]
rylin8: Sounds good, will make the change.

timothyfroehlich (changes requested):
I'd like you to add some documentation for this, at least in the guide docs. I would really like for us to go through everything and add kdoc to it so that we can start generating API doc, but at least add something to the guides that describes this.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -817,6 +817,11 @@ enum class LiveUpdateMode {
     OFFLINE // Live updates off (load from serialized file)
 }
 
+class DesignComposeCallbacks(
+    val docReadyCallback: ((String) -> Unit)? = null,
+    val docUpdateCallback: ((ByteArray?) -> Unit)? = null,
```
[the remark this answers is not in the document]
urpatil: Maybe via DesignSettings.

common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
rylin8: Oh yea, good point.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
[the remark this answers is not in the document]
rylin8: I mean, yea this is a bit clunky, I was trying to throw something together quickly for Umesh, but also, if you don't want to get serialized file updates every time it updates, then just don't register for the callback. If we want a callback to just inform you that there has been an update, we can add that functionality or rename this callback.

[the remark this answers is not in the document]
timothyfroehlich: I still feel that serializing the file shouldn't happen automatically as part of the library calling the callback. It should be the code that registers the callback that makes requests for further work to be done by the library.
And yeah, I think it'd be better to handle registering of listeners via DesignSettings, and that's where we could have the function to request the serialized file.

designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -447,20 +450,24 @@ internal fun DocServer.doc(
                 }
                 targetDoc
             }
+        docUpdateCallback?.invoke(targetDoc?.c?.toSerializedBytes(Feedback))
```
urpatil, line 453: Regarding the discussion around registering callbacks:

@timothyfroehlich, based on this change, the serialization won't happen if docUpdateCallback is null.

i.e. the code doesn't unnecessarily generate that data.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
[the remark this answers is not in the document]
urpatil: Rob, I am fine with this change since the data is not fetched if the callback is null.  What you have currently is an alternate form for registering callbacks - like with add/remove way via DesignSettings.

urpatil (approved):

designcompose/src/main/java/com/android/designcompose/DocServer.kt
timothyfroehlich: Agreed. What I'm saying is that having a "docUpdateCallback" is a generically useful feature, but it should not automatically serialize the doc whenever it is called. I'm sure there are use cases where an app developer would want to react to the doc updating without wanting to do anything with the serialized file.

Wouldn't it be much more useful to add a function to provide the serialized file on demand, rather than requiring the user to register a callback? I'm asking for a better generic design, not one that fulfills only exactly Umesh's use case.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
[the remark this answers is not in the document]
rylin8: I did explore using DesignSettings when I started this and it's not that simple. The reason I did it this way is that the generated code knows the ID of the Figma document you care about, and that's needed so that your callback is called on the correct document getting updated.

I can also call this function something like `newDocDataCallback`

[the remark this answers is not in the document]
timothyfroehlich: Alright, overruled. I still want to have it added to the documentation though.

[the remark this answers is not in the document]
rylin8: Let me think about this a bit over the weekend, and I'll also add some documentation when I update this.

designcompose/src/main/java/com/android/designcompose/DocServer.kt
urpatil: If LiveUpdate is disabled, is this where the document would be loaded from disk and the callback invoked?

rylin8: Yes

rylin8:
Updated callback name and added some documentation.

docs/_docs/modifiers/modifiers.md
```diff
@@ -843,6 +843,13 @@ generated @DesignComponent function generates these three optional parameters:
     }
     ```
 
+*   `designComposeCallbacks: DesignComposeCallbacks? = null`. Set this parameter to register callbacks for a certain events. The `DesignComposeCallbacks` class supports the following event callbacks:
+    * `docReadyCallback: ((String) -> Unit)? = null`. \
+    This callback is called when the Figma document has loaded is ready to be rendered.
+    * `newDocDataCallback: ((ByteArray?) -> Unit)? = null` \
+    This callback is called both when the Figma document has first loaded and whenever it updates from live update changes. The ByteArray passed in contains the serialized Figma file.
+)
```
timothyfroehlich, line 851: Extra close-parentheses
```suggestion
```

docs/_docs/modifiers/modifiers.md
```diff
@@ -843,6 +843,13 @@ generated @DesignComponent function generates these three optional parameters:
     }
     ```
 
+*   `designComposeCallbacks: DesignComposeCallbacks? = null`. Set this parameter to register callbacks for a certain events. The `DesignComposeCallbacks` class supports the following event callbacks:
+    * `docReadyCallback: ((String) -> Unit)? = null`. \
```
timothyfroehlich, line 847: Extra period after the code backtick.

timothyfroehlich (approved):

docs/_docs/modifiers/modifiers.md
rylin8: Fixed

docs/_docs/modifiers/modifiers.md
rylin8: Fixed

common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
urpatil: Rob, just checking the use of toByteArray - would asByteArray be better or does it not work in this case?

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -817,6 +817,11 @@ enum class LiveUpdateMode {
     OFFLINE // Live updates off (load from serialized file)
 }
 
+class DesignComposeCallbacks(
+    val docReadyCallback: ((String) -> Unit)? = null,
+    val newDocDataCallback: ((ByteArray?) -> Unit)? = null,
```
urpatil, line 822: Please include the docID as well as the first argument.

common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
rylin8: Yes it should work and I remember changing it, but it somehow didn't make it. Hmm, I'll make a follow up pull request.

PR #143, opened 2023-06-26
Fix a bug where clicking "load" on the design switcher without setting an ID will soft-lock the switcher.

### LINKED ISSUE #56: Bug: Clicking "load" on the design switcher without setting an ID will soft-lock the switcher.
Repro:

Launch an app, set the figma key.
Expand the design switcher, click "Change", then click "load" without setting a new ID. You won't be able to click Change any more.

rylin8:
Check for empty string before executing a load in the design switcher. This also prevents the design switcher overlay when clicking "load" with an empty ID. Check for empty string before switching documents in the design switcher as an extra safety check.

This commit also does a small amount of cleanup in the design switcher code so that all the composable functions use the same LiveUpdateMode setting.

Fixes #56

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt
+++ b/designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt
@@ -77,7 +77,6 @@ private interface DesignSwitcher {
             "#ShowRecompositionCheckbox",
             "#Name",
             "#Id",
-            "#GoButton",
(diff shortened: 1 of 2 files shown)


timothyfroehlich (approved):
Gave it a test, works well!

rylin8:
Had to fix a typo in the commit: 'but' -> 'bug'

PR #144, opened 2023-06-26
Follow up to document data update callback

### LINKED ISSUE #135: Provide access to a Figma document's serialized content.

rylin8:
Add the document ID to the callback.

Fixes #135


urpatil (approved):
Thx!

PR #148, opened 2023-06-28
Reference app not found on device when setting up figma token

narko:
This PR tries to solve an issue when running the `figma-token-task` using:
```
FIGMA_ACCESS_TOKEN=<YOUR_ACCESS_TOKEN> \
    ./gradlew setFigmaTokenDebug
```

Without this change, Gradle would just show the following output:
```
> Task :tutorial-app:setFigmaTokenDebug
Skipping setFigmaTokenDebug, com.android.designcompose.tutorial not found on device
```

This makes it impossible to set the Figma token.

I've tried this on different targets:
- Polestar 2 emulator
- Several pixel phones

This seems to be related to the changes introduced [here](https://github.com/google/automotive-design-compose/pull/124/files#diff-8c18816857b5d7b270c84e0854c09cf3844caede95d3cf47fcd1007c815d198fR80)

### DIFF
--- a/build-logic/src/main/kotlin/designcompose/features/figma-token-task.kt
+++ b/build-logic/src/main/kotlin/designcompose/features/figma-token-task.kt
@@ -77,7 +77,7 @@ abstract class FigmaTokenTask @Inject constructor(private val executor: DefaultE
         val execResult =
             executor.exec {
                 executable = adbPath.get().toString()
-                args("shell", "pm", "list", "packages", "--user", "current", appID.get())
+                args("shell", "pm", "list", "packages", appID.get())
                 standardOutput = stdOut
                 errorOutput = stdErr
                 isIgnoreExitValue = true // Handle it ourselves


timothyfroehlich:
Interesting, I'd needed to add that bit in order to support Android devices with multiple users. I wonder if this is an API issue. What API level were you using?

narko:
> Interesting, I'd needed to add that bit in order to support Android devices with multiple users. I wonder if this is an API issue. 
Thanks for having a look at it. It could be that is an API issue. 

>What API level were you using?
The public Polestar 2 emulator uses API 29. I've got also API 29 in my Pixel 2 device.

timothyfroehlich:
Hi @narko , we had a quick discussion about it and realized that we hadn't documented the API levels we support. The intention was to support API 30 and above. I realized that we'd used the wrong variable to set the minimum SDK levels in the apps and I've put up #162 to fix that. I also filed #163 to properly document the API levels we support.

I checked myself and confirmed that the API key can be set on the API 30 Automotive image ( system-images;android-30;android-automotive-playstore;x86_64 ). Can you use that instead?

narko:
Thanks for the info @timothyfroehlich! I'll use API 30+.

Closing this PR then.

PR #149, opened 2023-06-28
Add a "Change" button to the Design Switcher loading state.

### LINKED ISSUE #71: Bug: Cannot switch design if current one isn't loading
I'm testing the Tutorial app right now with 0.19.0-rc02 and the serialized file is missing from the checkout. (It should be in `reference-apps/tutorial/app/src/main/assets/figma`, see #72 ). The Design Switcher stays in a loading state and doesn't display the change design button

image

Essentially an app in this state is stuck.

rylin8:
This lets you change the Figma document if the current one has a problem loading.

Fixes #71


timothyfroehlich (approved):
Gave it a test and it's working for me. Thanks!

PR #153, opened 2023-06-28
Update MediaCompose to build with latest release of Unbundled AAOS

### LINKED ISSUE #67: MediaCompose is incompatible with the current release of Unbundled AAOS
Error:
```
e: file:///usr/local/google/home/froeht/git/designcompose/reference-apps/aaos-unbundled/standalone-projects/mediacompose/media/src/main/java/com/android/designcompose/reference/media/MediaAdapter.kt:467:39 Unresolved reference: artist
e: file:///usr/local/google/home/froeht/git/designcompose/reference-apps/aaos-unbundled/standalone-projects/mediacompose/media/src/main/java/com/android/designcompose/reference/media/MediaAdapter.kt:468:38 Unresolved reference: albumTitle
```

We can continue using the previous release. To fetch a clean copy of it run this:
```
    HOST=https://android.googlesource.com/platform/superproject
    BRANCH=ub-automotive-master-20230303
    git clone "$HOST" "$BRANCH" --branch="$BRANCH" \
    --single-branch --filter=blob:none --also-filter-submodules \
    --recurse-submodules --jobs=4
```

rylin8:
Also updated agp plugin version to be compatible with unbundled aaos.

Fixes #67

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -178,7 +178,7 @@ jobs:
     - name: Check out the Unbundled AAOS Repo
       run: |
         HOST=https://android.googlesource.com/platform/superproject
-        BRANCH=ub-automotive-master-20230303
+        BRANCH=ub-automotive-master-20230622
         mkdir "$ORG_GRADLE_PROJECT_unbundledAAOSDir"
         git clone "$HOST" "$ORG_GRADLE_PROJECT_unbundledAAOSDir" --branch="$BRANCH" \
         --single-branch --filter=blob:none --recurse-submodules --jobs=4
--- a/docs/_docs/media-center-demo/index.md
+++ b/docs/_docs/media-center-demo/index.md
@@ -68,7 +68,7 @@ them available to the Media Center app.
     mkdir aaos-unbundled
     cd aaos-unbundled
 
-    repo init -u https://android.googlesource.com/platform/manifest -b ub-automotive-master-20230303
+    repo init -u https://android.googlesource.com/platform/manifest -b ub-automotive-master-20230622
     repo sync -cq -j16
     ```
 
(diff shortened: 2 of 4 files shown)


rylin8:
I updated to the latest unbundled aaos repo and tested this change. I needed to update unbundledAAOSAndroidGradlePluginVer to make it build. When testing with Spotify, I noticed that the now playing data never showed the artist or album anymore. I checked with Arnaud and he had me verify with some other sources that it does in fact work, and commented that Spotify is likely not populating the artist and album fields properly.

timothyfroehlich (changes requested):
Changes look good to me, but we need to update the tests and doc to stop pulling down the old release. Change all the instances of "ub-automotive-master-20230303" in the repo to just "master".

.github/workflows/main.yml
```diff
@@ -178,7 +178,7 @@ jobs:
     - name: Check out the Unbundled AAOS Repo
       run: |
         HOST=https://android.googlesource.com/platform/superproject
-        BRANCH=ub-automotive-master-20230303
+        BRANCH=ub-automotive-master
```
iamralpht, line 181: Do we want to just use master, or do we want to pin to a particular release again? (This looks like the most recent: https://android.googlesource.com/platform/manifest/+/refs/tags/ub-automotive-master-20230622 )

rylin8: Given that master is failing to build here (though it builds on my machine with a sync from yesterday), I'll give this a try

iamralpht: Sounds good -- looks like the CI checks passed using the tag, too.

timothyfroehlich (approved):

PR #156, opened 2023-06-29
Fix variant selection not propagating to children properly

### LINKED ISSUE #146: Variant selection does not propagate to children properly
The dynamic variant selection system where we select the appropriate variant based on enum values does not propagate if there are nested variants. For example if the main frame #stage has a variant `DriveMode` for which we select the variant #drivemode=Driving, and within that variant there is another variant `PowerMode` for which we want to select the variant #powermode=regen, this second variant selection will fail and just use whatever is used in Figma.

rylin8:
When we use the @DesignVariant annotation to declare a variant customization and then change the variant by passing in an enum argument, the code takes a path in DesignFrame() where it resolves the variant parameters and finds the correct node to actually compose. Then it calls a customComposable() composable function that composes the variant we want. There was a problem where the customComposable function did not propagate all customizations down to its children properly: specifically, it did not propagate the customComposable() function itself. This has been fixed in the CustomizationContext.mergeFrom() function.

The Variant Properties test has been modified to test for nested variants. While testing this another issue came up where variants can sometimes render two states at the same time: the one saved in Figma, and the one set dynamically in code. This has been fixed by only calling frameRender() if we are not calling the customComposable() function to render a different variant.

Fixes #146

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
+++ b/designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
@@ -94,10 +94,18 @@ internal fun DesignFrame(
     // it until frameRender() since that function is not a composable.
     val meterValue = customizations.getMeterFunction(name)?.let { it() }
     meterValue?.let { customizations.setMeterValue(name, it) }
-    var m =
-        Modifier.layoutStyle(name, style)
-            .frameRender(style, shape, customImage, document, name, customizations, maskInfo)
-            .then(modifier)
+
+    // Check to see if this node is part of a component set with variants and if any @DesignVariant
+    // annotations set variant properties that match. If so, variantNodeName will be set to the
+    // name of the node with all the variants set to the @DesignVariant parameters
+    val variantNodeName = customizations.getMatchingVariant(componentInfo)
+
+    var m = Modifier.layoutStyle(name, style)
+    // Only render the frame if we don't have a custom variant node that we are about to
+    // render instead
+    if (variantNodeName.isNullOrEmpty())
+        m = m.frameRender(style, shape, customImage, document, name, customizations, maskInfo)
+    m = m.then(modifier)
 
     val customModifier = customizations.getModifier(name)
     if (customModifier != null) {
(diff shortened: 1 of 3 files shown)


designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
```diff
@@ -124,10 +132,7 @@ internal fun DesignFrame(
...
     }
 
-    // Check to see if this node is part of a component set with variants and if any @DesignVariant
-    // annotations set variant properties that match. If so, variantNodeName will be set to the
-    // name of the node with all the variants set to the @DesignVariant parameters
-    val variantNodeName = customizations.getMatchingVariant(componentInfo)
+    // If we a custom variant, compose it instead and then return.
     if (variantNodeName != null) {
         // Get the generated CustomVariantComponent() function and call it with variantNodeName
         val customComposable = customizations.getCustomComposable()
```
iamralpht, line 138: Is it possible for `variantNodeName` to be non-null, but `customComposable` is null? Then I think we'd render the node without its background (but maybe this is an error and we should do something else?)

rylin8: It was possible before, which was the first part of the fix in the `mergeFrom()` function. From the commit:
"There was a problem where the customComposable function did not propagate all customizations down to its children properly: specifically, it did not propagate the customComposable() function itself. This has been fixed in the CustomizationContext.mergeFrom() function."

Now, it *shouldn't* be possible anymore because `customizations.setCustomComposable()` is called in generated code, and anytime we pass this down to a child it gets copied to the child customizations.

iamralpht: I don't know if there's a cleaner way to express the logic, because it takes a bit of reasoning to get there, but this makes sense to me.

iamralpht (approved):

designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
rylin8: Ok, this logic may be moving to Rust soon, so I'm going to leave it for now.

PR #165, opened 2023-06-30
Add more development process doc

timothyfroehlich:


README.md
```diff
@@ -150,7 +133,71 @@ npm run build
...
+
+## Updating serialized Figma files
+
+Any changes that include an update to the Figma serialized file version (set in crates/figma_import/src/serialized_document.rs) will require the committed serialized files to be updated with new versions that have been fetched with the updated version. At a minimum this includes any files that are being tested with AndroidIntegratedTests, and most especially the Tutorial's welcome screen, which is a special case.
+
+The Tutorial's DesignDoc is set to the main development Figma file, to assist in development of the app, but the committed serialized file is a separate file that presents a "welcome" page. This file's ID is `BX9UyUa5lkuSP3dEnqBdJf`. To update the file, fetch the `BX9UyUa5lkuSP3dEnqBdJf` file, then replace `reference-apps/tutorial/app/src/main/assets/figma/TutorialDoc_3z4xExq0INrL9vxPhj9tl7` with the serialized file. **The file name will remain the same**, the `TutorialDoc_3z4xExq0INrL9vxPhj9tl7` fill will contain the serialized `BX9UyUa5lkuSP3dEnqBdJf` file. The Tutorial app project has an AndroidIntegratedTest to ensure that the correct file is set, and it will be run as part of running the `./dev-scripts/test-all.sh` script.
```
rylin8, line 142: This process is so confusing. Thanks for explaining it here. We really need a way to automate this.

README.md
```diff
@@ -150,7 +133,71 @@ npm run build
...
+    3. Update the Plugin
+        1. Open the **Resources** menu(right of the `T` text menu) and switch to the **Plugins** tab
+        2. Change the dropdown under the search bar to **Development**
+        3. If you already have an entry for the plugin then hover over it, click the `...` menu and click **Remove local version**
+        4. The plugin will now have a **Locate local version** option, click it, then navigate to the `manifest.json` of the plugin artifact you downloaded from GitHub
+        5. Click the `...` menu again and click Publish. Add any release notes and click **Publish new version**
+    4. Do the same as above for the widget (except use the widget menu)
+5. Update the Tutorial
+    1. Open the Tutorial Figma File and create a new branch
+    2. Find each instance of the widget and replace it with the newly published version, matching the settings to the current one
```
rylin8, line 193: This is also a huge pain and it would be great to automate this. Since the widget shows the version now we should make sure that it does in fact keep the old version in the doc when we update the widget to a newer version. Maybe opening up the widget in the doc will automatically update to a newer version? We should explore automating this also if not.

rylin8 (approved):

README.md
timothyfroehlich: Yep, that's a big reason why I've kept bringing up the fetcher stuff. Once that's done then it'll be simple.

README.md
timothyfroehlich: According to https://www.figma.com/widget-docs/stability-and-updates/#updating-your-widget-safely they're working on it, maybe we should get in touch. But one way to do it would be to have some kind of "export settings" button that generates something that we can paste into the new instance.

PR #194, opened 2023-07-17
Fetch images via proxy

urpatil:
Updated the code that fetches images to use a proxy if it is configured.


timothyfroehlich:
Cherry-pick into 0.21 and release a 0.21.1

PR #228, opened 2023-07-20
Fix layout issue with vectors using scale

### LINKED ISSUE #227: Layout issue with vectors using scale
Vectors with constraints set to scale and then rendered at a different size sometimes have layout and size problems.

rylin8:
Vectors with their constraints set to "scale" had their style's left/right/top/bottom properties set to Dimension::Percent, but these properties were not being used to scale the vectors. This commit adds some additional size data for vectors that are then used in comparison of the frame size to scale the vector.

Fixes #227


designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -522,11 +525,27 @@ internal fun ContentDrawScope.render(
 
     fun getPaths(
         path: List<com.android.designcompose.serdegen.Path>,
-        stroke: List<com.android.designcompose.serdegen.Path>
+        stroke: List<com.android.designcompose.serdegen.Path>,
+        vectorSize: Optional<List<Float>>,
```
timothyfroehlich, line 529: Nit: Why not use a nullable?

timothyfroehlich (commented):
Kotlin looks good, will wait for Ralph to review the Rust code though.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: Just because in Rust, this was an Option<(f32, f32)> and serde generates this into Optional<List<Float>>

iamralpht (approved):
Looks good; I think this may have been the root cause for a misalignment in one of the telltale validation docs, so it would be good to do a before/after on those.

rylin8:
Oops that doesn't look right...
Screenshot 2023-07-26 at 5 17 58 PM

rylin8:
Ok I fixed the massive scale factor bug shown in the screenshot above, but @iamralpht can you think of where this can fail? I feel like I'm missing something here, I don't think I should need to decompose the transform. This doesn't change the rendering of the telltale test. The rendering failures there still exist, and I think are related to the other rotation bug.

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1122,15 +1122,27 @@ fn visit_node(
         }
     };
 
+    // Save the vector frame size which is needed to compare against the
+    // runtime frame size to calculate the scaling factor if a vector's
+    // constraints are set to scale.
+    let mut vector_size: Option<(f32, f32)> = None;
+    if let Some(size) = &node.size {
```
iamralpht, line 1129: I think you can collapse this --
```rust
if let Some(Size { x: Some(x), y: Some(y) } = &node.size {
...
```

rylin8: Ok fixed, plus rebased and added serialized files.

iamralpht (approved):
Looks good; let's get this one in, and later I can rebase the border radius change (and remember to update the serialized files)

iamralpht:
Hm, this change caused some issues with Battleship and some of the vector elements in the Design Switcher. I'll see if I can make a fix, otherwise I'll put up a revert.

iamralpht:
This improves matters; things are mostly aligned, but Battleship has a bad tile (it always did, but now it's different).
```patch
diff --git a/designcompose/src/main/java/com/android/designcompose/FrameRender.kt b/designcompose/src/main/java/com/android/designcompose/FrameRender.kt
index 76e6ce3..c890819 100644
--- a/designcompose/src/main/java/com/android/designcompose/FrameRender.kt
+++ b/designcompose/src/main/java/com/android/designcompose/FrameRender.kt
@@ -544,8 +544,8 @@ internal fun ContentDrawScope.render(
         vectorSize.ifPresent {
             val sizeList = vectorSize.get()
             if (sizeList.size == 2) {
-                val vecWidth = sizeList[0] * vectorScaleX
-                val vecHeight = sizeList[1] * vectorScaleY
+                val vecWidth = sizeList[0] * vectorScaleX * density
+                val vecHeight = sizeList[1] * vectorScaleY * density
                 if (style.left is Dimension.Percent) scaleX = frameSize.width / vecWidth
                 if (style.top is Dimension.Percent) scaleY = frameSize.height / vecHeight
             }
```

PR #234, opened 2023-07-24
Add an argument to the test-all script for setting the Unbundled repo path

timothyfroehlich:


dev-scripts/test-all.sh
```diff
@@ -19,20 +19,24 @@ cat  <<END
...
     Have \$FIGMA_ACCESS_TOKEN set to your actual Figma token
+    To run the full suite your system must be able to run emulator tests.
+      Use the -s argument to skip emulator tests.
 END
 }
 
 OPTIND=1         # Reset in case getopts has been used previously in the shell.
 run_emulator_tests=1
-while getopts "s" opt; do
+while getopts "su:" opt; do
```
rylin8, line 32: What's the colon here for?

timothyfroehlich: It's part of getopts, it specifies that this option takes an argument and isn't just a flag.

rylin8 (approved):

PR #245, opened 2023-07-27
Fix Dependabot config

timothyfroehlich:


rylin8:
Are you able to force the dependabot check to run?

timothyfroehlich:
> Are you able to force the dependabot check to run?

I don't think so, why do you ask?

rylin8:
> 

To see if it fixes the problem

timothyfroehlich:
> > 
> 
> To see if it fixes the problem

Oh, I think we're good. Turns out I did have a linter running in VSCode and it was pointing out problems, but I missed it. It's not reporting any problems any more.

rylin8 (approved):

PR #251, opened 2023-07-28
Add new Dependabot registries to the Gradle package ecosystem

timothyfroehlich:
Dependabot encountered the following error when parsing your .github/dependabot.yml:

The property '#/registries' includes the "maven-google" registry which is not used in any of the configurations
The property '#/registries' includes the "gradle-plugin-portal" registry which is not used in any of the configurations

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -93,6 +93,9 @@ updates:
       interval: weekly
 
   - package-ecosystem: gradle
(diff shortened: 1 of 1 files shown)


timothyfroehlich:
(This one wasn't caught by the linter)

rylin8 (approved):

PR #265, opened 2023-07-31
Remove unused GRPC dependency from DesignCompose

timothyfroehlich:
This should resolve https://github.com/google/automotive-design-compose/security/dependabot/73 and related high priority alerts.

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/DocServer.kt
+++ b/designcompose/src/main/java/com/android/designcompose/DocServer.kt
@@ -34,7 +34,6 @@ import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.lifecycleScope
 import com.android.designcompose.common.DocumentServerParams
 import com.google.common.annotations.VisibleForTesting
-import io.grpc.StatusRuntimeException
 import java.io.BufferedInputStream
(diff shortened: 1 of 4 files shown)


designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -218,24 +217,15 @@ internal fun DocServer.initializeLiveUpdate() {
     periodicFetchRunnable =
         Runnable() {
             thread {
-                try {
-                    if (!fetchDocuments(firstFetch)) {
-                        Log.e(TAG, "Error occurred while fetching or decoding documents.")
-                        return@thread
-                    }
-                    firstFetch = false
-                } catch (e: StatusRuntimeException) {
```
rylin8, line 227: I don't remember clearly -- did `fetchDocuments()` used to throw a GRPC exception and we just didn't remove this code?

timothyfroehlich: Yep, it came from the old FSAAS code. This bit didn't get cleaned up when FSAAS was removed.

rylin8 (approved):

PR #280, opened 2023-08-09
Replace AWS keys iin cached figma response with dummies

timothyfroehlich:
Should resolve issues where git commits are blocked


rylin8:
Hard to diff this, as everything in that blob changed. @johnoneil when you get a chance can you see if this fixes the problem you encountered?

johnoneil (approved):
See comment above.
Was able to recreate issue and this fixed it.

PR #294, opened 2023-08-18
Update Compose dependencies to the latest, update compileSdk to match.

timothyfroehlich:
Suppress warning about Api 34 needing a newer version of AGP since the newer version hasn't been released yet.


rylin8:
Can I assume you did some basic tests with Validation and such so that I don't have to?

timothyfroehlich:
@iamralpht @rylin8 

I haven't been able to figure out how to fix that layout issue in the screenshot diff, but there's been changes to some of the row and column APIs: https://developer.android.com/jetpack/androidx/releases/compose-foundation#1.5.2

I'm thinking that whatever fix we apply here won't matter with the upcoming layout changes, so I feel like this shouldn't block a merge.

rylin8:
> @iamralpht @rylin8
> 
> I haven't been able to figure out how to fix that layout issue in the screenshot diff, but there's been changes to some of the row and column APIs: https://developer.android.com/jetpack/androidx/releases/compose-foundation#1.5.2
> 
> I'm thinking that whatever fix we apply here won't matter with the upcoming layout changes, so I feel like this shouldn't block a merge.

Are you seeing what's shown on the right of the image when you run? Those are using `LazyVerticalGrid` and `LazyHorizontalGrid`. The new layout stuff shouldn't affect how those look since the contents of the list widget are not handled by the layout manager, but by the `LazyVerticalGrid` composable.

PR #295, opened 2023-08-18
Overhaul GitHub Actions

timothyfroehlich:
I've been working on these in the background for a bit and there's a number of related changes that I've tested on my fork. 

- Chaining builds rather than duplicating work in seperate jobs. Brings down our cache use (which keeps our cache from being evicted so quickly) and leads to faster builds. (2-5 minutes faster for the main SDK build job

- Better cache control in general, setting non-presubmit jobs to not update the cache.

- Proper setup of Gradle Dependency Graph submission, whiich powers Dependabot

- Consolidation of the presubmit and nightly check jobs

- Consolidating lints into the main workflow

- Splitting off the "Experimental" jobs which test against future releases, are only run nightly and allowed to fail.

- Cleaning up some unnecessary build artifact zipping and unzipping.

- Also caught some formatting fixes.


rylin8 (approved):
Again, I don't follow everything but looks ok!

timothyfroehlich:
> Again, I don't follow everything but looks ok!

Thanks!

PR #314, opened 2023-08-24
Move layout code into Rust

rylin8:


johnoneil:
I only have superficial feedback on this so far, but it has been helpful in my starting to understand how we will use the Rust code here in our pure rust client code to manage layout.

In the layout crate the primary thing I'd like to see would be a pure Rust example that shows a kind of example integration. That is, how to use in a pure Rust environment.

I imagine we'd have to provide functions that stand in where existing `.kt` currently provides information, for example with text geometry.

Currently we have a (very basic, and static) layout implementation which is struggling with how to manage variants, so despite the difficulty here this code is very much needed. This is much appreciated Rob.

rylin8:
Honestly I haven't fully thought through how to integrate this with Rust and would like to have a conversation about that soon. However I have been testing the layout code purely in Rust in a number of scenarios and plan to add a bunch of unit tests. This should be a good starting point on how to use it. From there I'll try to add another level of abstraction for the Rust code to use, similar to the JNI interface used by Android.

iamralpht:
@rylin8 I think the `layout` crate is missing; it's mentioned in the Cargo workspace decl, but isn't in the PR.

rylin8:
> @rylin8 I think the `layout` crate is missing; it's mentioned in the Cargo workspace decl, but isn't in the PR.

Currently the layout crate isn't used. I made it initially and put all the layout code in it, then decided to use the live update crate because at the time, layout was tied to live update. Now I need to move it back to the other crate, after which I'll add the code back into this PR.

iamralpht:
> This is too big for me to really be able to review, but for the JNI stuff, it'll be best to only have one library / one interface, rather than two. I would split the crates such that one crate (currently live_update) contains all of the Android/JNI related code, and the other crate (layout?) contains all of the shared code.

Yeah, that makes sense. I think if the library is being used for native targets then developers will come in looking from a native perspective -- given that, I think it would also be good to rename `live_update` to `jni` since it contains glue logic and not functional logic relating to the live update feature.

rylin8:
> > This is too big for me to really be able to review, but for the JNI stuff, it'll be best to only have one library / one interface, rather than two. I would split the crates such that one crate (currently live_update) contains all of the Android/JNI related code, and the other crate (layout?) contains all of the shared code.
> 
> Yeah, that makes sense. I think if the library is being used for native targets then developers will come in looking from a native perspective -- given that, I think it would also be good to rename `live_update` to `jni` since it contains glue logic and not functional logic relating to the live update feature.

Done!

iamralpht:
I tried this out a bit on an emulator with 1.0 density.
 * In general -- initially I got a lot of out-of-memory exceptions, but I didn't try tracking them down. Somewhere we're making a huge allocation.
 * **Alignment**: the text isn't aligned correctly; I can take a look at this today, or next weekend.
 * **Battleship**: the board is getting stretched in a weird way; I think this might be the parent layout not having the expected values (I definitely saw this design fail like this before in the Rust toolkit).
 * **HConstraints** and **VConstraints**: these aren't stretching out, because I think the incoming `Modifier.fillMaxSize()` is currently ignored.
 * **Cross Axis Fills** looks correct now (and incorrect before)!
 * **Fill Container Test** looks correct (but not sure it's getting stretched).
 * The various interaction and visual (non-layout) validations look fine.
 * **Fancy Fills** still seems to crash the emulator (it was before, don't know why yet, no backtrace).

So overall most tests are working. I think making the root stretch/resize to parent by default will either reveal some other layout issues or prove that it's behaving as expected. Text metrics are pretty challenging to get right, and the current implementation is quite complicated because at the time it was written, Compose always included some additional vertical padding on text. Compose changed the text measurement logic in 1.2.0 (hence this confusing "Added in 1.2.0, Deprecated in 1.2.0" note on [PlatformTextStyle#includeFontPadding](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/PlatformTextStyle)), so we might be able to remove our measurement logic which only existed to remove the extra padding since we only target Compose > 1.2.0.

iamralpht:
@rylin8 on the most recent change for vertical alignment, how does `text_size` relate to line height? (You can change text size and line height within a single "Text" node, and we put some of that stuff in the text run structure).

rylin8:
> @rylin8 on the most recent change for vertical alignment, how does `text_size` relate to line height? (You can change text size and line height within a single "Text" node, and we put some of that stuff in the text run structure).

`text_size` is just the size of the text bounding box in Figma -- it doesn't have anything to do with the line height. The bounding box of a text node in Figma often had a smaller height than the height measured in Compose, so I had to take that into account when calculating layout and where to vertically position the text.

iamralpht:
> `text_size` is just the size of the text bounding box in Figma -- it doesn't have anything to do with the line height. The bounding box of a text node in Figma often had a smaller height than the height measured in Compose, so I had to take that into account when calculating layout and where to vertically position the text.

Got it, yes, that makes sense. Android has really beautiful text layouts, but they tend to be a tiny bit wider than what harfbuzz generates (at least, when we used harfbuzz directly for text layout it seemed to perfectly match what Figma generates itself). I think it could be a glyph offset rounding thing, but I don't know Android's text implementation well enough.

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -214,6 +214,21 @@ fn compute_layout(node: &Node, parent: Option<&Node>) -> ViewStyle {
                     }
                 }
             }
+
+            if frame.layout_mode != LayoutMode::None {
+                let width_points = bounds.width().ceil();
+                let height_points = bounds.height().ceil();
+                style.width = match frame.layout_sizing_horizontal {
+                    LayoutSizing::Fixed => Dimension::Points(width_points),
+                    LayoutSizing::Fill => Dimension::Points(width_points),
```
iamralpht, line 223: Does it work if you just set min_width / min_height? (You probably tried that...).

rylin8: Nope that doesn't work, that makes the size grow too large also. I'm thinking I may have to make autolayout frames with children that have a FILL sizing mode to somehow ignore the children when calculating layout size.

iamralpht: Hm, ok, this is another dumb idea, but what are the default settings for flex and align-items? I'm wondering if there's some default style value that's set inappropriately.

rylin8: flex grow/shrink both default to 0, and align-items defaults to Stretch. However align-items should get set to something else if it is a frame, based on the value of counter_axis_align_items. I can dig it into it more and try and reproduce with a flexbox example somewhere else.

timothyfroehlich:
@rylin8 Are we ready to close this PR since we're working off the new feature branch?

rylin8:
Merged into feature/rustlayout

Issue #315, opened 2023-08-24
Paths not visible when children of Groups


johnoneil:
I've been struggling on this issue for a couple of days, so I'll try to describe it methodically. I think I've got it pinned down, but I could be wrong.

I'm currently working on a fix and may present that shortly. I'm open to any questions or suggestions to debug/further explore this issue.

it's also possible there may be a fix on the client side. Not 100% sure yet. I'll take that option if possible.

I believe this issue may be at least partially related to issue #244 . A fix for this issue may also help clarify that (and hopefully not break things!).

First, this is the Figma element we're dealing with. There is a parent node, then a child group (g24) which is offset from the parent, and then a bunch of child paths.

![Screenshot from 2023-08-24 11-29-07](https://github.com/google/automotive-design-compose/assets/2950232/22bf551d-43f2-4397-9c65-404c0cefa898)

When I render this on the client side, using the path absolute position for translation, and the (accumulated) parent transforms, we see an offset creeping into the paths. Like in the following image.

![Screenshot from 2023-08-24 11-26-57](https://github.com/google/automotive-design-compose/assets/2950232/8a1fe706-ecc8-4ee4-a4f4-5948602a4536)

As you can see from the image, the individual paths are drawn correctly, and their general arrangement is okay, but they are offset incorrectly.

I believe the origin of this issue is this section in `transform_flexbox.rs`

```
            let r = r.post_translate(euclid::vec3(
                -(bounds.x() - parent_bounds.x()),
                -(bounds.y() - parent_bounds.y()),
                0.0,
            ));
```
This section tries to remove translation from the `relative_transform` matrix provided by figma. That is, it removes (leading minus signs) the offset from the local rect bounds from its parent.

However there are cases when this will be incorrect. I'm referring to this doc here: 

https://www.figma.com/plugin-docs/api/properties/nodes-relativetransform/#:~:text=The%20relative%20transform%20of%20a,group%20or%20a%20boolean%20operation.

Note the section: 

```
The relative transform of a node is shown relative to its container parent, which includes canvas nodes, frame nodes, component nodes, and instance nodes. Just like in the properties panel, it is not relative to its direct parent if the parent is a group or a boolean operation.
```

So in our case, since the immediate parent is a Group (`g24`) the translation data present in the `relative_transform` matrix will not be entirely removed. Instead we should be "removing" translation data from the (further offset!) goup g24's parent.

[...]

PR #322, opened 2023-08-25
#315 Include original relative_transform in serialized output.

johnoneil:
This is to address the issue I saw and filed as #315.

The primary change is to include the original unaltered `relative_transform` in the serialized figma output. Since this addition is merely additive, that change should not break anything.

~~**However** I'm also making a correction to the `transform` calculation which I believe to be correct, but which could break downstream clients. Maybe we can talk about how to test this.~~

The addition of the original `relative_transform` in the serialzied document has allowed me to decrease the complexity of of our renderer (slighly) and also fixes the original issue. I'll attach a screenshot of our (non-flattened) telltales below.

Screenshot from 2023-08-24 23-37-05

### DIFF
--- a/crates/figma_import/src/toolkit_style.rs
+++ b/crates/figma_import/src/toolkit_style.rs
@@ -496,6 +496,7 @@ pub struct ViewStyle {
     pub stroke: Stroke,
     pub opacity: Option<f32>,
     pub transform: Option<LayoutTransform>,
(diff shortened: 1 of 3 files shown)


rylin8:
Interesting! @johnoneil do you have a Figma doc that you've been using to test this?

rylin8 (approved):

PR #324, opened 2023-08-28
Fix #176: Add support for non-uniform stroke widths on rects and frames.

### LINKED ISSUE #176: Setting Figma stroke for individual side will render in compose for all sides
render_compose
outline_bottom_figma

iamralpht:
This change updates the file format by adding non-uniform stroke widths.

The Kotlin renderer interprets non-uniform stroke widths for frames and rects. Paths and arcs are rendered with a uniform stroke width as before. The Figma UI only exposes non-uniform stroke widths for frames and rects, so I think this change implements the complete feature.

Tested:
 * The shadows validation page entirely uses frames, and generates the same image after this change. We might want to extend the shadows test to exercise paths.
 * A new "variable borders" test doc has been added which exercises various shapes that the variable borders feature can create. There is some special handling of the border radius offsetting for zero width strokes meeting non-zero strokes.


iamralpht:
@rylin8 how do I rebase all of the serialized files?

rylin8:
> @rylin8 how do I rebase all of the serialized files?

Rebase? Don't you just need to regenerate them? I did that by going through all the validation doc examples then pulling them off the device using Android Studio's device explorer. Are you going to bump the version up from 15 to 16? If we get this in before the next release you probably don't have to.

designcompose/src/main/java/com/android/designcompose/Shape.kt
```diff
@@ -0,0 +1,620 @@
+package com.android.designcompose
```
rylin8, line 1: Need to add that copyright blurb here

designcompose/src/main/java/com/android/designcompose/Shape.kt
```diff
@@ -0,0 +1,620 @@
...
+
```
rylin8, line 2: And maybe call this file something that indicates it does path rendering, like ShapeRender or something?

designcompose/src/main/java/com/android/designcompose/Shape.kt
timothyfroehlich: Just run ./dev-scripts/format-all.sh

designcompose/src/main/java/com/android/designcompose/Shape.kt
```diff
@@ -0,0 +1,620 @@
...
+    /// is based on the stroke bounds, and not the background bounds.
+    val shadowClips: List<Path>,
+
+    /// The paths that are filled for shadows. The shadow style field is
+    /// referenced, which includes details like the inset/outset of the
+    /// shadow, its offset, and its color.
+    val shadowFills: List<ComputedShadow>,
+)
+
+class ComputedShadow(val fills: List<Path>, val shadowStyle: BoxShadow)
```
timothyfroehlich, line 54: Data class?
```suggestion
data class ComputedShadow(val fills: List<Path>, val shadowStyle: BoxShadow)
```

timothyfroehlich (commented):
Looks fine as far as I can tell. I may have a PR comming in the next day or so that'll make it easier to update the serialized files. (Not fetch, but testing to support it), so you might want to wait for that before updating them.

designcompose/src/main/java/com/android/designcompose/Shape.kt
iamralpht: Yeah, OK. I always forget that `data class` exists. We don't compare these or dump them to string anywhere, but I'll see if it applies (to this and ShapePaths).

designcompose/src/main/java/com/android/designcompose/Shape.kt
iamralpht: Yeah, good point. I think `ComputePaths` is a good name for the file because `computePaths` is the only publicly exported method. I should probably rename `ShapePaths` to `ComputedPaths` which sounds more like a derived value.

designcompose/src/main/java/com/android/designcompose/Shape.kt
iamralpht: Ok, I've added the copyright blurb to the new file using the `./dev-scripts/format-all.sh` tool :).

designcompose/src/main/java/com/android/designcompose/Shape.kt
iamralpht: Done. I think this makes it a lot clearer what it's doing and why you might need it.

iamralpht:
I addressed most of the feedback, but I noticed that vector content is not rendering correctly after rebasing on main. I'm going to track down if that's an issue in main or a regression in this PR next (although I know Battleship was rendering correctly before, and now it's a big mess, so I'm guessing it's a regression in main).

iamralpht:
> I addressed most of the feedback, but I noticed that vector content is not rendering correctly after rebasing on main. I'm going to track down if that's an issue in main or a regression in this PR next (although I know Battleship was rendering correctly before, and now it's a big mess, so I'm guessing it's a regression in main).

Filed #344 and made PR #345 -- this change also includes the same fix. So we can merge one or the other ahead of the next release.

rylin8 (approved):

designcompose/src/main/java/com/android/designcompose/Shape.kt
iamralpht: Ok, I didn't do this in this change because we don't compare these and not all of the contents are directly comparable in that way (but will try to remember `data class` for new objects that need comparisons & eqality).

PR #342, opened 2023-09-01
Fix-forward PR #322, add documentation for doing so

timothyfroehlich:
Update design switcher with a new seriallized file.


iamralpht:
Is this broken due to serialized file breakage?

timothyfroehlich:
I've tried updating the serialized file but it didn't help.

iamralpht:
Ok, on top-of-tree, I got a new serialized file (by flipping the `DISABLE_LIVE_MODE` flag in DesignSwitcher.kt, and I got the design switcher working OK, although some of the vector content seems incorrect (see the close icon on the right in this screenshot).

@timothyfroehlich is this the breakage you're talking about, or it was not working at all on top-of-tree when you fetched an updated doc?

broken close icon

rylin8:
> That's odd, I just built and ran top of tree, enabled live mode for the design switcher and it looks like this:
> 
> Screenshot 2023-09-04 at 6 16 35 PM

Oh maybe you have a different density

iamralpht:
> Oh maybe you have a different density

Yeah, I was using an emulator target created from the "Nexus 10" template; I think it has a density somewhere between 2 and 3 (but haven't logged it).

iamralpht:
@timothyfroehlich it looks like this no longer contains the revert -- can you update the PR title if that's true?

timothyfroehlich:
Ok, so `DISABLE_LIVE_MODE` wasn't documented and I didn't know about it. I've added documentation for it now, pulled down the new version and confirmed that it works. I'll try to add a test for the Design Switcher soon.

timothyfroehlich:
> @timothyfroehlich it looks like this no longer contains the revert -- can you update the PR title if that's true?

In the middle of that.

timothyfroehlich:
@iamralpht this is good to go.

rylin8 (approved):

PR #345, opened 2023-09-04
Fix #344: vector scale is wrong when density isn't 1.0.

### LINKED ISSUE #344: Regression of vector rendering when density != 1.0
The fix for scaled layout elements introduced a regression when density isn't 1.0; elements appear too large or small (scaled by the density). Marking P0 because this is quite a visible regression.

iamralpht:


iamralpht:
I'll also put this change into the non-uniform border width change, then we can pick this one or that one on Tuesday.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -544,8 +544,8 @@ internal fun ContentDrawScope.render(
         vectorSize.ifPresent {
             val sizeList = vectorSize.get()
             if (sizeList.size == 2) {
-                val vecWidth = sizeList[0] * vectorScaleX
-                val vecHeight = sizeList[1] * vectorScaleY
+                val vecWidth = sizeList[0] * vectorScaleX * density
+                val vecHeight = sizeList[1] * vectorScaleY * density
```
rylin8, line 548: Good catch, thanks. In the new layout system I'm working on, the size of the draw context is determined by layout, and the size to draw is determined by the size of the vector from Figma. I'll have to think about if this will need to change again in the new system. I'll test this change out on Tuesday.

iamralpht: Yeah, I think it will; I'm making an emulator with dpi 160 to test out the layout change right now...

rylin8:
I approved https://github.com/google/automotive-design-compose/pull/324 instead of this one

iamralpht:
Yep, closing this PR since #324 merged, and it includes the same fix.

PR #352, opened 2023-09-06
Enhance DesignCompose testability

timothyfroehlich:
This adds a new field to DesignSettings which records when and how
a DesignDoc was last fetched. This allows for instrumented tests to
confirm that the file be rendered was fetched by Figma.

This also adds a Semantics tag that is applied to the DesignDoc's root DesignView.
This allows a tester to tell whether a DesignDoc is being rendered

A test fixture was added which exposes a method for enabling
live update within a test via a key provided via test arguments. The key
is not embedded in the app this way. At this time AGP doesn't support
compiling Kotlin test fixtures (long story) so it's written in Java.

I ended up cleaning up DesignCompose's dependencies as well, there should be nothing affected by this.

TO TEST:

We do not have a way to run Instrumentation tests or Figma fetches via GitHub actions. The tests added here must be performed manually. 

To watch and/or debug the tests: Launch an emulator and run `./gradlew connectedTest`

To run the tests automatically using automated test devices: run `./gradlew tabletAtdApi30DebugAndroidTest`
Edit: After #363 merges that will be `./gradlew gmdTestQuick`

To run the full test suite: run `./dev-scripts/test-all.sh`


designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -321,6 +324,10 @@ internal data class MaskInfo(
     val type: MaskViewType?,
 )
 
+// Custom Compose Semantic that identifies a DesignCompose Composable
+val docIdSemanticsKey = SemanticsPropertyKey<String>("DocId")
+var SemanticsPropertyReceiver.sDocId by docIdSemanticsKey
```
iamralpht, line 329: I spent far too long reviewing these two lines because I wasn't familiar with the `var x by y` syntax, but I think this is just establishing `sDocId` as a global variable (so it's actually not very complicated, just very compact).

timothyfroehlich: Yeah, it's kind of wild. I'm not sure I fully grasp what happens there, it's code from https://developer.android.com/jetpack/compose/testing#custom-semantics-properties

designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -66,7 +67,32 @@ internal class LiveDocSubscriptions(
...
 )
 
+/**
+ * Design doc status
+ *
+ * Publicly accessible status for a DesignDoc. Accessible via DesignSettings.designDocStatuses
+ */
+class DesignDocStatus() {
+    // If set, the time when the doc was last loaded from storage. Otherwise, this has not happened
+    var lastLoadFromDisk: Instant? = null
```
iamralpht, line 77: Do we need times for these? Would a count work OK?

designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -66,7 +67,32 @@ internal class LiveDocSubscriptions(
     val subscribers: ArrayList<LiveDocSubscription> = ArrayList(),
 )
 
+/**
+ * Design doc status
+ *
+ * Publicly accessible status for a DesignDoc. Accessible via DesignSettings.designDocStatuses
```
iamralpht, line 73: Should this be a new public API, or only available to tests? (I think there might be some annotations to mark something as accessible during a compilation for tests only). I would not want to see production logic get built on top of `isRendered` etc because it'll be fragile if the doc ID changes (we added the callback API for that particular use case).

designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -314,6 +340,11 @@ internal fun DocServer.fetchDocuments(
 
                 // Remember the new document
                 synchronized(documents) { documents[id] = doc }
+                synchronized(DesignSettings.designDocStatuses) {
```
iamralpht, line 343: I don't think we're consistently locking `designDocStatuses` when updating it; I don't think reads from `designDocStatuses` by client code are synchronized at all (unless Kotlin adds that invisibly).

designcompose/src/main/java/com/android/designcompose/DocServer.kt
timothyfroehlich: Hm, yeah, maybe I should make this test-only.

Is there a way to get an immutable reference to a DesignDoc?

designcompose/src/main/java/com/android/designcompose/DocServer.kt
timothyfroehlich: Kotlin adds that, but yeah, I'm not sure, it seemed like a better safe than sorry thing since we're mixing threads and coroutines.

designcompose/src/main/java/com/android/designcompose/DocServer.kt
iamralpht: ok, that makes sense (but can be an internal API in that case)

designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -416,6 +448,7 @@ internal fun DocServer.doc(
 
     // Create a state var to remember the document contents and update it when the doc changes
     val (liveDoc, setLiveDoc) = remember { mutableStateOf<DocContent?>(null) }
+    DesignSettings.designDocStatuses.putIfAbsent(docId, DesignDocStatus())
```
rylin8, line 451: Cool didn't know about `putIfAbsent()`. So you are populating an empty entry for `docId` here, and then in `fetchDocuments()` the data is filled out, is that correct? When, if ever, would `DesignSettings.designDocStatuses[id]` be null there?

designcompose/src/main/java/com/android/designcompose/DocServer.kt
rylin8: I think Ralph is saying that the other places where you set variables in `DesignSettings.designDocStatuses[id]` are not synchronized.

designcompose/src/main/java/com/android/designcompose/DocServer.kt
timothyfroehlich: I'm not sure, are you asking about the safe calls below? I'm getting errors if I don't make them safe calls

designcompose/src/main/java/com/android/designcompose/DocServer.kt
timothyfroehlich: I think the `isRendered` flag is _probably_ redundant vs checking for the `docIdSemanticsKey`, so I'll just remove it.

designcompose/src/main/java/com/android/designcompose/DocServer.kt
timothyfroehlich: I see what he means now, I read it too quickly. I'm still not happy with how it's set up, we should talk offline about it.

designcompose/src/main/java/com/android/designcompose/DocServer.kt
rylin8: Yes I'm referring to all the safe calls `designDocStatus[docId]?.` but I'm not asking you to remove them. I'm just wondering if you expect them to ever be null and not set anything?

designcompose/src/main/java/com/android/designcompose/TestUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+package com.android.designcompose
+
+import com.google.common.annotations.VisibleForTesting
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.flowOf
+
+/**
+ * For use in tests only
+ *
+ * Enable Live Update using a given Figma Token This skips the LiveUpdateSettings setup and
```
rylin8, line 26: Missing a period here. I'm confused about this and TestUtils.java. Are they supposed to be named the same?

designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -413,9 +443,13 @@ internal fun DocServer.doc(
     }
 
     val id = "${resourceName}_$docId"
+    Log.d("froehtdebug", "$id recomposed")
```
rylin8, line 446: debug line

designcompose/src/main/java/com/android/designcompose/DocServer.kt
timothyfroehlich: It could be. I've made the setters internal so it's essentially read only. Do you think it's worth hiding that information?

designcompose/src/main/java/com/android/designcompose/TestUtils.kt
timothyfroehlich: They don't need to be named the same. This file is specifically holding off a test only function that must be part of the package because it accesses internal variables. It can't be part of the TestUtils.java, because that is part of the new test fixture, and they're a seperate compile unit so they can't access internal variables

rylin8 (approved):

cmwarrington (commented):
The AGP related changes here look good

timothyfroehlich:
Thank you!

PR #343, opened 2023-09-01
Add tests to fetch and render all docs in Validation

timothyfroehlich:
This uses the test fixtures added in the base CL and generates tests for all entries in Validation's examples list, skipping any that don't use a Figma Doc (one of them doesn't).

TO TEST:

We do not have a way to run Instrumentation tests or Figma fetches via GitHub actions. The tests added here must be performed manually. 

To watch and/or debug the tests: Launch an emulator and run `./gradlew connectedTest`

To run the tests automatically using automated test devices: run `./gradlew tabletAtdApi30DebugAndroidTest`

To run the full test suite: run `./dev-scripts/test-all.sh`

PR #358, opened 2023-09-11
Update Validation Serialized files

timothyfroehlich:
Files generated from the tests from #343


rylin8:
Hmm these changed recently from @iamralpht 's PR https://github.com/google/automotive-design-compose/pull/324. Not sure why they would be different again.

rylin8:
Did you find that this wasn't needed?

timothyfroehlich:
It was, but I merged it into #365

PR #386, opened 2023-09-22
Remove repeated ProxyConfig Log print

timothyfroehlich:
Seems unnecessary to print it for every fetch attempt.

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/DocServer.kt
+++ b/designcompose/src/main/java/com/android/designcompose/DocServer.kt
@@ -308,7 +308,6 @@ internal fun DocServer.fetchDocuments(
     }
 
     val proxyConfig = getProxyConfig()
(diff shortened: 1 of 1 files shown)


designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -289,7 +289,6 @@ internal fun DocServer.fetchDocuments(
     }
 
     val proxyConfig = getProxyConfig()
-    Log.i(TAG, "HTTP Proxy: ${proxyConfig.httpProxyConfig?.proxySpec ?: "not set"}")
```
urpatil, line 292: The proxy config is set in Android using adb and hence can be changed when the app is running. 
 This is also repeated just like the "Live update for <doc> unchanged..." because it can change.

 This is useful for debugging purposes.  One way to make this less verbose is to remember the previous value and only print when it changes.

PR #424, opened 2023-10-04
Add cargo deny to our presubmits

timothyfroehlich:
Inspired by ureq [doing the same](https://github.com/algesten/ureq/pull/661)

This will help us keep track of our dependencies and make sure we aren't using ones with the wrong licenses.


iamralpht (approved):
nice - thanks for adding this, I really like `cargo deny` for ensuring ongoing compliance.

PR #425, opened 2023-10-04
Remove Rocket from our dependencies

timothyfroehlich:
It's not being used by anything and is causing cargo-deny issues

### DIFF
--- a/crates/figma_import/README.md
+++ b/crates/figma_import/README.md
@@ -1,4 +1,5 @@
 # figma_import
+
 =======
 
 This library (figma_import) and binaries perform conversion from a Figma document to a self-contained serialized doc that can be interpreted and rendered by the DesignCompose library.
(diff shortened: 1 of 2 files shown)


crates/figma_import/README.md
```diff
@@ -8,25 +9,28 @@ The Figma API provides access to documents, images, vectors, and interactions th
...
+* `image_context` manages all of the images in the document. Figma's API takes a big set of "image references" and then returns a document containing URLs for each of those image references. `image_context` then fetches the images, and stores the network bytes and dimensions. It can encode and decode its list of fetched images and sizes -- this is how the list of previously fetched images is sent to the client and reused for the next request (avoiding having `image_context` fetch images that the client already has).
+  * `svg` is used by `document` and `image_context` to rasterize complex vectors to rasters (storing the resulting rasters as images in `image_context`). Currently vectors get rasterized at 1x, 2x, and 3x sizes so that the DesignCompose client can choose the best image based on the device's pixel density. Figma provides an API to rasterize vectors, which we don't use because it clips the edges off of rasterized vectors.
```
iamralpht, line 32: not sure why this is in the diff, but it's also wrong (the client rasterizes vectors in all client impls now).

iamralpht (approved):
good to remove rocket, nice to clean up the README or remove the bits that didn't change from this diff.

crates/figma_import/README.md
timothyfroehlich: I think it got run through my formatter. I'll raise an issue to get the readme updated.

timothyfroehlich:
I'll leave the formatted changes. They match the format that we use for the rest of the docs.

PR #430, opened 2023-10-05
Begin recording build profiles

timothyfroehlich:
I've been wanting to track our build times and see how they improve. This will add a CI job that will build everything on debug and build the library, with and without cache, with Gradle's profiling enabled.


iamralpht (approved):
Can't wait to start tracking build times -- nice add!

PR #452, opened 2023-10-12
Begin outputting individual files for each @DesignDoc

timothyfroehlich:
This change will will provide a significant reduction in compile time for packages containing a large number of @DesignDocs. With this change, KSP will only need to regenerate files for the source files that contain a change, rather than regenerating the entire source if anything in the package change.


rylin8 (approved):
Would be nice to do the same for the output JSON files, though not high priority.

PR #465, opened 2023-10-14
Closes #404: Layout is wrong when pixel density < 1

### LINKED ISSUE #404: Layout is wrong when pixel density < 1
Oddly this seems to work with pixel density > 1 but when < 1 there are some issues mainly with text.

rylin8:
Fix text render height calculation. The render height is calculated taking into account pixel density, but when rendering should divide the pixel density back out of the value.


designcompose/src/main/java/com/android/designcompose/DocServer.kt
iamralpht, line 1: I don't think there are any changes here -- I wonder why the formatting is different?

iamralpht (approved):
ok to merge into the feature branch

PR #470, opened 2023-10-16
[Rust-layout] Fix app builds

timothyfroehlich:


codegen/src/main/kotlin/com/android/designcompose/codegen/BuilderProcessor.kt
```diff
@@ -135,6 +135,7 @@ class BuilderProcessor(private val codeGenerator: CodeGenerator, val logger: KSP
         file += "import com.android.designcompose.setVisible\n"
         file += "import com.android.designcompose.TapCallback\n"
         file += "import com.android.designcompose.ParentComponentInfo\n"
+        file += "import com.android.designcompose.ParentLayoutInfo\n"
```
rylin8, line 138: I think this needs `ReplacementContent` also. Does Validation build without it? Can you test building/running Validation to make sure?

timothyfroehlich: Doesn't look like it, I was able to run `./gradlew assemble` and everything built.

rylin8 (approved):

PR #472, opened 2023-10-16
Closes #471: Fix live update not updating positional changes

### LINKED ISSUE #471: Live update does not update fixed positional changes
When a live update occurs where nodes change position within a non-autolayout parent frame, the parent frame is not included in the list of changed nodes, so the parent isn't recomposed. Since node positions are performed in a parent frame's layout() function, the node that moved doesn't update its position.

rylin8:
When computing layout and determining nodes that changed, include the parent of the node as well. For text nodes, recompute text layout when the text style changes.

### DIFF
--- a/crates/layout/src/layout.rs
+++ b/crates/layout/src/layout.rs
@@ -108,9 +108,10 @@ impl LayoutManager {
     fn update_layouts(&mut self, taffy: &Taffy) -> Vec<i32> {
         fn update_layout(
             layout_id: i32,
+            parent_layout_id: i32,
             manager: &mut LayoutManager,
             taffy: &Taffy,
-            changed: &mut Vec<i32>,
+            changed: &mut HashSet<i32>,
         ) {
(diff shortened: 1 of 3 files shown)


iamralpht (approved):

crates/layout/src/layout.rs
```diff
@@ -108,9 +108,10 @@ impl LayoutManager {
     fn update_layouts(&mut self, taffy: &Taffy) -> Vec<i32> {
         fn update_layout(
             layout_id: i32,
+            parent_layout_id: i32,
             manager: &mut LayoutManager,
             taffy: &Taffy,
-            changed: &mut Vec<i32>,
+            changed: &mut HashSet<i32>,
```
iamralpht, line 114: I'm guessing the Kotlin side doesn't care what order the invalidations are in, so it's OK to move from a Vec to a Set (and avoid duplicate parents...).

rylin8: Yea it doesn't -- it just notifies them all in one pass.

Issue #490, opened 2023-10-20
Don't recompute layout when removing children of lazy grid view


rylin8:
While scrolling through a lazy grid view, whenever a row or column goes out of view it gets removed from composition. This triggers unsubscribe() to get called from onDispose(), and unsubscribe() will trigger a layout recompute. So if you scroll a grid view and 5 items disappear, and each item has 8 nodes, then 40 recomputes will be called. This adds up fast when you scroll quickly. None of these recomputes should need to happen since the child of the lazy grid view is gone from view. Figure out a way to remove these recomputes.

Issue #488, opened 2023-10-19
Reduce layout computation when text nodes are measured


rylin8:
Text nodes are measured in Android, and then their size is sent to Rust layout. Every time this happens, a layout computation happens. If a view has X text nodes, all X text nodes will be measured and individually sent to Rust layout, and layout will be computed X times. Find a way to optimize this.

Issue #487, opened 2023-10-19
Cache layout data in Android


rylin8:
When recomputing layout, we currently only send the layout IDs of nodes that changed from Rust to Android. Then when Android needs to get layout info, it calls into Rust via JNI to get the layout. We can speed this up by sending the layout data initially whenever it changes, where it can be cached in Android and accessed without needing to make a JNI call.

PR #491, opened 2023-10-20
Closes #486, #487, #488, #490: Performance improvements

### LINKED ISSUE #486: Reduce JNI add-view calls
Every time a view subscribes to layout, a JNI call is made to add it to the view to the Rust layout system. Since we now have logic to only perform layout when all of a node's children have been added to the layout system, we can wait until this time to add all the views as well, saving a large number of JNI calls.

rylin8:
When subscribing to layout, add the node to a list of layout nodes instead of adding them directly via a JNI call. When we detect we are done with layout and need to compute it, make one JNI call to add all the views and compute layout on the root node.

The function to compute layout now needs the layout id of the node to compute, instead of computing layout on all root level nodes. This helps performance by not computing layout on nodes that we know have not changed.

When computing layout in Rust, send back not only the layout IDs that changed, but all the layouts as well. The layout manager in Android now caches these layouts, so retrieving them no longer requires a JNI call.

Change layout subscriptions to add ViewStyle instead of View
Subscribing to layout with a View caused unnecessary serialization of data. In particular, serializing a View also serialized all of a View's children, meaning we were serializing O(n^2) data that wasn't even used. Since the Rust layout only uses the style within View, and the name for debugging purposes, it now takes those as parameters.

When subscribing to layout, use the ViewStyle created from merging with overrideStyle instead of the style that came with the View originally. This fixes the style used for variants and replacement nodes and means we no longer need the "baseView" mechanism of overriding styles.

Add another "rootLayoutId" parameter to ReplacementContext for ContentReplacement customizations. Replacement customizations should use this in the "parentLayout" parameter for the replacement content.

After computing layout for a specific layout_id, calculate layouts that changed for only the specified layout_id and its children instead of all root nodes.

Reduce layout computation when text nodes are measured. Simplify and improve text measurement and layout by measuring text before subscribing to layout, and then sending the text size when subscribing. This is an improvement in a few ways:
- Every text node visible no longer triggers a layout recomputation since they now subscribe in the same manner as frames where layout computation is done after all nodes have subscribed
- Text is laid out at the same time as frames, avoiding the problem of text flickering

Don't recompute layout when removing nodes that are ancestors of a list widget. Nodes are frequently removed from lazy grid views as they scroll and no layout needs to be recomputed since the lazy grid view handles layout itself.

Add another optimization when removing nodes by only computing layout for the root node of the node removed.

### DIFF
--- a/crates/figma_import/src/layout.rs
+++ b/crates/figma_import/src/layout.rs
@@ -19,7 +19,7 @@
 //! The goal of this crate is to perform the mapping from Figma to the toolkit; it does
 //! not provide any kind of UI logic mapping.
 
-use crate::toolkit_schema;
+use crate::{toolkit_schema::Layout, toolkit_style::ViewStyle};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
@@ -28,7 +28,7 @@ use std::collections::HashMap;
 #[derive(Serialize, Deserialize, Debug)]
 pub struct LayoutChangedResponse {
     pub layout_state: i32,
-    pub changed_layouts: HashMap<i32, toolkit_schema::Layout>,
+    pub changed_layouts: HashMap<i32, Layout>,
 }
 impl LayoutChangedResponse {
     pub fn unchanged(layout_state: i32) -> Self {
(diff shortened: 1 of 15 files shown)


rylin8:
Found an issue with variants rendering in the wrong position.

timothyfroehlich:
Sounds great! I'd like to see this with the screenshot tests in #500, just to make sure this didn't change any behavior.

iamralpht (approved):
These changes are good to merge to `feature/rustlayout`; I've also been using them in the squoosh-3 branch, so I feel like I got fairly familiar with the JNI interface from that work. I do think it would be good to have an overall object (like the root or something) that is a `taffy` instance behind the scenes so that multi threaded usage is clearer.

rylin8:
The small screenshot diff is fixed in https://github.com/google/automotive-design-compose/pull/524

PR #492, opened 2023-10-23
WIP: squoosh down DesignCompose kotlin renderer a bit.

iamralpht:
I made this super stripped down version of the DesignCompose renderer to make profiling easier. It doesn't support text or interactions (text will be needed for profiling soon...). I'm not sure if it makes sense as a direction to take the library in, or if it will only be useful for profiling the new layout work.


iamralpht:
abandoned for squoosh-3

PR #494, opened 2023-10-24
Change rendering tests to detect semantic on the root DesignDoc 

timothyfroehlich:
Targeting main with this PR instead of #459 which targeted the feature branch.

This provides a more consistent method of detecting and testing DesignCompose docs and whether they rendered.

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/DesignView.kt
+++ b/designcompose/src/main/java/com/android/designcompose/DesignView.kt
@@ -941,7 +941,6 @@ internal fun DesignDocInternal(
         }
     }
 
-    Box(modifier = Modifier.semantics { sDocRenderStatus = docRenderStatus })
 
     var noFrameErrorMessage = ""
     var noFrameBgColor = Color(0x00000000)
@@ -950,10 +949,12 @@ internal fun DesignDocInternal(
     if (doc != null) {
         val startFrame = interactionState.rootNode(rootNodeQuery, doc, isRoot)
         if (startFrame != null) {
+
+    Box(modifier = Modifier.size(0.dp).semantics { sDocRenderStatus = docRenderStatus })
             LaunchedEffect(docId) { designComposeCallbacks?.docReadyCallback?.invoke(docId) }
             CompositionLocalProvider(LocalDesignIsRootContext provides DesignIsRoot(false)) {
                 DesignView(
-                    modifier,
+                    modifier.semantics { sDocRenderStatus = docRenderStatus },
                     startFrame,
                     variantParentName,
                     docId,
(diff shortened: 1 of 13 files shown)


rylin8:
Confused -- doesn't the screenshot diff show a regression?

timothyfroehlich:
https://github.com/takahirom/roborazzi-compare-on-github-comment-sample/issues/6

rylin8:
Do you have a plan on merging this into the rustlayout branch?

timothyfroehlich:
> Do you have a plan on merging this into the rustlayout branch?

I'm just going to bring this in. The scope of this PR ended up getting narrowed. Originally this included a workaround for the rustlayout branch dropping the modifiers that were passed in, but like you saw yesterday, that caused layout regressions. So I've dropped that workaround and instead this PR is only for changing the semantics and how the tests detect the DesignDoc and whether it's rendering properly. The net result of this is just a better testing API, essentially.


Separately I'm going to finish up my change to get the libraries building for your host and hooked up to the tests, which will allow the screenshot tests to run (and fail). Roborazzi should be able to generate screenshots at that point, but if not then I'll push another change to disable the assertions. That way the test won't fail and Roborazzi should be able to grab the screenshots no problem.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -954,7 +951,7 @@ internal fun DesignDocInternal(
             LaunchedEffect(docId) { designComposeCallbacks?.docReadyCallback?.invoke(docId) }
             CompositionLocalProvider(LocalDesignIsRootContext provides DesignIsRoot(false)) {
                 DesignView(
-                    modifier.semantics { sDocId = docId },
+                    modifier.semantics { sDocRenderStatus = docRenderStatus },
```
rylin8, line 954: How will this part work with in the rustlayout branch?

timothyfroehlich: I don't know, and so I'm basically punting that. 

Basically, we don't have a solution for getting the test assertions to work yet. But I'm pretty sure I can get screenshot tests working either way.

I still think this is a good change. The changes that I made to the actual tests will persist to the rustlayout branch. Basically we'll need to figure out how to get the semantics applied in rustlayout, but the tests themselves won't need to change.

rylin8: Ok that's fine, you'll just have to deal with some conflicts merging to the feature branch or merging the feature branch back to main.

timothyfroehlich: Yep

rylin8 (approved):

PR #524, opened 2023-11-03
Closes #464: Improve grid view load and scroll performance

### LINKED ISSUE #464: List widgets load slowly
When loading a design with list widgets that have many items in them, the load time is quite slow. This is apparent in the media browse views. This is likely happening because each item in a list widget triggers a recompute. So after every item is added, a recompute occurs, where a recompute includes computing layout for the root node, all nodes in the queries() generated function, and every item in every list widget. Find a better way to optimize this.

rylin8:
When the lazy grid view first loads and when scrolling through it quickly, some of its children don't have their layout computed yet, resulting in a child that takes up no space, which makes the lazy grid view load more children. Sometimes at load time, all the children are loaded because of this.

This commit fixes this issue by having the list preview widget write size values for all nodes that can go in it. We parse this data out of the shared plugin data and then ultimately use it as the default/initial size of a node if its layout has not yet been computed.

Clean up some code relating to list widget
- Change list content customization so that the replaced https://github.com/composable content function includes parent layout info. This should be passed to the replacement content so that when it is composed, DesignFrame knows it is the child of a widget and can handle it differently.
- Remove unnecessary Box() under DesignFrameLayout()
- Clean up some unnecessary logging

Changed layout to use constraints. Our custom layout function now gets the layout for a node's children and uses each child's width and height to construct a fixed size constraint for it. This lets us remove the special case Box() for children of a Grid, where there were some layout oddities without this change.

Update serialized files and screenshots


crates/figma_import/src/extended_layout_schema.rs
```diff
@@ -45,6 +45,20 @@ pub struct SpanData {
     pub max_span: bool,
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GridNodeSize {
+    /// The node name of an item that goes into a grid layout
+    #[serde(default)]
+    pub node_name: String,
+    /// Width of the node in Figma
```
iamralpht, line 54: I think we should say "Width of the node in the design source" or something like that (and not mention Figma at all in the toolkit_schema side of the house).

designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
```diff
@@ -623,9 +619,10 @@ internal fun DesignFrame(
         }
         is LayoutInfoAbsolute -> {
             if (parentLayout?.isWidgetChild == true) {
+                val childLayout = layout ?: LayoutManager.getGridNodeSize(name)
                 // For direct children of a widget, render the frame as a box with the calculated
                 // layout size, then compose the frame's children with our custom layout
-                Box(m.layoutSizeToModifier(layout)) {
+                Box(m.layoutSizeToModifier(childLayout)) {
```
iamralpht, line 625: Pity we need a Box here - is there something different about how we report the size from `DesignFrameLayout` than how `Box` does it? I'm wondering about cases where the evaluated layout ends up being a different size than the design-tool computed size.

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -105,6 +107,11 @@ internal object LayoutManager {
         density = pixelDensity
     }
 
+    internal fun setGridNodeSize(nodeName: String, width: Int, height: Int) {
```
iamralpht, line 110: Can we have collisions on this? I can imagine the same node name being in two docs (especially in the validation app!); I'm not sure it could happen in a single doc.

designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
rylin8: This was really confusing when I implemented it and it's possible it's not needed anymore. I'll investigate.

designcompose/src/main/java/com/android/designcompose/Layout.kt
rylin8: Yes, there can be collisions. I thought about this but decided not to worry about it yet since I didn't want to put too much complexity into a solution that might not last. If we move forward I can add support for that or make a ticket for it.

designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
rylin8: I've made some changes here such as removing the `Box()` under `DesignFrameLayout()` as it's not needed anymore, but without this particular `Box()`, some of the children are laid out incorrectly and I'm not sure why. For example, the child in the widget at the top is stretched to fit all four columns, and then the green square gets moved to the center instead of hugging the left like it should The bottom node is an example of how it renders outside of a grid layout.
Screenshot 2023-11-03 at 4 53 00 PM

iamralpht: Hm, there's probably something else going on underneath that's not quite right, then... I'm guessing layout is generating the same output either way, so I'm confused what's going wrong (worst case, we have a bug where we don't show what layout generated correctly).

iamralpht:
One other question on this one - do you know why the list items initially measure as 0x0? I'm thinking that this might cause issues in other places, too (anywhere DC is embedded as a widget/view and not filling its parent), so if there's a different approach where the initial compose-initiated layout pass yields good bounds then we should do that instead (or as well?).

rylin8:
> One other question on this one - do you know why the list items initially measure as 0x0? I'm thinking that this might cause issues in other places, too (anywhere DC is embedded as a widget/view and not filling its parent), so if there's a different approach where the initial compose-initiated layout pass yields good bounds then we should do that instead (or as well?).

Yea, they are initially 0x0 because we haven't calculated the layout for them yet. Normally when we calculate layout, a node waits for all of its children to be composed before doing the layout calculation, but lazy grid works differently. Since children of a lazy grid aren't added to our Rust layout manager, the lazy grid and its ancestors all the way up to the root node calculate their layout first. Once the grid layout is done and starts to try to figure out what children it should display, those children are composed, added to Rust layout and computed, but the lazy grid doesn't wait for layout to be done before trying to fit more children in. At this point it's a race condition I believe -- I don't know the details of the order in which lazy grid does all of this.

rylin8:
@iamralpht I updated this to use constraints on children, and that let me remove the `Box()` in `DesignFrame`. However I still needed a bit of a hack to get the list widgets to size properly.

rylin8:
Rebased and fixed conflicts

rylin8:
The screenshot diff looks good. The dials/gauges change is just because I changed the parameter in Figma for the first arc. The other changes are mostly correct in that the old screenshot was wrong. Some of the grid/list changes look exactly the same to me but some outline differences show up -- I think this is acceptable. The one I was concerned about was the OpenLink test where the new screenshot renders the outer frame slightly wider, but after looking into it I think this is correct and the old one is wrong. I tried adding five squares to the bottom #Content in the Figma file and it stretches the frame by 10 pixels, matching the screenshot.
Screenshot 2023-11-08 at 6 49 55 PM

designcompose/src/main/java/com/android/designcompose/Layout.kt
rylin8: https://github.com/google/automotive-design-compose/issues/544

iamralpht (approved):
Cool - I am really excited to see the screenshot testing providing valuable feedback already! These changes are good for `feature/rustlayout`.

iamralpht:
> Yea, they are initially 0x0 because we haven't calculated the layout for them yet. Normally when we calculate layout, a node waits for all of its children to be composed before doing the layout calculation, but lazy grid works differently. Since children of a lazy grid aren't added to our Rust layout manager, the lazy grid and its ancestors all the way up to the root node calculate their layout first. Once the grid layout is done and starts to try to figure out what children it should display, those children are composed, added to Rust layout and computed, but the lazy grid doesn't wait for layout to be done before trying to fit more children in. At this point it's a race condition I believe -- I don't know the details of the order in which lazy grid does all of this.

Ok, that makes sense -- I guess we can't evaluate the layout tree without all of the child Composables getting materialized, which means we won't know the size before the first compose, which I think is what lazy grid needs.

The squoosh branch does report the size on the first compose, because it doesn't use Compose to build the tree of children. So hopefully we'll be able to retire the pre-measurement in the future (assuming we don't find any blockers to not using Compose to manage all of the design nodes).

PR #530, opened 2023-11-06
Give HelloWorld it's own standalone project

### LINKED ISSUE #511: helloworld app missing wrapper gradle task
using commit 92f397b2028129c35e95982abc25290b8441e4f5, I opened the helloworld app as its own project.  The Gradle sync task fails with the error `Task 'wrapper' not found in project ':helloworld-app'.`

image

timothyfroehlich:
Some side changes that were needed to support this

I removed the Android integration test for HelloWorld because it needed the (currently unpublished) test fixtures from DesignCompose. The test had been redundant anyways, ever since I added `designcompose/src/androidTest/kotlin/com/android/designcompose/figmaIntegrationTests/LiveUpdateBehaviorTests.kt`

I also had to move the app's project directory, but this should be a seamless change because I had already been manually setting the project directory for the app. 

Note: The new helloworld app's tests won't compile yet, they use new code from DesignCompose that hasn't been published to the maven repository yet. However the app can still be built and installed.

Fixes #511


rylin8:
Is it supposed to be called "app.DesignCompose.helloworld-app" in the dropdown?

Screenshot 2023-11-07 at 2 46 04 PM

timothyfroehlich:
Hm, yeah, I'm seeing that too. I didn't notice that when I set this up for Tutorial. Looking into it.

timothyfroehlich:
Ok, I think the issue is that the project directories for both helloworld and tutorial are named "app". Android Studio seems to be relying on that a bit too much. But removing the colon from their `include()` lines in `settings.gradle.kts` fixes the issue. I don't think I was supposed to have those colons in the first place, I don't see them in the documentation.

rylin8:
> Ok, I think the issue is that the project directories for both helloworld and tutorial are named "app". Android Studio seems to be relying on that a bit too much. But removing the colon from their `include()` lines in `settings.gradle.kts` fixes the issue. I don't think I was supposed to have those colons in the first place, I don't see them in the documentation.

I tested your latest change but it's the same on my machine. I restarted Android Studio, anything else to try?

rylin8:
> The new entry won't go away on it's own, it'll have a little red x on it which means it doesn't point to anything anymore, basically. The old entry (helloworld-app) will work again.

That's not it -- the old entry (helloworld-app) has an X on it so I removed it, and the new one with the long name is still there and builds and runs HelloWorld.

timothyfroehlich:
I just pulled it down on a fresh checkout and it properly named helloworld-app. I'm trying to remember what I did to get it to detect helloworld after I'd deleted mine, I think I had to go to File -> Invalidate Caches

rylin8 (approved):
I still have the weird run configuration name and can't get it fixed. It's not a big deal so I'll approve. I don't want to checkout a new repo or anything.

PR #531, opened 2023-11-06
Split Validation's examples into individual files

timothyfroehlich:
This will improve the recompilation time of Validation when a change is made to one of the examples. ( 5.9 seconds without the change vs 3.8 seconds) This will be most noticeable when working with Compose Previews.


rylin8:
If you fix conflicts and rebase, I'll test this again, as I'm seeing some weird issues as is.

timothyfroehlich:
Alright, should be good to go. I took the diff of Validation's main activity from #491 and applied it to the individual tests. I had the diffs lined up next to each other in VS Code so I'm confident that I got it right.

rylin8 (approved):

PR #542, opened 2023-11-08
Create a separate Gradle plugin that allows fetching DCF files from unit tests

timothyfroehlich:
This is super rough, but it works. I've added a new internal gradle plugin that adds new tasks that will tell a unit test that has been set up correctly to fetch the Figma files for the tests that it's running. the `fetchFigmaFiles` task will do just that, and `fetchAndUpdateFigmaFiles` will copy the fetched files into the `src/main/assets/figma` folder. Note: `fetchAndUpdate` will only copy the fetched files over once the entire test task is done (so for Validation, once everything has finished fetching). But if you cancel early, the files that have already been fetched will be in build/outputs/designComposeFiles

Right now I only have it set up for Validation.

This is essentially hacky version to make it easier to update Validation tests. I don't intend to make this publicly available because it lacks features, relies on hard coded paths and isn't configurable.


designcompose/src/testFixtures/java/com/android/designcompose/TestUtils.java
```diff
@@ -31,6 +31,9 @@ public class TestUtils {
     public static void triggerLiveUpdate() {
         String actualFigmaToken =
                 InstrumentationRegistry.getArguments().getString("FIGMA_ACCESS_TOKEN");
+        // Check for a system property as a backup
+        if (actualFigmaToken == null)
+            actualFigmaToken = System.getProperty("designcompose.test.figmaToken");
         if (actualFigmaToken == null)
             throw new RuntimeException("This test requires a Figma Access Token");
```
iamralpht, line 38: Include a link to the doc?

iamralpht (approved):
This change looks good, but not sure why the rust layout deserialize test is failing with it applied (maybe it's also failing in `feature/rustlayout`?). For the no-token-set errors, it is nice to point to the doc, I think we do that in some cases already...

designcompose/src/testFixtures/java/com/android/designcompose/TestUtils.java
timothyfroehlich: This is an implementation detail right now, so I dunno. Also it's not available outside of our root project, we're not publishing the test fixtures right now.

Issue #402, opened 2023-09-28
Support text with constraints


rylin8:
Currently constraints on text cause the style’s width/height/margins to be set to something that is ultimately ignored because Android handles text layout. We need to let taffy handle some of the layout in this case – for example if constraints set to LEFT + RIGHT, then taffy should calculate width based on parent container.

Issue #475, opened 2023-10-18
Make the Rust layout manager object an instantiable non-singleton class


rylin8:
This will allow tests to run in parallel, as each test can instantiate its own layout manager. This will require client (Android) and JNI changes. Android will need to instantiate a new layout manager on startup and refer to it for any subsequent layout calls.

rylin8:
I was going to make a map of LayoutManager objects and give each a unique ID. The Android side would call a new JNI function to create one and get an ID back, then it would use that ID in future layout calls. However there's no need for that unless we need multiple layout managers -- can you think of any other reasons we might need this?  Otherwise I can just move the singleton layout manager into jni.rs, make `LayoutManager` functions public, and have the layout tests each instantiate a LayoutManager. Thoughts? @iamralpht @timothyfroehlich

iamralpht:
I think singletons are generally a bad direction, because it's a kind of hidden global state. So no singletons unless it's really needed (like stashing the `JavaVM` pointer in `JNI_OnLoad` into a global).

In this case, I'm not sure when it's best to have one taffy instance, or when to have multiple. So on the Compose side I guess we'd make one and put it in the root-most `CompositionLocal`, then we can move nodes between roots still if needed.

timothyfroehlich:
> However there's no need for that unless we need multiple layout managers -- can you think of any other reasons we might need this? 

Multiple screens?

> I think singletons are generally a bad direction, because it's a kind of hidden global state. So no singletons unless it's really needed (like stashing the JavaVM pointer in JNI_OnLoad into a global).

Agreed here. I think that all singletons need to have a compelling reason to justify being a singleton. 

We need to make sure we're developing with testability in mind and singletons automatically make testing more difficult, especially when they're statically allocated.

rylin8:
I'm not trying to promote singletons. I've removed the one from layout.rs and the unit tests all create their own layout manager instance now and can run with multiple threads. Maybe I'm not understanding JNI correctly, but I'm just trying to find the correct place to store the `LayoutManager` instances in memory when running Android. I don't want to serialized a `LayoutManager` and have Android hold onto it. Maybe `JNIEnv` has some way to store objects within the lifetime of itself?

Issue #468, opened 2023-10-16
Cargo deny fails for Rust Layout


timothyfroehlich:
https://github.com/google/automotive-design-compose/actions/runs/6527204689/job/17721745890

```
warning[duplicate]: found 2 duplicate entries for crate 'jni'
   ┌─ /github/workspace/Cargo.lock:46:1
   │  
46 │ ╭ jni 0.1.0 /github/workspace/crates/jni
47 │ │ jni 0.21.1 registry+https://github.com/rust-lang/crates.io-index
   │ ╰────────────────────────────────────────────────────────────────^ lock entries
   │  
   = jni v0.1.0
   = jni v0.21.1
     └── jni v0.1.0

error[unlicensed]: layout = 0.1.0 is unlicensed
  ┌─ layout 0.1.0 (path+file:///github/workspace/crates/layout):2:9
  │
2 │ name = "layout"
  │         ^^^^^^ a valid license expression could not be retrieved for the crate
3 │ version = "0.1.0"
4 │ license = ""
  │            - license expression was not specified
  │
  = layout v0.1.0
    └── jni v0.1.0

error[rejected]: failed to satisfy license requirements
  ┌─ slotmap 1.0.6 (registry+https://github.com/rust-lang/crates.io-index):4:[12](https://github.com/google/automotive-design-compose/actions/runs/6527204689/job/17721745890#step:6:13)
  │
4 │ license = "Zlib"
  │            ^^^^
  │            │
  │            license expression retrieved via Cargo.toml `license`
  │            rejected: not explicitly allowed
  │
  = slotmap v1.0.6
    └── taffy v0.3.[15](https://github.com/google/automotive-design-compose/actions/runs/6527204689/job/17721745890#step:6:16)
        ├── figma_import v0.1.0
        │   ├── jni v0.1.0
        │   └── layout v0.1.0
        │       └── jni v0.1.0 (*)
        └── layout v0.1.0 (*)

bans ok, licenses FAILED, sources ok
```

For the first one we might need to rename our JNI crate. (Ticket forthcoming on that).

For the second one we just need to specify a license for the new layout crate. (See the Cargo.tomls for the old crates)

For the last one you'll need to add it to the `[licenses.deny]` block of `deny.toml`, but make sure that it's an allowed license first (go/thirdpartylicenses)

PR #548, opened 2023-11-09
Closes #546: Improve text measurement performance

### LINKED ISSUE #546: Improve text measurement performance
Use ParagraphIntrisics to improve text measurement performance. Simplify the text measurement path by always subscribing to layout with the measure function.

rylin8:
Use Paragraph intrinsics to improve performance of measuring text size. Remove one of the two text measurement paths by always subscribing to layout with the measure function. This lets us remove a large amount of code that is no longer called.


crates/layout/src/layout.rs
```diff
@@ -380,13 +380,9 @@ impl LayoutManager {
     ) {
         if let Some(fixed_width) = fixed_width {
             style.min_size.width = taffy::prelude::Dimension::Points(fixed_width as f32);
-            style.max_size.width = taffy::prelude::Dimension::Points(fixed_width as f32);
-            style.size.width = taffy::prelude::Dimension::Points(fixed_width as f32);
```
iamralpht, line 384: Did this end up fixing the wrapping issue?

rylin8:
One change that this makes is that text nodes set to auto-width used to stretch indefinitely on one line. Now they might wrap. From what I can tell, they wrap if they are in a parent frame and their width exceeds the parent frame's width because that width is passed into the measure function's `availableWidth`. This might be ok -- I don't think it's common that you would want the text to grow past the parent frame, but it also is different behavior than Figma.

To support the way it worked before (meaning before this change, but not in the main branch), we would need to have the measure func determine if the text node's text sizing is set to auto-width. To do that we need to look at the Figma node's `text_style.text_auto_resize` property which is not exposed, or figure out a combination of values from what is currently exposed in `ViewStyle`. Either way, I think it is fine to let auto-width text behave as it is right now.

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -325,106 +291,6 @@ internal object LayoutManager {
     }
 }
 
-private class LayoutDeltaCanvas :
```
iamralpht, line 328: Pretty satisfying to see all this junk go away! Thank goodness the Compose folks added the `PlatformTextStyle` stuff :).

designcompose/src/main/java/com/android/designcompose/DesignText.kt
```diff
@@ -318,7 +310,7 @@ internal fun DesignText(
                 )
             } else {
                 // Text needs to use a modifier that sets the size so that it wraps properly
-                val height = renderHeight ?: layout?.height() ?: 0
+                val height = layout?.height() ?: 0
                 val textModifier = modifier.sizeToModifier(layout?.width() ?: 0, height)
                 BasicText(
```
iamralpht, line 315: Can this take `ParagraphIntrinsics` instead of `annotatedText`? I think it's going to make a `ParagraphIntrinsics` internally somewhere, and that seems slow.

iamralpht:
Looks like this solved the Battleship instruction numbers being in the wrong place, so that's good. squoosh-3 also has the broken "WordArt!" text in the fancy fills test. I haven't looked into that, but I think it's exactly the issue you're calling out in your comment -- can you file a ticket to fix it at some point?

rylin8:
The screenshot diffs show mostly text changes, but I don't know why it shows some of the others like the dials/gauges change.

designcompose/src/main/java/com/android/designcompose/DesignText.kt
rylin8: `BasicText()` doesn't take `ParagraphIntrinsics` -- am I missing something?

crates/layout/src/layout.rs
rylin8: No, I remember adding this toward the beginning when I was. having some layout problems. This fixed my issue at the time but it didn't seem right, so I think it's good to remove, as I didn't see any regressions when I removed it.

rylin8:
Updated to support auto width text nodes expanding past their parent.

rylin8:
I believe this fixes #402 also, since I made that to fix the Battleship numbers.

designcompose/src/main/java/com/android/designcompose/DesignText.kt
iamralpht: No, you're not, I was assuming that there would be a `BasicText` constructor that took `ParagraphIntrinsics` instead of an `AnnotatedString` and a `TextStyle`... but there doesn't appear to be. Looking at the implementation, `BasicText` eventually uses this [`MultiParagraphLayoutCache`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/text/modifiers/MultiParagraphLayoutCache.kt) which keeps a `MultiParagraphIntrinsics` instance around for the text and style that's passed in.

So Compose uses `MultiParagraph` and `MultiParagraphIntrinsics` instead of `Paragraph` and `ParagraphIntrinsics`, because a `BasicText` supports multiple paragraph styles in a run. We don't encounter that, so `Paragraph` and `ParagraphIntrinsics` is OK for us, but unfortunately I don't see any way for us to use Compose's `MultiParagraphLayoutCache` -- so while we use `BasicText`, we're always going to be laying out text once for ourselves, and once for Compose. (So what you have now is as good as it gets).

iamralpht:
> I believe this fixes #402 also, since I made that to fix the Battleship numbers.

Yes, because the rust layout is always just laying out the box to contain the text using the measure func to get height-for-width, we now support constraints on text (or any kind of layout stuff that can be applied to frames or other types).

iamralpht:
I think we might need to update the licenses file to say that the zlib license is OK, and add the apache-2.0 license declaration to the `Cargo.toml`s, and rename the `jni` crate to something else (`dcjni` would be fine; it's a bit weird having a `libjni.so` in our release -- of course a native shared object packaged in an android sdk is going to have a jni interface...).

rylin8:
> I think we might need to update the licenses file to say that the zlib license is OK, and add the apache-2.0 license declaration to the `Cargo.toml`s, and rename the `jni` crate to something else (`dcjni` would be fine; it's a bit weird having a `libjni.so` in our release -- of course a native shared object packaged in an android sdk is going to have a jni interface...).

What changed here that needs these license changes? Or do you mean just for this whole layout branch?

iamralpht:
> What changed here that needs these license changes? Or do you mean just for this whole layout branch?

I am guessing this license compliance job has always been failing for the rust layout branch; I just happened to notice it when looking at the CI status. 

If it's not going to prevent merging this change then let's address the license config in a dedicated PR.

rylin8:
> > What changed here that needs these license changes? Or do you mean just for this whole layout branch?
> 
> I am guessing this license compliance job has always been failing for the rust layout branch; I just happened to notice it when looking at the CI status.
> 
> If it's not going to prevent merging this change then let's address the license config in a dedicated PR.

I think that's this ticket https://github.com/google/automotive-design-compose/issues/468

iamralpht (approved):
I think you also need to fix the rust layout unit tests, but this is OK for `feature/rustlayout`.

rylin8:
> I think you also need to fix the rust layout unit tests, but this is OK for `feature/rustlayout`.

The unit tests will be fixed with the change to address https://github.com/google/automotive-design-compose/issues/475

PR #562, opened 2023-11-15
Closes #475: Make the Rust layout manager object an instantiable non-…

### LINKED ISSUE #475: Make the Rust layout manager object an instantiable non-singleton class

rylin8:
…singleton class

Make the LayoutManager a normal instantiable object, removing the singleton and mutex code out of layout.rs. Public helper functions have been removed from layout.rs and member functions of LayoutManager have been made public. Unit tests now each instantiate a LayoutManager and use their instance to perform layout tests so that they can be run in parallel.

jni.rs now handles instantiation of LayoutManager for use with Android. A new jni create_layout_manager() function creates a layout manager and returns a unique ID for the manager. Internally we keep a hashmap of LayoutManager objects indexed by the unique ID.


timothyfroehlich (changes requested):
This should allow cargo test to succeed, correct? The tests are still failing due to other issues.

crates/layout/src/layout.rs
timothyfroehlich, line 1: The diff for layout was too large for GitHub to display without expanding it. That's a good indicator that the file is too long. I'd like the file to be split so that the diff isn't hidden any more.

crates/layout/src/layout.rs
```diff
@@ -414,147 +395,34 @@ impl LayoutManager {
...
-    let layout_node = manager.layout_id_to_taffy_node.get(&layout_id);
-    if let Some(layout_node) = layout_node {
-        let layout = taffy.layout(*layout_node);
-        if let Ok(layout) = layout {
-            let name_result = manager.layout_id_to_name.get(&layout_id);
-            if let Some(name) = name_result {
-                println!("{}Node {} {}:", space, name, layout_id);
-                println!("  {}{:?}", space, layout);
-            }
+    fn print_layout_recurse(&self, layout_id: i32, space: String) {
```
timothyfroehlich, line 408: I'd say either remove the println()s or replace them with log.verbose calls.

timothyfroehlich (changes requested):

rylin8:
> This should allow cargo test to succeed, correct? The tests are still failing due to other issues.

Oops I will fix this up.

crates/layout/src/layout.rs
rylin8: My understanding is the diff is hidden when there are a lot of changes to a file. I changed every function in this file which is why it's hidden, but also reduced its size. It is at ~430 lines which I don't think is too long. This file has the implementation for the LayoutManager struct, which is not that large and which I don't think makes sense to split up.

timothyfroehlich: Ok, yeah, 430 isn't too bad.

crates/layout/src/layout.rs
rylin8: Apparently our logging crate doesn't like to print to stdout. I only use this for fetch_layout, a binary I use for testing. I made this take a print function instead, and fetch_layout uses println because I want it in stdout.

timothyfroehlich:
Looks good to me now, though Ralph should take a look as well.

rylin8:
Rebased

crates/jni/src/jni.rs
```diff
@@ -21,12 +21,43 @@ use figma_import::{fetch_doc, ConvertRequest, ProxyConfig};
...
+fn javavm() -> MutexGuard<'static, Option<JavaVM>> {
+    JAVA_VM.lock().unwrap()
+}
+
+fn create_layout_manager() -> i32 {
+    let manager = Arc::new(Mutex::new(LayoutManager::new()));
+    let mut hash = LAYOUT_MANAGERS.lock().unwrap();
+
+    let manager_id = LAYOUT_MANAGER_ID.fetch_add(1, Ordering::SeqCst);
+    println!("### manager_id {}", manager_id);
```
iamralpht, line 47: probably delete this (or make it log!)

crates/jni/src/jni.rs
```diff
@@ -21,12 +21,43 @@ use figma_import::{fetch_doc, ConvertRequest, ProxyConfig};
...
+}
+static LAYOUT_MANAGER_ID: AtomicI32 = AtomicI32::new(0);
+
+fn javavm() -> MutexGuard<'static, Option<JavaVM>> {
+    JAVA_VM.lock().unwrap()
+}
+
+fn create_layout_manager() -> i32 {
+    let manager = Arc::new(Mutex::new(LayoutManager::new()));
+    let mut hash = LAYOUT_MANAGERS.lock().unwrap();
```
iamralpht, line 44: I think we'll need to define a policy on `unwrap()` soon (also for HAR); in this file an un-unwrappable value probably should generate a java exception and return.

iamralpht:
Minor comments on `unwrap()` otherwise LGTM.

crates/jni/src/jni.rs
rylin8: Made https://github.com/google/automotive-design-compose/issues/670

crates/jni/src/jni.rs
rylin8: Deleted.

timothyfroehlich (approved):

iamralpht (approved):

PR #567, opened 2023-11-20
Add Traces to some high-impact DesignCompose Functions, Create Battleship Playfield benchmark

timothyfroehlich:
See `docs/_docs/working-with-source/profiling-and-benchmarks.md` for general documentation.

I've added one benchmark for startup launches and one for a large refresh of the Battleship app. Neither are amazing, but they're a good starting point and example.


benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/ClickPlayfield.kt
```diff
@@ -0,0 +1,80 @@
...
+                    TraceSectionMetric(DCTraces.DESIGNDOCINTERNAL),
+                    TraceSectionMetric(DCTraces.DESIGNVIEW),
+                    TraceSectionMetric(DCTraces.DESIGNFRAME_DE_SUBSCRIBE, Mode.Sum),
+                    TraceSectionMetric(DCTraces.DESIGNFRAME_FINISHLAYOUT, Mode.Sum),
+                    TraceSectionMetric(DCTraces.DESIGNTEXT_DE, Mode.Sum),
+                    TraceSectionMetric(DCTraces.JNIADDNODES, Mode.Sum),
+                    TraceSectionMetric(DCTraces.DOCSERVER_DOC, Mode.Sum),
+                    TraceSectionMetric(DCTraces.DESIGNVIEW_INTERACTIONSCOPE, Mode.Sum),
+                ),
+            iterations = 5,
```
bentrengrove, line 58: 5 is pretty low, good for development but I wouldn't draw any conclusions from it. To give some idea, I normally develop on ~3, make initial judgements on 10, and then before telling the team "I improved performance by x" I run it on 30.

bentrengrove: It can help to plot the results to see how noisy each iteration is, we have a WIP tool for it at https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:development/plot-benchmarks/;l=1?q=plot&sq=&ss=androidx

benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/ClickPlayfield.kt
```diff
@@ -0,0 +1,80 @@
...
+            /**
```
bentrengrove, line 59: This is because (as confusing as this is) COLD benchmarks kill the process between the setupBlock and the measureBlock. So if you have a cold benchmark you need to start the activity in the measure block.

Apparently this was done to match the behaviour of hawkeye, but it causes constant confusion

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -501,6 +503,7 @@ internal fun DesignView(
                     //
                     // The interaction test document covers all of these cases.
                     interactionScope.launch {
+                        beginSection("DesignView InteractionScope")
```
bentrengrove, line 506: Just curious, does this work? I would have thought you would need to have the trace in onPress to get a result

designcompose/src/main/java/com/android/designcompose/Jni.kt
```diff
@@ -35,32 +36,92 @@ internal class TextSize(
     var height: Float = 0F,
 )
 
+internal interface JniIntefrace {
```
bentrengrove, line 39: typo

docs/_docs/working-with-source/profiling-and-benchmarks.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Build variants
+
+Benchmarks use a separate build variant (`benchmark`) which extends from the `release` variant. The main DesignCompose library also has a `benchmark` variant, specifically to let us use the Cargo plugin's ABI filtering to only build the ABI of the Rust libraries that we need, rather than all 4 every time.
+
+## Tips
+
+When writing a new benchmark, use [Dry Run mode](https://developer.android.com/topic/performance/benchmarking/macrobenchmark-instrumentation-args#dryrunmode-enable) to test benchmark behavior using only a single iteration, rather than altering the number of runs in the test.
+
+Whenever investigating traces, use the Perfetto WebUI, it's much better than Android Studio's.
```
bentrengrove, line 32:
```suggestion
Whenever investigating traces, use the [Perfetto WebUI](https://ui.perfetto.dev/), it's much better than Android Studio's.
```

benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/ClickPlayfield.kt
timothyfroehlich: Awesome, I needed something like that.

benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/ClickPlayfield.kt
timothyfroehlich: Ok, I think I get it. Is this a valid way to reset the app's state between runs? I don't want to capture startup behavior in the benchmark, I specifically want to test a specific interaction. I could always write code to click the right buttons on the app to get it back to home, but then I'd be worried about the result being affected by some state that was left behind, and not a relatively clean state. 

Like, it makes sense to try to measure the first interaction with an app, right?

benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/ClickPlayfield.kt
timothyfroehlich: > 5 is pretty low, good for development but I wouldn't draw any conclusions from it. To give some idea, I normally develop on ~3, make initial judgements on 10, and then before telling the team "I improved performance by x" I run it on 30.

Noted.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
timothyfroehlich: Yep :)

image

Blue bar at the top

benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/ClickPlayfield.kt
bentrengrove: Yeah what you are doing is valid. I think you can actually pass null to startup mode and then completely manage it yourself

timothyfroehlich: Yep, that works, and makes it so that it doesn't start up the app for the warm start only for me to kill it on the first go.

designcompose/src/main/java/com/android/designcompose/Jni.kt
```diff
@@ -35,32 +36,92 @@ internal class TextSize(
     var height: Float = 0F,
 )
 
+internal interface UniInterface {
```
iamralpht, line 39: What is `UniInterface`? Why do you need an interface if `TracedJni` is the only type exported from this file?

timothyfroehlich: Wow, I replaced the typo with another typo.

I was hesitant to unilaterally add the tracing code around the JNI interface and thought it might be best to leave the traced and untraced versions. I'm not really sure if I need the interface actually, I was kind of figuring it out. I suppose I could just have TracedJNI be a subclass. Or I could just move the tracing versions of the functions into the single JNI class.

iamralpht:
@rylin8 ptal - this looks good to me, but I wonder if we should do tracing explicitly around where we're performing layout. Afaik layout subscription is unlikely to ever be relevant, so I'm not sure about the value of pulling that out especially (if we have a mystery/unexpected perf then simpleperf is going to be more useful than perfetto, then we can put trace points around the things that simpleperf found).

@timothyfroehlich Longer term, this will let us gather detailed performance data commit-to-commit, so we don't regress layout by updating taffy, or something like that, right? (For the nitty gritty performance grind, I'm currently spending a lot of time with simpleperf and dreaming about perf-to-kcachegrind conversion scripts ;)).

designcompose/src/main/java/com/android/designcompose/Jni.kt
iamralpht: I'd just export one class with the methods you want, and have tracing on the ones that we know are expensive. You can subclass and override, but my expectation would be that if a method name starts `jni` then it's native code, and I'd be confused by a stacktrace with Java code in a `jni`-prefixed function that I know doesn't call `Call*Method`. (I would think "ugh! the stack has been corrupted!").

timothyfroehlich: How's this?

timothyfroehlich:
> 
> @timothyfroehlich Longer term, this will let us gather detailed performance data commit-to-commit, so we don't regress layout by updating taffy, or something like that, right? (For the nitty gritty performance grind, I'm currently spending a lot of time with simpleperf and dreaming about perf-to-kcachegrind conversion scripts ;)).

This gives us something to build off of. Gathering data commit-to-commit will require Firebase Test Lab infrastructure and dashboards to hold the result and visualize it. But in the mean time benchmarks can be run and compared against manually.

benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/StartupBenchmarks.kt
```diff
@@ -16,65 +16,38 @@
...
+                listOf(
+                    StartupTimingMetric(),
+                    TraceSectionMetric(DCTraces.DECODEDISKDOC),
+                    TraceSectionMetric(DCTraces.DESIGNDOCINTERNAL),
+                    TraceSectionMetric(
+                        DCTraces.DESIGNFRAME_DE_SUBSCRIBE,
+                        TraceSectionMetric.Mode.Sum
+                    ),
+                    TraceSectionMetric(DCTraces.FETCHDOCUMENTS)
+                ),
```
rylin8, line 49: What does this list do? Does it sum up all of these timings?

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -936,6 +940,7 @@ internal fun DesignDocInternal(
     parentComponents: List<ParentComponentInfo> = listOf(),
     parentLayout: ParentLayoutInfo? = null,
 ) {
+    beginSection(DCTraces.DESIGNDOCINTERNAL)
```
rylin8, line 943: Where's the `endSection()` for this, and other `beginSection()` calls that don't have an `endSection()`?

designcompose/src/main/java/com/android/designcompose/DCTraces.kt
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.android.designcompose
+
+object DCTraces {
+    const val DESIGNFRAME_DE_SUBSCRIBE = "DesignFrame DisposableEffect subscribe"
+    const val DESIGNFRAME_FINISHLAYOUT = "DesignFrame.finishLayout"
+    const val DESIGNTEXT_DE = "DesignText DisposableEffect"
```
rylin8, line 22: Seems a bit odd that this doesn't have "SUBSCRIBE" in the name but the DesignFrame one does.

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -132,12 +132,15 @@ internal object LayoutManager {
         name: String,
         textMeasureData: TextMeasureData,
     ) {
-        textMeasures[layoutId] = textMeasureData
-        subscribe(layoutId, setLayoutState, parentLayoutId, childIndex, style, name, true)
-
-        // Text cannot have children, so call computeLayoutIfComplete() here so that if this is the
-        // text or text style changed when no other nodes changed, we recompute layout
-        computeLayoutIfComplete(layoutId, rootLayoutId)
+        trace(DCTraces.LAYOUTMANAGER_SUBSCRIBEWITHMEASURE) {
```
rylin8, line 135: Doesn't DESIGNTEXT_DE cover this?

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -132,12 +132,15 @@ internal object LayoutManager {
...
+            textMeasures[layoutId] = textMeasureData
+            subscribe(layoutId, setLayoutState, parentLayoutId, childIndex, style, name, true)
+
+            // Text cannot have children, so call computeLayoutIfComplete() here so that if this is
+            // the
+            // text or text style changed when no other nodes changed, we recompute layout
+            computeLayoutIfComplete(layoutId, rootLayoutId)
```
rylin8, line 142: This function asks Rust to compute layout, so it could be worth tracing.

rylin8 (commented):
I'll need to spend some more time thinking about strategic places to add trace sections, but this is a great start!

benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/StartupBenchmarks.kt
timothyfroehlich: That's described here: https://developer.android.com/topic/performance/benchmarking/macrobenchmark-metrics

designcompose/src/main/java/com/android/designcompose/DesignView.kt
timothyfroehlich: Good catch, looks like I lost the endSection for this.

designcompose/src/main/java/com/android/designcompose/DCTraces.kt
timothyfroehlich: True

designcompose/src/main/java/com/android/designcompose/Layout.kt
timothyfroehlich: It's wrapped in the LAYOUTMANAGER_SUBSCRIBEWITHMEASURE trace

timothyfroehlich:
About the traces I placed, I'm just spitballing. Like I said in the previous comment, there's a lot more work to be done to start capturing and comparing results longer term. But this is a good base and working example if you two want to work on performance.

benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/ClickPlayfield.kt
```diff
@@ -0,0 +1,77 @@
...
+    @get:Rule val benchmarkRule = MacrobenchmarkRule()
+
+    @Test
+    fun changePlayers() =
+        benchmarkRule.measureRepeated(
+            packageName = PACKAGE_NAME,
+            metrics =
+                listOf(
+                    TraceSectionMetric(DCTraces.DESIGNDOCINTERNAL),
+                    TraceSectionMetric(DCTraces.DESIGNVIEW),
```
rylin8, line 49: Is there no `Mode.Sum` for these two trace types because you're assuming that the first call to them is the root node, so we can ignore all the children? Also, I think I asked this already but you can tell me again how it works when trace calls are nested? `DesignDocInternal` calls `DesignView`, which calls `DesignFrame` which calls `subscribe`, etc.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -1016,6 +1023,7 @@ internal fun DesignDocInternal(
                     }
                 }
 
+                beginSection(DCTraces.DESIGNVIEW)
```
rylin8, line 1026: I don't see an `endSection()` for this either.

designcompose/src/main/java/com/android/designcompose/Layout.kt
rylin8: Yes, but it's called in other places also, in particular from `DesignFrame`.

timothyfroehlich:
@rylin8 Ok, moved the trace from subscribe with measure to computeLayout.

designcompose/src/main/java/com/android/designcompose/Layout.kt
timothyfroehlich: Ok, moved the trace from subscribe with measure to computeLayout.

benchmarks/battleship/benchmark/src/main/java/com/android/designcompose/benchmark/battleship/benchmark/ClickPlayfield.kt
timothyfroehlich: Yeah, I was thinking that they should only be called once, but I dunno. It'd be best to experiment to see what works and what doesn't.
Nested trace calls don't influence each other, they record and report individually.

timothyfroehlich:
@bentrengrove pointed out how I can add a `TraceSectionMetric()` for a composable directly, so I've added that and addressed all other review comments. This should be good to go now. I don't want to keep fiddling with trace names and placement because that's all very particular to what you're measuring. I won't be recording numbers or trying to draw any conclusions from what I have here. This is just a framework to build off of and to support the current performance work on rustlayout.

rylin8 (approved):

PR #571, opened 2023-11-21
Fix #570: add support for Brush customizations.

### LINKED ISSUE #570: Support `Brush` customizations
If we add support for `Brush` and `() -> Brush` customizations, then on Android > T we can replace parts of the UI with fragment shaders.

Initially, adding support for `Brush` allows us to test out the infrastructure. Later on, we can figure out how to let the designer specify the fragment shader and its uniform values.

iamralpht:
This change adds basic support for customizing the `Brush` that a frame renders with.


iamralpht:
Let's see what CI does with this! I haven't added Brush customizations for text yet, but I'll add that later this evening.

iamralpht:
@rylin8 I think this change is ready for your review. I am wondering:

- should the `BrushFunction` take an `ImageCustomizationContext`? Then someone could put the color in as a shader uniform and make some common shaders that they might use for different backgrounds (from different docs).
- text doesn't seem to update live; I'm not sure what's up with that... it's good enough for a first pass but we'll need a ticket to come back.

docs/_docs/modifiers/modifiers.md
```diff
@@ -122,6 +122,57 @@ called `artwork`, which is a nullable `android.graphics.Bitmap` instance. To use
 the background fill from the document, pass a value to replace the background
 fill or pass `null`.
 
+### Brush content {#brush-content}
```
rylin8, line 125: Wow documentation too!

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -318,18 +318,33 @@ internal fun ContentDrawScope.render(
...
             val p = Paint()
-            val b = background.asBrush(document, density)
-            if (b != null) {
-                val (brush, fillOpacity) = b
-                brush.applyTo(frameSize, p, fillOpacity)
-                p
-            } else {
-                null
+            customFillBrush.applyTo(frameSize, p, 1.0f)
+            listOf(p)
```
rylin8, line 333: That's pretty simple.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/MainActivity.kt
```diff
@@ -1761,6 +1777,78 @@ fun VariableBorderTest() {
     VariableBorderTestDoc.Main()
 }
 
+@DesignDoc(id = "oetCBVw8gCAxmCNllXx7zO")
```
rylin8, line 1780: @timothyfroehlich I thought these were in separate files now, or is that only in the layout branch? If so what was the reason for that?

rylin8:
> @rylin8 I think this change is ready for your review. I am wondering:
> 
> * should the `BrushFunction` take an `ImageCustomizationContext`? Then someone could put the color in as a shader uniform and make some common shaders that they might use for different backgrounds (from different docs).
> * text doesn't seem to update live; I'm not sure what's up with that... it's good enough for a first pass but we'll need a ticket to come back.

`ImageCustomizationContext` is a new thing? Do you mean like specifying a color in Figma to use in the shader? We have one called `ImageContext` already.

designcompose/src/main/java/com/android/designcompose/CustomizationContext.kt
```diff
@@ -102,6 +107,7 @@ private fun Customization.clone(): Customization {
     c.text = text
     c.image = image
     c.imageWithContext = imageWithContext
+    c.brush = brush
```
rylin8, line 110: I think textFunction (which I probably missed) and brushFunction need to be cloned here also.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/MainActivity.kt
iamralpht: I can do a cherry-pick and rework it if needed.

iamralpht:
> `ImageCustomizationContext` is a new thing? Do you mean like specifying a color in Figma to use in the shader? We have one called `ImageContext` already.

No, I meant `ImageContext`, doh. I think I should probably use `ImageContext` instead of nothing...

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
iamralpht: Yeah, it's nice how they added `RuntimeShader` and `ShaderBrush` -- made this super simple coz we were already using `Brush` everywhere anyway.

docs/_docs/modifiers/modifiers.md
iamralpht: I'm living in @timothyfroehlich shadow ;) -- I just followed the amazing docs on testing to generate a baseline image for the shader test. Literally the only manual step I had to do was pull the new `dcf` file off the device, and that's already automated in the `feature/rustlayout` branch :).

designcompose/src/main/java/com/android/designcompose/CustomizationContext.kt
iamralpht: Yep, good catch! Fixed!

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/MainActivity.kt
timothyfroehlich: Yes, it's only in the layout branch. The reason is that there is and will be a big diff for lots of things in Validation, so I didn't want to try to continuously handle the conflict between the two. That's part of why we agreed to do the majority of development in the rustlayout branch when we talked a few weeks ago.

rylin8 (approved):

iamralpht:
ok, filed #572 and #573 for follow-up work.

PR #566, opened 2023-11-17
Merge Main into RustLayout

timothyfroehlich:

PR #575, opened 2023-11-21
User/ralph/merge main to f rustlayout

iamralpht:


rylin8:
Why is the screenshot diff like that?

rylin8:
Hmm it's weird that there are 11 commits in this PR, most of which @timothyfroehlich merged in yesterday in https://github.com/google/automotive-design-compose/pull/566

iamralpht:
> Hmm it's weird that there are 11 commits in this PR, most of which @timothyfroehlich merged in yesterday in https://github.com/google/automotive-design-compose/pull/566

Yes, I'm not sure why it did that -- I did a merge commit with both branches HEADs. I did try to do a rebase but because of previous merges it looked like a lot of work.

It should be the case that this change means a "git pull origin main" on the feature branch is clean, and a "git merge feature/rustlayout" on the main branch is also clean...

iamralpht:
> Why is the screenshot diff like that?

I had the wrong version of the DCF file in the first push; since fixed and the screenshot diff is clean now.

rylin8 (approved):

PR #576, opened 2023-11-22
Abstract test code out, Create a published test library, re-home the benchmarks

timothyfroehlich:
- Create the `com.android.designcompose.test` package to hold testing utilities and functions. This is a new published artifact and will allow users of DesignCompose to test their apps. 
  - This is just the start, and there isn't enough for proper tests yet. The test-fixture from the main DesignCompose library needs to be moved to it, and the host-built Rust libraries must also be embedded in the library. But this does allow me to preserve the unit test in HelloWorld.
- Abstract out some of the functions and repeated declarations from the test and AndroidTest code into a common test-internal package
- Create an integration-tests/app-common library to hold common app code (currently just the Interfont)
- Move the Benchmarks apps to be within the integration-tests directory. This collects all of the code that is used for integration-tests into a single directory, better separating them.
- Fix the integrated tests
- Add a new Github workflow for making sure that all code compiles in the main project
- Reduce the existing workflow that created the maven repository for the reference apps to use to just do that, this should help speed up CI overall.
- Add some Android Studio run configs to make it easier to run certain checks, such as making sure the code compiles, compiling everything and testing everything other than integration tests, and doing a test dry run of one of the benchmarks.


iamralpht (approved):
I think you'll need to rebase, but this is a good improvement. When do you think we'll be ready to put docs together for folks using DC?

timothyfroehlich:
> I think you'll need to rebase, but this is a good improvement. When do you think we'll be ready to put docs together for folks using DC?

No idea, not until we can spend some time on it. We should do some planning first, and I'll need availability from Rob to help me with certain decisions and with reviews.

PR #578, opened 2023-11-22
WIP: squoosh-3

iamralpht:
Alternate implementation of the DesignCompose renderer that uses Compose more strategically, making layout easier to integrate, and new features like animation a little easier to achieve.


iamralpht:
Closing PR because `user/ralph/squoosh-4` is further along; it should get merged into `feature/rustlayout` in a disabled state and then worked on from there.

Issue #96, opened 2023-05-31
Unknown Figma Connection failure in embedded environment


timothyfroehlich:
Specifically Android Automotive OS, where multiple users can be logged in.

Bug: Even when you switch back to a user that had the token set already, the token was "lost", and you need to set the token again

iamralpht:
Aspirational for 0.21; not clear how much of a blocker this is for partners (Favorites might be higher priority)

timothyfroehlich:
High priority

timothyfroehlich:
@urpatil I've just tried to reproduce it on an emulator ( running system-images;android-32;android-automotive-playstore;x86_64) but couldn't get it to "lose" the token. I'll need more info here, starting with the system image (or at least API level) and reproduction steps.

I did find one issue that prevented setting the token using the Gradle task on multi-user systems, and I'm fixing that in #124 . I'm also including the rough script I wrote to test live update.

iamralpht:
Repros on Trout UDC, per @urpatil

timothyfroehlich:
Awaiting setup instructions for a Trout UDC build from @urpatil

iamralpht:
moving to next release since @timothyfroehlich doesn't have a repro setup (and 0.21 cut-off for changes is tomorrow).

iamralpht:
This is hard to reproduce, and might be because of poor error messages during no connectivity.

timothyfroehlich:
Further investigation has shown that this is a networking issue caused by the host that results in all network connectivity failing. Closing this.

PR #603, opened 2023-12-11
Add support for larger Figma Files

### LINKED ISSUE #582: Use `serde_json::from_reader` instead of `serde_json::from_str` in document.rs
Some folks are bumping into `ureq`'s [10MB limit](https://docs.rs/ureq/latest/src/ureq/response.rs.html#466) for `into_string()`, and it's not a configurable limit. Additionally, using `into_string()` on every response means we have at least one copy of the whole doc in heap, and likely more since we use Strings and not slices in the `figma_schema` form.

We should switch to `serde_json::from_reader` and impose our own limit.

timothyfroehlich:
Taking John's code and adding some testing around it.

Fixes #582 and #96


crates/figma_import/src/document.rs
```diff
@@ -50,8 +50,13 @@ fn http_fetch(api_key: &str, url: String, proxy_config: &ProxyConfig) -> Result<
...
         .set(FIGMA_TOKEN_HEADER, api_key)
         .timeout(Duration::from_secs(90))
-        .call()?
-        .into_string()?;
+        .call()?;
+
+    let mut body_reader = body.into_reader();
+    let mut buffer = Vec::new();
+    let _sz = body_reader.read_to_end(&mut buffer).unwrap();
+    let body = String::from_utf8(buffer).unwrap();
```
rylin8, line 58: Best to avoid `unwrap()` calls when possible. Here you should be able to replace them with the `?;` operator to propagate errors back to the caller. You can also remove the unused `_sz` variable.

rylin8 (changes requested):

rylin8 (approved):

Issue #553, opened 2023-11-14
Click interactions require two clicks


rylin8:
Click interactions seem to require two clicks to trigger after this change went in:
https://github.com/google/automotive-design-compose/pull/541

You can test this out with the "Interaction" Validation test.

iamralpht:
I think this is the change that caused our regression: https://medium.com/androiddevelopers/improvements-and-changes-to-composes-pointer-input-6026904ac972

PR #631, opened 2023-12-14
Fix "Click interactions require two clicks #553"

timothyfroehlich:
I'm not confident that this is the best fix, but it does seem to work. By making the coroutine start undispatched it runs immediately and handles the click properly again. 

I've added a unit test to exercise some of the interaction tests. More can be added later.


rylin8:
Added @iamralpht since he wrote this and knows it better.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -502,7 +503,7 @@ internal fun DesignView(
                     // event so that the interaction state can revert the change).
                     //
                     // The interaction test document covers all of these cases.
-                    interactionScope.launch {
+                    interactionScope.launch(start = CoroutineStart.UNDISPATCHED) {
```
iamralpht, line 506: I'm not sure why we need `interactionScope` to start with. I did write it, but I don't remember why. Reading the documentation on `CoroutineStart` it looks like the default behavior is sometimes lazy, which I guess explains why the coroutine wasn't getting started immediately... but I still don't know what the point of having an explicit coroutine scope is.

timothyfroehlich: My best guess is that this was trying to get the interaction to survive the destruction of the Composable and replacement with one of its variants. Perhaps we should get on GVC to explore it more?

iamralpht: Ah, yes, that does make sense; the drag and press-and-hold interactions need to be kept alive after they've started, even though the first part of the interaction might remove the Composable that's holding the interaction itself.

iamralpht (approved):

PR #639, opened 2023-12-15
Closes #630: Rotated nodes don't show linear gradients properly

### LINKED ISSUE #630: Rotated nodes don't show linear gradients properly
Nodes in Figma with a linear gradient background that are also rotated don't show the gradient properly. This is probably an issue for nodes with other types of backgrounds also. The new layout system calculates layout after transformations, whereas the old layout system calculated layout after transformations, so we probably need to account for that.

rylin8:
When getting the brush to render the background of a node, use same size as the one used when rendering the node. Consolidate the code that calculates this size into a function in Utils. This ensures that the linear gradient parameters specified in Figma match the node size. This also fixes an issue with angular gradients that regressed in the "Vector rendering" test, and can be seen in the new render example image.

Update serialized file version, dcf files, and images of tests that have changed.


rylin8:
This is targeted on another PR since they are somewhat related and I'd like them to get in at the same time. https://github.com/google/automotive-design-compose/pull/602

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -1245,6 +1245,22 @@ internal fun StrokeWeight.right(): Float {
 internal fun ViewStyle.isAutoWidthText() =
     width is Dimension.Auto && horizontal_sizing !is LayoutSizing.FILL
 
+internal fun getNodeRenderSize(
```
iamralpht, line 1248: Can you add a comment documenting the different scenarios here? I wasn't sure what all the cases were that `getRect` was handling previously were, and it took me a while to find the regressions when I changed it. (I was animating the layout size, and was surprised when some elements didn't animate).

rylin8: Done

iamralpht (approved):

iamralpht (approved):

Issue #65, opened 2023-05-18
Add NetworkAccess field to the Figma Plugin manifest


timothyfroehlich:
https://www.figma.com/plugin-docs/manifest/#networkaccess

According to Figma's announcement about it:

>This field allows you to specify the domains your plugins and widgets need access to, and we’ll surface this information on your resource’s community page. Resources that add this field to their manifest.json will either say “No,” “Restricted,” or “Unrestricted” network access on their community pages, while resources that don’t use this field will say “Unknown network access” 

>Limiting network access in this way to just the domains that are needed for your plugin or widget to work should increase the likelihood that more security-conscious users and organization admins will trust and use your resource. This is especially important for organizations that have [approved lists for plugins and widgets](https://help.figma.com/hc/en-us/articles/4420945686807-Approve-organization-widgets), which require admin approval before their organization members can only use plugins and widgets, and in particular many organization admins are especially interested in knowing if and how plugins and widgets access the network when determining whether to approve them for use.

This seems like a good idea to add.

PR #641, opened 2023-12-17
support-figma: add allowlisted domains for Figma plugins

jessesomerville:
This change defines the `networkAccess` field in the Figma plugin and widget's `manifest.json` files. This will prevent cross-origin requests to domains not matched by any item in the `allowedDomains` field and will update their Community pages to reflect that network access restrictions are in place (as opposed to "Unknown network access").

There are currently only two cross-origin requests being made. The first fetches the [figma-plugin-ds](https://github.com/thomas-lowry/figma-plugin-ds) stylesheet and the other(s) are made by the stylesheet in an attempt to load some Inter fonts from rsms.me using `@font-face`.

Figma pre-loads Inter for plugins to use making it unnecessary to fetch them again so I've intentionally excluded those domains from the allow-list. There are a few errors that show up in the dev console due to the CSP blocking the requests to fetch the fonts, but those can be safely ignored[^1].

I also added a section in the README's that calls out the network access restrictions.  The other changes were all formatting fixes applied by Prettier.

Fixes google/automotive-design-compose#65

[^1]: Technically only the Regular and Bold styles are being pre-loaded, but Figma appends `https://static.figma.com/webfont/` to the `font-src` directive in the CSP and the rest of the font family can be fetched from there if need be. This isn't officially documented anywhere as far as I can tell so it's probably better to just expand the allowlist if other fonts are needed.

### DIFF
--- a/support-figma/auto-content-preview-widget/README.md
+++ b/support-figma/auto-content-preview-widget/README.md
@@ -41,10 +41,25 @@ We recommend writing TypeScript code using Visual Studio code:
 That's it! Visual Studio Code will regenerate the JavaScript file every time you save.
 
 ## Adding the widget to Figma
+
 To add the development version of the widget, you need to be running the Figma desktop application on a Windows or MacOS machine.
 
-* In a doc, right-click and open the widget menu.
-* Select "Development"
-* Select "Import widget from manifest"
-* Navigate to your `manifest.json` and press OK.
-* Once you've done this once, the widget will be available directly in the "Development" menu option.
+- In a doc, right-click and open the widget menu.
+- Select "Development"
+- Select "Import widget from manifest"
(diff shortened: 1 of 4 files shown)


rylin8:
Is the target branch for this correct? I expect it to be [main](https://github.com/google/automotive-design-compose/tree/main) -- not sure why it says google:main

timothyfroehlich:
It is. It's because he had created his own fork of the repo to work on and to submit the change from. It's a normal GitHub workflow for external contributors: https://docs.github.com/en/get-started/exploring-projects-on-github/contributing-to-a-project

rylin8:
LGTM, thanks for doing this!

rylin8 (approved):

PR #671, opened 2023-12-28
Update AGP to 8.3.0-beta01

timothyfroehlich:
This will require the installation of the Android Studio Iguana Beta build


rylin8:
Does that mean I can't update through the Android Studio UI flow, and have to manually download?

timothyfroehlich:
Yes. You can start the update flow and it'll send you to the site. 

The good news is that the automatic update flow should work (it's only supposed to force a new install when upgrading from one release channel (canary/beta/stable) to another) and the Beta channel doesn't update that often.

rylin8:
What is the goal of this change? To get something more stable?

timothyfroehlich:
> What is the goal of this change? To get something more stable?

1. To stay up to date with the latest bug fixes and security updates for a pre-release of a dependency (the dependency being AGP).
2. The current committed version of AGP is incompatible with currently available builds of Android Studio. The beta release of AGP that we are upgrading to is compatible with the current beta build of Android Studio Iguana.

Reminder that we made the decision to go to the Canary version of AGP last year to take advantage of bug fixes and new features that are used in our Gradle plugin.

rylin8 (approved):
FYI I was able to build/run this without an Android Studio upgrade, 2023.2.1 Canary 18

Issue #458, opened 2023-10-13
RustLayout doesn't forward Modifiers


timothyfroehlich:
We'll need to figure out some way to make it work.

PR #674, opened 2023-12-28
Closes #396, #458: Support layout modifiers and propagate user-specified modifiers

### LINKED ISSUE #396: Modifiers that change the size or position don't work
Modifiers that change the size or position, such as Modifier.fillMaxSize(), Modifier.offset(), currently don't work because they are not passed into the layout system. Perhaps we need to make our own modifiers whenever it affects layout, i.e. fun Modifier.dcFillMaxSize()?

rylin8:


iamralpht:
Exciting to see the constraints tests getting stretched out again!

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -575,9 +591,23 @@ internal fun designMeasurePolicy(view: View, layoutId: Int) =
...
         val myWidth =
-            if (constraints.hasFixedWidth) constraints.maxWidth else myLayout?.width() ?: 0
+            if (constraints.hasFixedWidth && constraints.maxWidth != 0) constraints.maxWidth
+            else myLayout?.width() ?: 0
         val myHeight =
-            if (constraints.hasFixedHeight) constraints.maxHeight else myLayout?.height() ?: 0
+            if (constraints.hasFixedHeight && constraints.maxHeight != 0) constraints.maxHeight
+            else myLayout?.height() ?: 0
+
+        // If constraints have forced a size that does not match our layout size, set this size into
```
iamralpht, line 600: Oh neat - so `Modifier.fillMaxSize` etc just work directly because they set `hasFixed{Width|Height}` and `max{Width|Height}`? That's cool!

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -575,9 +591,23 @@ internal fun designMeasurePolicy(view: View, layoutId: Int) =
...
+        // the layout manager so that it can set this size in the Rust layout manager and then
+        // recalculate any layouts that many have changed. However, only do this if the layout state
+        // has not changed after computing child layouts. This is necessary in the case where
+        // calling setNodeSize on a node causes an ancestor to resize.
+        val layoutState2 = LayoutManager.getLayoutState(layoutId)
+        if (
+            layoutState == layoutState2 &&
+                (myWidth != myLayout?.width() || myHeight != myLayout?.height())
+        )
+            LayoutManager.setNodeSize(layoutId, rootLayoutId, myWidth, myHeight)
```
iamralpht, line 610: Does this end up causing a second layout? Or is this only applied in rendering?

iamralpht: I guess I mean: does this mean we have to run another iteration of Rust layout? I think we're consulting the Rust-computed layout values above, and then we're calling `setNodeSize`, which I think doesn't take effect until the next iteration of layout.

I'm trying to figure out if there's any recursive stuff here, which might result in more than one additional layout iteration being needed (like Rust layout generates a value, it is used to calculate incoming Constraints because a parent uses the value for child placement, the child then uses it via incoming Constraints... it looks like the child is always placed using a value directly from Rust layout, so I think that means if one layout iteration causes a node size to change, then the children might get a new size on the next iteration, which could cause a further layout iteration if one of those children has a layout modifier.

I guess the layout modifier needs to know the computed size -- so we can't create a pretend size and run it through the Modifier up front. I'm wondering if there's any way to make this more efficient or get it down to a single pass, or if we have to define our own Modifiers that are analogues to the Compose ones (which would be a pity from an API perspective).

rylin8: I'm still testing this which is why it's in a draft state, but I think it probably does cause a second layout pass. If say, a call to `fillMaxSize()` causes `#stage` to resize from 200x200 to 400x400, it Rust layout will compute it first as 200x200. Then the layout function will see that constraints have changed it to 400x400, so it calls `setNodeSize(400, 400)`. This calls into Rust layout and recomputes layout for children, then sends the new layouts back. However by this point, the layout for `#stage` has already measured all children with the old layouts. The new layouts with a new `layoutState` ID will then cause all nodes that changed to call layout again.

timothyfroehlich (changes requested):
Can you confirm that the screenshot changes were expected?

rylin8:
> Can you confirm that the screenshot changes were expected?

Yep, I spent a lot of time looking at the screenshots and fixed a lot of issues due to unexpected changes, so thanks for putting that together it has been very useful.

timothyfroehlich (commented):
Kotlin looks fine to me

timothyfroehlich (commented):
Kotlin looks fine to me

designcompose/src/main/java/com/android/designcompose/Layout.kt
iamralpht: We discussed online recently and thought it might be possible to use the measure func to glue the measurement of a custom composable (or customized child) into the rust layout world in a single pass, to avoid the complexity and performance issue with a multi-pass solution. I think there's also nothing preventing us from rendering a frame that's mid-way through layout, here, too, which isn't ideal.

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -256,16 +261,25 @@ internal object LayoutManager {
         return layoutCache[layoutId]
     }
```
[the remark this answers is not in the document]
rylin8: Done

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -256,16 +261,25 @@ internal object LayoutManager {
...
         val adjustedWidth = (width.toFloat() / density).roundToInt()
         val adjustedHeight = (height.toFloat() / density).roundToInt()
         val responseBytes =
             Jni.jniSetNodeSize(managerId, layoutId, rootLayoutId, adjustedWidth, adjustedHeight)
         handleResponse(responseBytes)
     }
 
+    internal fun hasModifiedSize(layoutId: Int): Boolean {
+        return modifiedSizes.contains(layoutId)
+    }
```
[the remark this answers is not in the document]
rylin8: Done, did to a few functions

timothyfroehlich (approved):

timothyfroehlich (approved):
LGTM

rylin8:
Found an issue with list widgets that are setup as rows or columns with scrolling turned on. I will hopefully have a fix for this tonight and will update this PR.

rylin8:
> Found an issue with list widgets that are setup as rows or columns with scrolling turned on. I will hopefully have a fix for this tonight and will update this PR.

Fixed.

designcompose/src/main/java/com/android/designcompose/Layout.kt
rylin8: As we discussed offline, this is not possible with the way the code is currently structured because if a child changes, there's no way for it to force the root node to run layout again.

iamralpht (approved):

PR #696, opened 2024-01-08
squoosh: wip: compute layout in Compose's layout phase, and consult i…

iamralpht:
…ncoming constraints.

This change regresses animation interruption in the "SA: Basics" and
"SA: Instances" tests. This will need to be fixed before merging.

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
+++ b/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
@@ -25,7 +25,6 @@ import androidx.compose.animation.core.VectorConverter
 import androidx.compose.animation.core.spring
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
@@ -44,7 +43,6 @@ import androidx.compose.ui.platform.LocalFontLoader
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.unit.Constraints
 import androidx.compose.ui.unit.Density
-import androidx.compose.ui.unit.dp
 import com.android.designcompose.AnimatedAction
 import com.android.designcompose.ComponentReplacementContext
 import com.android.designcompose.CustomizationContext
@@ -64,13 +62,18 @@ import com.android.designcompose.common.DocumentServerParams
 import com.android.designcompose.doc
 import com.android.designcompose.rootNode
 import com.android.designcompose.rootOverlays
+import com.android.designcompose.serdegen.Dimension
 import com.android.designcompose.serdegen.Layout
 import com.android.designcompose.serdegen.NodeQuery
+import com.android.designcompose.serdegen.Size
 import com.android.designcompose.squooshAnimatedActions
 import com.android.designcompose.squooshCompleteAnimatedAction
 import com.android.designcompose.squooshFailedAnimatedAction
 import com.android.designcompose.squooshVariantMemory
 import com.android.designcompose.stateForDoc
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
 import kotlin.math.roundToInt
 
 const val TAG: String = "DC_SQUOOSH"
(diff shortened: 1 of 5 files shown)


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -478,6 +417,110 @@ fun SquooshRoot(
                     animationValues,
                 ),
             measurePolicy = { measurables, constraints ->
+                // Update the root node style to have the incoming width/height from our parent
+                // Composable.
+                root.applyLayoutConstraints(constraints, density)
+
+                // Perform layout on our resolved tree.
+                layoutTree(root, rootLayoutId, layoutIdAllocator, layoutCache, layoutValueCache)
```
iamralpht, line 425: This is the main point of the change; evaluating layout happens in Compose's `measure` phase. That means we have the incoming constraints already, which are applied to the root's style on L422, so we can do layout in one phase.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -478,6 +417,110 @@ fun SquooshRoot(
...
+                        TAG,
+                        "Updating animations resulted in ${nextAnimations.size} animations"
+                    )
+                }
+
+                val currentAnimationJob = animationJob.animationJob
+                val needsAnimationJob = interactionState.animations.isNotEmpty() || variantTransitions.transitions.isNotEmpty()
+                val hasAnimationJob = currentAnimationJob != null && currentAnimationJob.isActive
+                if (needsAnimationJob && !hasAnimationJob) {
+                    animationJob.animationJob = interactionScope.launch(start = CoroutineStart.UNDISPATCHED) {
```
iamralpht, line 482: I think I should be able to remove the "animation <-> compose" glue out of this file more easily now -- before it was a bunch of logic that had to be in a `LaunchedEffect`, but now it's more easily movable because it just needs a `CoroutineScope`.

In general in Squoosh, I am seeing that there's a bunch of code that doesn't care about Compose at all (it's just resolving variants, building trees, evaluating layout, computing animated transitions, etc) and then some glue code that uses all the Compose APIs. I'm hoping to enforce that separation more strictly, and then move the generic code to Rust.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -478,6 +417,110 @@ fun SquooshRoot(
...
+                                            variantTransitions.completedAnimatedVariant(anim.variant)
+                                    }
+                                }
+                                animationValues.value = animState
+                            }
+                        }
+                    }
+                }
+
                 val placeables = measurables.map { measurable ->
```
iamralpht, line 524: This block, where we take `measurables` and generate `Constraints` for them from the output of layout should be upgraded. There are two types of `@Composable` child node for a `SquooshRoot`:

 1. Any node that uses interaction (or in the future that sets an accessibility property) causes a `@Composable` child to be generated. These synthetic `@Composable`s have no appearance, they're just there to capture events.
 2. Any `@Composable () -> Unit` customization results in a child `@Composable`. These are non-synthetic `@Composable` children.

The synthetic `@Composable`s can continue to be placed after layout has been evaluated. They just need to go exactly where the visual content is rendered from the `View` tree.

The non-synthetic `@Composable`s should probably be placed directly by Rust layout via a custom measure func. Rust Layout accepts a measure func for a node, and it passes some constraints and gets a `Size` back again. This is almost the same as what a Compose `Measurable` does -- it takes `Constraints` and gives you a `Placeable` with a width and height (the `measuredSize`). I think if we use Rust Layout's measure func to call `measurable.measure` for us, then we'll be able to support custom child composables that cause their DC Row/Column/Whatever to resize as they change size, or that can split text into multiple lines and occupy the correct height.

iamralpht:
@rylin8 I had a go at moving layout evaluation to be in the Compose layout phase, instead of during Compose's composition phase. It worked out in Squoosh, although there is a regression in interrupted interaction-based animated transitions (if you tap the same shape over and over in "SA: Basics" then you'll see it doesn't interrupt properly; I think I'm messing up the layout cache by computing the transitionRoot before I've run layout on the basic root, but I should be able to fix this...).

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
iamralpht: Upon reflection, I think this approach can only work for a Layout that is the direct parent of the items that need to be custom measured. So it works for `SquooshRoot` in its current incarnation, but if we ever needed to change `SquooshRoot` so that it contained more of a tree (instead of squooshing everything into a single composable). I don't think it's possible to get measurables from deeper in the tree than immediate children.

iamralpht:
Updated with animation regression fixed. I think there's no harm in merging this into "squoosh-4", since that is just a user branch currently.

PR #706, opened 2024-01-10
Closes #705: Add a function to enable overriding document IDs

### LINKED ISSUE #705: Support setting the Doc id when the application starts
Today, Design compose supports switching between the doc Ids at runtime using the design switcher tool when the Live updates are enabled.

But there is NO support to change the default ID when the application starts. For example, say we have 2 doc ids abc and xyz what is needed is that when we launch target gcar_x86_64 it picks up abc and when we launch target gcar_dd_x86_64 it picks xyz

Since the runtime support for changing the docId is already there on Design Compose side if we have an API to set the docId at the start of application maybe in onStart that would suffice this design request. Or if its possible to assign a non-static variable in the annotation @DesignDoc(id = variable)

Created from https://buganizer.corp.google.com/issues/319161864

rylin8:
Add a function DesignDocOverride(id) that can be used to override the document ID used by any child/descendent generated DesignCompose functions. DesignDocOverride internally just uses a CompositionLocalProvider to provide the document ID, which is checked and used in the DesignDocInternal function.

Add a HelloBye example that uses this function to change the document ID.


timothyfroehlich (commented):
To the extent that this unblocks a partner, this looks fine, however we are opening ourselves up to a lot of problems. Use of the function will break apps that use multiple root documents, introducing a bug that users will be unable to root cause or fix themselves. The function isn't documented in our main documentation either.

IMO there should be strong language related to the fuction in the comments to discurage use. I think the [@Discouraged](https://developer.android.com/reference/androidx/annotation/Discouraged) annotaiton should be applied to it.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -778,6 +778,16 @@ internal val LocalDesignIsRootContext = compositionLocalOf { DesignIsRoot(true)
 // Current customization context that contains all customizations passed through from any ancestor
 val LocalCustomizationContext = compositionLocalOf { CustomizationContext() }
 
+// Current document override ID that can be used to override the document ID specified from the
+// @DesignDoc annotation
+internal val LocalDocOverrideContext = compositionLocalOf { String() }
+
+// Public function to set the document override ID
+@Composable
```
timothyfroehlich, line 786: Add the Discouraged annotation:
https://developer.android.com/reference/androidx/annotation/Discouraged

iamralpht:
I agree it would be good to comment the limitations and annotate the API appropriately, as Tim suggests.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
rylin8: Done

timothyfroehlich (approved):
LGTM

timothyfroehlich (approved):
LGTM

iamralpht (approved):

PR #720, opened 2024-01-12
Closes #397: Implement scrolling support for autolayout containers

### LINKED ISSUE #397: Autolayout containers with scrolling don't scroll
Autolayout containers that have a sizing mode other than Hug with scrolling enabled in Figma don't scroll in Android. Scrolling works in the list widget because that is implemented using Row() or Column(), but non-widget autolayout frames are implemented using Box(). To support scrolling with a box, we may need something like this:

Box(scrolling = true) { Box(contents) }

rylin8:
Use Modifier.scrollable() to enable scrolling on autolayout frames. Handle the scrolling logic manually by keeping track of the scroll offset and passing it to layout. The layout function also sets the extents of the scrollable frame's children in order to establish the scrollable bounds.

Note that scrolling to the end of a list does not have the same overscroll bounce effect that native scrollable containers have.

Cleaned up unused fields and data structures for text measurement

Optimized text remeasurement by only resubscribing to layout if data in the text changes. Previously, it was resubscribing too often because ParagraphIntrinsics is not a data class and its equality comparison would return false, triggering the DisposableEffect that subscribes to layout to run unnecessarily.


timothyfroehlich (changes requested):
You've got some merge conflicts, plus it seems like the screenshot diff didn't re-run when the base branch changed (which makes sense, I don't think I had the action set up to handle that). I'm manually re-running it.

timothyfroehlich:
Looks like the update comment job of the screenshot comparison is failing, probably because the base branch changed. I suspect it'll work again once a new commit is pushed.

crates/layout/tests/layout_tests.rs
```diff
@@ -343,6 +343,105 @@ fn test_horizontal_fill_resize() {
...
+    // Change root node size and check that child stretches correctly
+    let result = manager.set_node_size(0, 0, 200, 200);
+    assert!(result.changed_layouts.contains_key(&1));
+
+    let child_layout_result = manager.get_node_layout(1);
+    assert!(child_layout_result.is_some());
+    let child_layout = child_layout_result.unwrap();
+    assert!(child_layout.width == 150.0);
+    assert!(child_layout.height == 150.0);
+}
```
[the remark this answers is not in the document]
rylin8: Fixed

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -541,81 +568,116 @@ internal inline fun DesignFrameLayout(
...
+                val childLayout = LayoutManager.getLayoutWithDensity(layoutData.layoutId)
+                if (childLayout == null) {
+                    Log.d(TAG, "Place error null layout: $name child $index layoutId $layoutId")
                 } else {
-                    val childLayout = LayoutManager.getLayoutWithDensity(layoutData.layoutId)
-                    if (childLayout == null) {
-                        Log.d(TAG, "Place error null layout: $name child $index layoutId $layoutId")
-                    } else {
-                        placeable.place(x = childLayout.left(), y = childLayout.top())
+                    var hScrollOffset = 0
```
iamralpht, line 647: I was surprised to see that Compose seems to [implement the scroll offset using layout, too](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/Scroll.kt).

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -541,81 +568,116 @@ internal inline fun DesignFrameLayout(
...
+                                        .coerceAtLeast(0F)
+                            } else {
+                                val vMargin = view.style.padding.bottom.pointsAsDp(density).value
+                                designScroll.scrollMax.value =
+                                    (childLayout.top + childLayout.height - myHeight + vMargin)
+                                        .coerceAtLeast(0F)
+                            }
+                        }
                     }
+                    placeable.place(
```
iamralpht, line 672: Compose's own scrollable implementation uses [`placeable.placeWithLayer`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/Scroll.kt;l=396) which means that the rendering of the children will be cached in a display list (which seems like a good choice for these containers because most likely the children aren't changing much if at all).

iamralpht (changes requested):
Please try with `placeable.placeWithLayer` as it aligns better with the Compose implementation.

rylin8:
> Please try with `placeable.placeWithLayer` as it aligns better with the Compose implementation.

Done. I also noticed that the text in DesignSwitcher was triggering layout frequently, and found that `ParagraphIntrinsics` was triggering the `DisposableEffect` even when the text or style did not change. I optimized this by only resubscribing to layout if data in the text changes.

designcompose/src/main/java/com/android/designcompose/Layout.kt
rylin8: Done

iamralpht (approved):
Excellent, thanks for digging in.

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -490,7 +484,22 @@ internal fun calcLayoutInfo(
...
+        var horizontalScroll = false
+        var verticalScroll = false
+        when (view.scroll_info.overflow) {
+            is OverflowDirection.VERTICAL_SCROLLING -> {
+                verticalScroll = true
+            }
+            is OverflowDirection.HORIZONTAL_SCROLLING -> {
+                horizontalScroll = true
+            }
+            is OverflowDirection.HORIZONTAL_AND_VERTICAL_SCROLLING -> {
```
timothyfroehlich, line 496: We don't have any way of surfacing this to the user, right? That we don't support both?

timothyfroehlich (approved):

PR #750, opened 2024-01-25
Closes #746: Fix Design switcher error message layout sizes

### LINKED ISSUE #746: Design switcher error messages are laid out with the wrong size
The text has the correct size, but its parent or ancestor has too small of a height, so the message ends up overlapping other messages in the status list. I have been unable to reproduce this in a testing environment.
Screenshot 2024-01-24 at 5 01 11 PM

rylin8:
The #MessageFailed node was getting an incorrect height even though all its children were correct. Downloading a new version of the node fixes the issue.


rylin8:
How it looks with this change: Screenshot 2024-01-25 at 1 52 11 PM

rylin8:
I spent hours trying to figure out why this was happening but all test scenarios resulted in the correct layout being computed. I made a branch of the Design Switcher Figma doc and started playing around with the layout of `#MessageFailed`, and at some point it started working. I undid my changes, but it still worked. I compared the node data of the branch with the original and found that the only difference was in the "overrides" section below. Even so, I don't quite understand what changed since I don't see a difference in the overridden fields. My guess is that the Figma dev REST api changed slightly and my updates triggered a small change to the response that fixes the layout.
```
            "overrides": [
              {
                "id": "I2:152;2:144",
                "overriddenFields": [
                  "sharedPluginData",
                  "pluginData"
                ]
              },
              {
                "id": "2:152",
                "overriddenFields": [
                  "counterAxisSizingMode",
                  "name"
                ]
              },
              {
                "id": "I2:152;2:143",
                "overriddenFields": [
                  "layoutAlign"
                ]
              }
            ]
```

timothyfroehlich (commented):
Odd. I guess if it works it works!

iamralpht (approved):

PR #754, opened 2024-01-26
Closes #753: Fix variant resolution with components instances whose names differ from their component set.

### LINKED ISSUE #753: Variant resolution gets confused with multiple components that use the same properties
Our variant search algorithm currently works by taking all the variants set from @DesignVariant annotations and looking for matches in the Figma document. However if there are multiple components that have the same properties, variant resolution sometimes picks the wrong component and cannot resolve the variant.

For example, we have a `GreenComponent` and a `BlueComponent`, each of which use three properties: `#scheme`, `#border`, and `#shade`. The variant resolution algorithm looks for a component to match the node `#scheme=blue, #border=square, #shade=three`. The component instance is an instance of `BlueComponent` so it should be easy to find, but instead the algorithm looks in `GreenComponent` because it. has the same number of matching properties, and fails to find a variant.

The algorithm should not need to look at other components since it knows that the component being resolved is an instance of `BlueComponent`.
Screenshot 2024-01-26 at 10 18 00 AM
Screenshot 2024-01-26 at 10 18 09 AM

rylin8:
Deleted a bunch of code that would look for the component set to use given the variant properties and values being used. Instead, just use the component set name associated with the component instance being composed. Added more to the VariantProperties test to test this.


rylin8:
It's been awhile since doing anything with the variant resolution algorithm. Here I deleted a bunch of code that would look for the component set to use given the variant properties and values being used. I'm instead using the component set name associated with the component instance being composed. I'm not sure why the deleted code was ever needed -- am I forgetting about a specific use case or was this just overcomplicated code? Media looks fine and all the Validation tests look correct too.

timothyfroehlich (commented):
This is probably a good time to add a set of screenshot tests for VariantProperties. `integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/InteractionTriggers.kt` would work as a good template.

rylin8:
> This is probably a good time to add a set of screenshot tests for VariantProperties. `integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/InteractionTriggers.kt` would work as a good template.

Tests added!

timothyfroehlich:
I love it!

timothyfroehlich (approved):

iamralpht:
Oh good - this code was also showing up in the profile for Battleship last time I ran, and now we also don't need to migrate it to Rust... many benefits!

rylin8:
> Oh good - this code was also showing up in the profile for Battleship last time I ran, and now we also don't need to migrate it to Rust... many benefits!

Showing up meaning it was slow?

iamralpht (approved):

iamralpht:
> > Oh good - this code was also showing up in the profile for Battleship last time I ran, and now we also don't need to migrate it to Rust... many benefits!
> 
> Showing up meaning it was slow?

Yes - it was slow, but really only in aggregate (which Battleship shows)

PR #774, opened 2024-02-01
Closes #714: Create option in dials & gauges plugin to make vertical progress bar

### LINKED ISSUE #714: Create option in dials & gauges plugin to make vertical progress bar
Also do this for progress markers.
Needs design from @aubreydagoog.

rylin8:
Add a horizontal/vertical option for progress bars and progress markers, and render the bar/marker accordingly when moving the progress slider. Save additional data into the shared plugin data for these nodes and update the Compose rendering code to support vertical mode.

The Dials and Gauges Validation test has been updated to include some vertical progress bars and markers, including skewed ones as well.

Screenshot 2024-02-01 at 2 32 28 PM


support-figma/extended-layout-plugin/src/code.ts
```diff
@@ -949,27 +949,48 @@ else if (figma.command === "move-plugin-data") {
...
     }
     
     let meterData: any = {};
     if (progressMarker) {
       // The progress marker means we don't resize the node; we just move it along the x axis
-      const startX = -node.width / 2;
-      const endX = parent.width - node.width / 2;
-      const moveX = percentToValue(value, startX, endX);
-      node.x = moveX;
+      // if horizontal or y axis if vertical
```
timothyfroehlich, line 958: Especially here, we should try to make this as generic as we can.

Also, is there a way that we can make it completely generic and not dependent on "height" and "width"? Could there be a way for the user to designate the start and end points within the node? I really don't know if that's feasible, but if there were something like that I think it'd be much more intuitive.

timothyfroehlich (changes requested):

timothyfroehlich (changes requested):
Would love to see some more Validation tests here too.

support-figma/extended-layout-plugin/src/code.ts
rylin8: Making this completely generic would be hard, especially taking into account the fact that you can transform these rectangles and frames in Figma. I think what we have is sufficient for most if not all progress bar scenarios currently. Sure there may be some complicated and obscure scenarios we don't support yet. If we really want to revamp how this is done we should get @aubreydagoog and/or Richard involved with a redesign. I was just trying to get this in quickly since it's a relatively simple feature that would have a lot of benefit since multiple users have had issues building vertical progress bars with our current plugin.

rylin8:
> Would love to see some more Validation tests here too.

I tried adding some tests that set the progress bar/indicator value, then check that the node has the appropriate width, height, or position using functions like `assertLeftPositionInRootIsEqualTo()` or `assertWidthIsAtLeast()`. However they don't seem to work -- they always report the size or position of the nodes as they are set in the Figma document. I think the way that the progress bar sizes or positioned are modified in `FrameRender.kt` are not reflected in the test framework.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -150,16 +150,28 @@ private fun calculateRotationData(
...
     val discretizedMeterValue =
         meterValue.coerceDiscrete(progressBarData.discrete, progressBarData.discreteValue)
 
-    // Resize the progress bar by interpolating between 0 and endX
-    val barWidth = lerp(0F, progressBarData.endX, discretizedMeterValue, density)
-    return Size(barWidth, height)
+    // Resize the progress bar by interpolating between 0 and endX or endY depending on whether it
+    // is a horizontal or vertical progress bar
+    if (progressBarData.vertical) {
+        val width = style.width.pointsAsDp(density).value
```
iamralpht, line 163: This is the same logic as it was before, but I think we should be using layout produced dimensions here, and not consulting style. This approach won't work if the meter is sized using constraints or flexbox, and we're at a point where we have already computed layout, and have the layout info available.

Either we should update in this PR or make a new one to migrate to using layout produced values (I think almost everything in FrameRender should use layout produced values, except the appearance stuff like fill/stroke style -- this is where it would be super nice in the future to tease out ViewStyle into layout bits and presentation bits, because they get consumed at different points).

support-figma/extended-layout-plugin/src/code.ts
iamralpht: I think the "totally generic version" would just be growing a path. We've previously discussed using the stroker to generate a single dash that grows/shrinks along a path (where only the dash is filled). This would create the illusion of the path growing/shrinking, but it requires some careful attention on things like supporting stroke caps (rounded, square, butt, etc).

Currently, I think we still need all of the various options, so I think this change is a good add. In the future, we might be able to make the path be a default, but we'd need to be able to support moving a marker along the path, and we'd need it to feel acceptable to the person operating Figma.

iamralpht:
@rylin8 I like this change, but there are some conflicts, so I can't approve. I do really want FrameRender to not use layout style if possible. I wonder if using the layout computed values would also help with the test cases that weren't working?

rylin8:
> @rylin8 I like this change, but there are some conflicts, so I can't approve. I do really want FrameRender to not use layout style if possible. I wonder if using the layout computed values would also help with the test cases that weren't working?

What test cases that weren't working?
I'll look into using layout values -- from what I recall, there was an issue with doing that which is why I reverted back to this, but I can't recall why.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: I think the reason I didn't end up using the layout width/height was that those values are calculated after transformation. So a horizontal progress bar with a small height that has been rotated will have a much taller height after the transformation, and layout will return that taller height. Here I need the height before transformation because we apply the transformation to the draw context after determining the size of the rect. Does that make sense? Can you see another way of doing this that isn't too complicated?

support-figma/extended-layout-plugin/src/code.ts
rylin8: Moving along a path would be great! We'll add it to the list!

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -254,13 +272,10 @@ internal fun ContentDrawScope.render(
...
                     val progressBarData = meterData.value
                     if (progressBarData.enabled) {
-                        rectSize =
-                            calculateProgressBarData(
-                                progressBarData,
-                                meterValue,
-                                style.height.pointsAsDp(density).value,
-                                density
-                            )
+                        val progressBarSizeTransform =
```
timothyfroehlich, line 275: (Style) This would be cleaner to [destructure](https://kotlinlang.org/docs/destructuring-declarations.html) the result, right?

timothyfroehlich (commented):
I think we're good. Do you have the plugin changes staged anywhere?

Also, just to confirm, the screenshot changes are correct / expected, right?

timothyfroehlich (approved):

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: Yea good point. I actually remember doing that, but either I lost it or it didn't work because I had incorrect syntax. I don't want to block this PR so I will make a note to fix this in another PR. I have a couple incoming.

rylin8:
> I think we're good. Do you have the plugin changes staged anywhere?
> 
> Also, just to confirm, the screenshot changes are correct / expected, right?

Yes screenshot changes expected. I didn't stage them -- is there somewhere we normally do that? I thought normally we just pull the changes down and then build them on our own machines if we want to use the latest developer version. Otherwise we'll publish them to a new version.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: I remember now -- the variables are already declared so destructuring isn't supported. https://discuss.kotlinlang.org/t/reusing-vars-in-destructuring-declarations/4077

PR #778, opened 2024-02-02
Add BSD-3 Clause license to cargo deny

timothyfroehlich:
An update that went in with dependabot added a dependency that uses it.


iamralpht (approved):
Permitted per go/thirdpartylicenses

PR #809, opened 2024-02-20
Closes #739: Fix support for inheritance with DesignCompose interfaces

### LINKED ISSUE #739: Design Component must be within the main document Figma interface
When creating a reusable Design Component there seems to be no way to avoid adding the nodes declared for the interface to the main figma document.  When adding a separate interface it is required to add **CustomComponent** implementation or it will not compile. I've attached a project with a Design Component Indicator that is declared in interface **FigmaIndicator** which will compile but show up blank. If I move this to interface **ExampleDoc** it will work. I understand I could override this in the main **ExampleDoc** interface however that would require anyone using this component elsewhere would need to do the same.
[Ex4Google.zip](https://github.com/google/automotive-design-compose/files/14024563/Ex4Google.zip)

rylin8:
Add support for inheriting from generated interfaces. Doing so will make the child's generated queries() and ignoredImages() functions combine their values with the parent class's values. Add a check to ensure that the parent class uses the same doc ID as the child class. Supporting a parent with a different doc ID will need additional work.

Fix support for multiple inheritance. Collect a list of all parent classes and use them in the generated interface.

Remove an unused generated function.


integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/InheritanceTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.width
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import com.android.designcompose.ComponentReplacementContext
+import com.android.designcompose.annotation.Design
+import com.android.designcompose.annotation.DesignComponent
+import com.android.designcompose.annotation.DesignDoc
+
```
timothyfroehlich, line 28: I honestly have no idea what half of the code is doing here, what is `ParentNoDesignDoc` and what is it demonstrating? I think I see what `replaceIndicator` is doing, but the naming is odd (it'd be more clear if it was `indicatorToBeOverridden` or something). And I'd really like to see a test that shows the base interface working as well.

timothyfroehlich (changes requested):
I really need some documentation here before I can review. Something like this needs to be fully explained in the core documentation (docs/) with clear examples of how it works and what its limitations are.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/InheritanceTest.kt
rylin8: This is just showing that the interface `Child` can inherit from multiple interfaces. It inherits from both `ParentGen`, an interface that is generated, and `ParentNoDesignDoc`, an interface that is not generated. The generated `ChildDoc` object can then call functions from `Child`, `ParentGen`, and `ParentNoDesignDoc`. I can add what I've just written in a comment.

rylin8:
> I really need some documentation here before I can review. Something like this needs to be fully explained in the core documentation (docs/) with clear examples of how it works and what its limitations are.

We had support for interface inheritance before as shown in MediaCompose, but it didn't work in all scenarios. You couldn't inherit from DesignCompose generated classes and call its generated functions. This just fixes it so that you can inherit from interfaces as you would normally in Kotlin.

iamralpht:
Ok, we discussed this last week, and said that we'd prefer to entirely prevent inheritance-based extension and support a new mechanism of interface flattening instead. We're taking this path because:

 * The interface inheritance mechanism is clumsy -- you need to know a lot about DesignCompose's generated Composable format (which we should be able to change over time without breaking source compatibility) to make it work.
 * You still have to populate the entire child Composable in one place -- that means that even once we broke MediaCompose into a library, we still had a bunch of boilerplate to use it in any other project.

Flattening addresses these two issues and (hopefully) is easier to scope to predictable behavior (inheritance has multiple inheritance at different levels in the inheritance hierarchy, and other fun stuff, but flattening could have nested flattening, I guess an interface could be parameterized, etc... we probably need to have a check that errors on anything complicated like that...).

So I think we should park this for now, land the interface flattening, and then update MediaCompose accordingly.

timothyfroehlich:
@rylin8 I think it's time to close this.

rylin8:
Not supporting inheritance.

PR #814, opened 2024-02-22
Closes #813: Create a tool that outputs the version info of a .dcf file

### LINKED ISSUE #813: Create a tool that outputs the version info of a .dcf file

rylin8:
Output the file name, ID, version, and timestamp. Also do this in the fetch tool.

### DIFF
--- a/crates/figma_import/src/bin/dcf_info.rs
+++ b/crates/figma_import/src/bin/dcf_info.rs
@@ -1,0 +1,85 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Utility program to parse a .dcf file and print out the header and version info for the file.
+
+use bincode::Options;
+use clap::Parser;
+use figma_import::{SerializedDesignDoc, SerializedDesignDocHeader};
+
+#[derive(Debug)]
+struct ParseError(String);
+impl From<bincode::Error> for ParseError {
+    fn from(e: bincode::Error) -> Self {
+        eprintln!("Error during deserialization: {:?}", e);
+        ParseError(format!("Error during deserialization: {:?}", e))
+    }
+}
+impl From<std::io::Error> for ParseError {
+    fn from(e: std::io::Error) -> Self {
+        eprintln!("Error opening file: {:?}", e);
+        ParseError(format!("Error opening file: {:?}", e))
+    }
+}
+
+#[derive(Parser, Debug)]
+struct Args {
+    // Path to the .dcf file to deserialize
+    dcf_file: std::path::PathBuf,
+}
+
+fn dcf_info(args: Args) -> Result<(), ParseError> {
+    let file_contents = std::fs::read(args.dcf_file)?;
+    let bytes = file_contents.as_slice();
+
+    // Deserialize the header
+    let header: SerializedDesignDocHeader = bincode::deserialize(bytes)?;
+    let header_size = bincode::serialized_size(&header)? as usize;
+
+    println!("Deserialized header");
+    println!("  DC Version: {}", header.version);
+
+    let current_version = SerializedDesignDocHeader::current().version;
+    if header.version != current_version {
+        println!(
+            "Header version {} does not match current version {}, aborting",
+            header.version, current_version
+        );
+        return Err(ParseError("Version mismatch".to_string()));
+    }
+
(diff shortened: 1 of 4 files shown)


crates/figma_import/src/bin/dcf_info.rs
```diff
@@ -0,0 +1,65 @@
...
+impl From<std::io::Error> for ConvertError {
+    fn from(e: std::io::Error) -> Self {
+        eprintln!("Error opening file: {:?}", e);
+        ConvertError(format!("Error opening file: {:?}", e))
+    }
+}
+
+#[derive(Parser, Debug)]
+struct Args {
+    #[arg(short, long)]
```
timothyfroehlich, line 37: Could you add some basic usage statement here?

crates/figma_import/src/bin/dcf_info.rs
```diff
@@ -0,0 +1,65 @@
...
+    dcf_file: std::path::PathBuf,
```
timothyfroehlich, line 38: Can you set the argument to be positional so we don't need to specify --dcf-file? Just remove the short and long traits. It's the only argument so I don't think it makes sense to require the flag

crates/figma_import/src/bin/dcf_info.rs
timothyfroehlich, line 1: Can you add a short bit about it in a readme? Either in figma_import's readme or in the main one. `cargo run --bin dcf_info --features="dcf_info -- --dcf-file <file>" is a mouthful.

timothyfroehlich (changes requested):
I gave it a shot with the Tutorial file and it didn't work for some reason:

```
cargo run --bin dcf_info --features="dcf_info" -- --dcf-file reference-apps/tutorial/app/src/main/assets/figma/TutorialDoc_3z4xExq0INrL9vxPhj9tl7.dcf 
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/dcf_info --dcf-file reference-apps/tutorial/app/src/main/assets/figma/TutorialDoc_3z4xExq0INrL9vxPhj9tl7.dcf`
Error during deserialization: SizeLimit
dcf_info failed: ConvertError("Error during deserialization: SizeLimit")
```

rylin8:
> I gave it a shot with the Tutorial file and it didn't work for some reason:
> 
> ```
> cargo run --bin dcf_info --features="dcf_info" -- --dcf-file reference-apps/tutorial/app/src/main/assets/figma/TutorialDoc_3z4xExq0INrL9vxPhj9tl7.dcf 
>     Finished dev [unoptimized + debuginfo] target(s) in 0.06s
>      Running `target/debug/dcf_info --dcf-file reference-apps/tutorial/app/src/main/assets/figma/TutorialDoc_3z4xExq0INrL9vxPhj9tl7.dcf`
> Error during deserialization: SizeLimit
> dcf_info failed: ConvertError("Error during deserialization: SizeLimit")
> ```

Oops, good catch. I was testing with a file retrieved from our fetch tool. I forgot this, but our fetch tool serializes and saves a file without the 4 byte version header, but files retrieved through live update do. So all of the dcf files saved for Validation and other apps have the header that I was not parsing.

I've updated this to parse the header and compare it with the current version. If they don't match, I print an error and abort since deserialization will fail. If they match I continue with deserialization of the rest of the file and print out the info.

crates/figma_import/src/bin/dcf_info.rs
rylin8: Done

crates/figma_import/src/bin/dcf_info.rs
rylin8: Done

iamralpht:
> Oops, good catch. I was testing with a file retrieved from our fetch tool. I forgot this, but our fetch tool serializes and saves a file without the 4 byte version header, but files retrieved through live update do. So all of the dcf files saved for Validation and other apps have the header that I was not parsing.
> 
> I've updated this to parse the header and compare it with the current version. If they don't match, I print an error and abort since deserialization will fail. If they match I continue with deserialization of the rest of the file and print out the info.

Ooh, it sounds like we generate different files from fetch versus live update. Is that still the case? (And if so, can we fix it in another PR?)

rylin8:
> > Oops, good catch. I was testing with a file retrieved from our fetch tool. I forgot this, but our fetch tool serializes and saves a file without the 4 byte version header, but files retrieved through live update do. So all of the dcf files saved for Validation and other apps have the header that I was not parsing.
> > I've updated this to parse the header and compare it with the current version. If they don't match, I print an error and abort since deserialization will fail. If they match I continue with deserialization of the rest of the file and print out the info.
> 
> Ooh, it sounds like we generate different files from fetch versus live update. Is that still the case? (And if so, can we fix it in another PR?)

Yep https://github.com/google/automotive-design-compose/issues/816

timothyfroehlich:
Oh and don't forget to try to avoid pushing amended commits. It makes it harder for me to pull down and test new changes.

iamralpht (approved):
Looks good, especially with the header fix landing, too.

iamralpht:
> I gave it a shot with the Tutorial file and it didn't work for some reason:

@timothyfroehlich I think you can resolve your change request now, since we found/fixed the cause of the issue.

timothyfroehlich (approved):

PR #819, opened 2024-02-23
Closes #816: Update fetch tool to save dcf files with a header

### LINKED ISSUE #816: Fetch tool saves dcf files without the header
Files fetched via our fetch tool save files without the 4 byte version header. This caused some issues when DriverUI started doing live updates and sending the serialized file to HAR months ago. We fixed HAR to parse the header, but never fixed the fetch tool.

rylin8:
Also update the layout unit test to deserialize this header.


crates/figma_import/src/bin/fetch.rs
```diff
@@ -86,14 +90,18 @@ fn fetch_impl(args: Args) -> Result<(), ConvertError> {
         id: doc.get_document_id(),
     };
     println!("Fetched document");
+    println!("  DC Version: {}", SerializedDesignDocHeader::current().version);
     println!("  Doc ID: {}", doc.get_document_id());
     println!("  Version: {}", doc.get_version());
     println!("  Name: {}", doc.get_name());
     println!("  Last Modified: {}", doc.last_modified().clone());
     // We don't bother with serialization headers or image sessions with
```
timothyfroehlich, line 98: This comment is no longer accurate, correct?

crates/figma_import/src/bin/fetch.rs
```diff
@@ -86,14 +90,18 @@ fn fetch_impl(args: Args) -> Result<(), ConvertError> {
...
     // this tool.
-    let output = std::fs::File::create(args.output)?;
-    bincode::serialize_into(output, &serializable_doc)?;
+    let mut output = std::fs::File::create(args.output)?;
+    let header = bincode::serialize(&SerializedDesignDocHeader::current())?;
```
timothyfroehlich, line 101: Is there common code that can be reused here?

crates/layout/tests/layout_tests.rs
```diff
@@ -129,10 +129,14 @@ fn add_view_to_layout(
 fn load_doc() -> std::io::Result<SerializedDesignDoc> {
```
timothyfroehlich, line 129: This function looks like something that should be generic so that it can be re-used in other tests. But that's out of scope for this PR.

timothyfroehlich (commented):
Are you aware of anything that users of the Fetcher will need to do after this change is merged? It sounds like HAR already handles `.dcf`'s with and without the header, but is there anything else you can think of that still expects `.dcf`'s to not have a header?

crates/figma_import/src/bin/fetch.rs
rylin8: Updated comment.

crates/figma_import/src/bin/fetch.rs
rylin8: Not really -- it's just one line. If it gets more involved or used more I'll separate it out.

crates/layout/tests/layout_tests.rs
rylin8: Agreed.

timothyfroehlich (approved):

iamralpht (approved):
Approved; @urpatil @johnoneil fyi, this change ensures we have the same header in DCF files regardless of which tool created them.

@johnoneil this may help with reproducibility for DCF file fetches, but it likely also either fixes an issue where Harry couldn't open a DCF file, or introduces one (since somehow we were using `fetch` before, and it was emitting files with no headers).

PR #826, opened 2024-02-29
Closes #792: Support component replacement of absolutely positioned nodes.

### LINKED ISSUE #792: Replacing an absolutely positioned node does not preserve position
A `@Composable (ComponentReplacementContext) -> Unit` type customization on a node that is absolutely positioned (not in an autolayout) will not preserve the position of the original node. To achieve this, we need to somehow provide the position of the node being replaced to the replacement composable and have it use that instead of its own position.

rylin8:
When performing a component replacement, pass the original node's external layout properties into the replacement context. When this  exists in DesignFrame, replace the current ViewStyle's external layout properties with the one in the replacement context. This ensures that the replacement node uses the original node's layout and size properties.


rylin8:
This gets the job done, but I don't like it and I think maybe I'm just missing something obvious. When performing the component replacement, we need to pass some additional context to the replacement node -- in this case the offset from its parent. Is there additional layout info we should be passing, like the constraints? Or should that not be inherited when doing component replacement?

It doesn't feel quite right putting the original node's offset into `ParentLayout`, but I don't want introduce another parameter in `ComponentReplacementContext` that would require the developer to write code that uses it when doing a component replacement. They already have to pass in `it.parentLayout` which is annoying (though they used to have to use `it.layoutModifier`). I also don't want to introduce additional parameters to `DesignView`. So maybe `ParentLayout` is ok, but maybe it should be renamed to something more generic. And perhaps we should make it so that the developer does not need to pass in `it.parentLayout` into their component replacement node, using `CompositionLocalProvider`.

timothyfroehlich:
I don't like it either.

Are we sure that a better fix isn't feasible? Should we rethink ParentLayout now before we have too many more issues like this? 

One gut question I have is "Why does a component that can be replaced not already save it's original information?

I don't know if I can offer much more than that, I'm not familiar enough with the code there. I'm willing to take some time to learn it though. I'm really curious about the whole ViewStyle.asBuilder part. I don't understand what that's doing.

rylin8:
> I don't like it either.
> 
> Are we sure that a better fix isn't feasible? Should we rethink ParentLayout now before we have too many more issues like this?
> 
> One gut question I have is "Why does a component that can be replaced not already save it's original information?
> 
> I don't know if I can offer much more than that, I'm not familiar enough with the code there. I'm willing to take some time to learn it though. I'm really curious about the whole ViewStyle.asBuilder part. I don't understand what that's doing.

All the fields in `ViewStyle` are final because that's how serdegen makes them. So in order to change one field you have to do something like this. I took this from @iamralpht's squoosh branch since he already needed it there.

iamralpht:
> All the fields in `ViewStyle` are final because that's how serdegen makes them. So in order to change one field you have to do something like this. I took this from @iamralpht's squoosh branch since he already needed it there.

Yeah, the Java objects that serdegen creates don't cover all of the use cases. It's good that they are immutable, otherwise I'm sure we'd have bugs where we're mutating something that other objects have a handle on, but it's annoying that there isn't a generated "clone with these edits". In Rust you can just clone an immutable object and store the clone as a mutable reference, then make edits, then turn the mutable ref into an immutable one before you share it with the rest of the program.

Hopefully we can move to protobuf serialization soon, which has more ergonomic Java codegen ;)

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -590,6 +590,122 @@ internal fun mergeStyles(base: ViewStyle, override: ViewStyle): ViewStyle {
...
+// returning a new ViewStyle.
+internal fun ViewStyle.withExternalLayoutData(data: ExternalLayoutData): ViewStyle {
+    val overrideStyle = this.asBuilder()
+    overrideStyle.margin = data.margin
+    overrideStyle.top = data.top
+    overrideStyle.left = data.left
+    overrideStyle.bottom = data.bottom
+    overrideStyle.right = data.right
+    // Only override the width and height if not null
+    data.width?.let { overrideStyle.width = it }
```
iamralpht, line 698: heh, I think this is idiomatic Kotlin, but it took me a few tries to understand it...

iamralpht:
> Ok I have changed this to save all of the original node's external layout properties -- at least, my guess at what those are. It got a bit more complicated to support constraints because if certain constraints are set, the width/height gets set to something like Dimension::Percent(), in which case we want the replacement node to inherit that. If it is a fixed size however, we don't want the replacement node to inherit it.

I think the list is good, and this change addresses the issue. I am not sure about min/max width/height - that's the only bit that I could think of which isn't currently covered.

Why do we want a fixed size child to honor its fixed size? (I guess I keep going back to the use case of embedding maps/camera/whatever, where we want the replacement element to just get the size from the thing it's replacing).

iamralpht (approved):

timothyfroehlich (commented):
>  > All the fields in `ViewStyle` are final because that's how serdegen makes them. So in order to change one field you have to do something like this. I took this from @iamralpht's squoosh branch since he already needed it there.

Yeah, the Java objects that serdegen creates don't cover all of the use cases. It's good that they are immutable, otherwise I'm sure we'd have bugs where we're mutating something that other objects have a handle on, but it's annoying that there isn't a generated "clone with these edits". In Rust you can just clone an immutable object and store the clone as a mutable reference, then make edits, then turn the mutable ref into an immutable one before you share it with the rest of the program.

Hopefully we can move to protobuf serialization soon, which has more ergonomic Java codegen ;) 

I'd be willing to take a stab at that if you'd like, Serdegen has bugged me for a while. I might need some help with the Rust side but hopefully not too much.

timothyfroehlich:
> > Ok I have changed this to save all of the original node's external layout properties -- at least, my guess at what those are. It got a bit more complicated to support constraints because if certain constraints are set, the width/height gets set to something like Dimension::Percent(), in which case we want the replacement node to inherit that. If it is a fixed size however, we don't want the replacement node to inherit it.
> 
> I think the list is good, and this change addresses the issue. I am not sure about min/max width/height - that's the only bit that I could think of which isn't currently covered.
> 
> Why do we want a fixed size child to honor its fixed size? (I guess I keep going back to the use case of embedding maps/camera/whatever, where we want the replacement element to just get the size from the thing it's replacing).

I think it'd be worth mocking things up and figuring that out (do we have a design doc for ComponentReplacement anywhere?) I suspect the answer is that you would want it to stay fixed if it were some kind of image that you wouldn't want stretched or something? Or should the designer be using auto-layout in that situation anyways? 

@rylin8 I'm going to request changes, I'd really like some documentation added about what's going on with ComponentReplacement and the implications of using different Figma layouts. I'd really like to see some use cases written out with instructions for how to implement them in Figma and in the code. I think now is the best time to do this while we're thinking about a change here.

timothyfroehlich (changes requested):
@rylin8 I'm going to request changes, I'd really like some documentation added about what's going on with ComponentReplacement and the implications of using different Figma layouts. I'd really like to see some use cases written out with instructions for how to implement them in Figma and in the code. I think now is the best time to do this while we're thinking about a change here.

iamralpht:
> @rylin8 I'm going to request changes, I'd really like some documentation added about what's going on with ComponentReplacement and the implications of using different Figma layouts. I'd really like to see some use cases written out with instructions for how to implement them in Figma and in the code. I think now is the best time to do this while we're thinking about a change here.

Yeah, nice to clarify our documentation. We have basic documentation here: https://google.github.io/automotive-design-compose/docs/modifiers/modifiers#comp-replace

We can expand this with a basic list of how properties are combined. In a future PR, we could go to a couple of examples on what happens with different sets of properties (like parent is in a flow, parent is using constraints, child is a flow/autolayout, child is larger/smaller than the space).

designcompose/src/main/java/com/android/designcompose/Utils.kt
rylin8: Yea there's also the `a?.let { b } ?: c`

rylin8:
> @rylin8 I'm going to request changes, I'd really like some documentation added about what's going on with ComponentReplacement and the implications of using different Figma layouts. I'd really like to see some use cases written out with instructions for how to implement them in Figma and in the code. I think now is the best time to do this while we're thinking about a change here.

I'd prefer to documentation in another pull request. I'll be looking at removing the need to use `ParentLayout` separately, so if that works out the documentation will look different.

rylin8:
> Why do we want a fixed size child to honor its fixed size? (I guess I keep going back to the use case of embedding maps/camera/whatever, where we want the replacement element to just get the size from the thing it's replacing).

That's a good point, why do we want it to? At first I thought we'd want the replacement to keep its size because of some recent examples I'd seen. In media, maybe the custom action buttons want to be different sizes than the placeholders in the design. Now that I think about it though, maybe it's more often the case that we want the replacement node to take the size of the original one. Maybe we want this to be configurable though -- so that the developer can choose which size to take?

rylin8:
What if we default to use the original node size, but let the developer override that? Something like:
```
        replacementNode = { ComponentReplaceDoc.BlueNode(parentLayout = it.parentLayout.useReplacementSize()) },
```

timothyfroehlich:
> What if we default to use the original node size, but let the developer override that? Something like:
> 
> ```
>         replacementNode = { ComponentReplaceDoc.BlueNode(parentLayout = it.parentLayout.useReplacementSize()) },
> ```

Yeah, I think using the original's size by default is the right answer, because can't you also replace it with a non-Figma-Component Composable?

How would the above work if you're able to remove the need for `parentLayout`? Can you find a way to add a boolean argument?

timothyfroehlich:
> > @rylin8 I'm going to request changes, I'd really like some documentation added about what's going on with ComponentReplacement and the implications of using different Figma layouts. I'd really like to see some use cases written out with instructions for how to implement them in Figma and in the code. I think now is the best time to do this while we're thinking about a change here.
> 
> I'd prefer to documentation in another pull request. I'll be looking at removing the need to use `ParentLayout` separately, so if that works out the documentation will look different.

Only if you promise to have it in by the next release we make. Can you create a ticket for it?

rylin8:
> How would the above work if you're able to remove the need for `parentLayout`? Can you find a way to add a boolean argument?
Not sure, if it's not a common scenario I'd be ok with having to do something like this because 90% of the time you wouldn't need to do anything with `ParentLayout`, but I'll have to think about it more.

rylin8:
> Only if you promise to have it in by the next release we make. Can you create a ticket for it?
Yea one for `ParentLayout` https://github.com/google/automotive-design-compose/issues/843 and one for documentation https://github.com/google/automotive-design-compose/issues/844

rylin8:
Ok I updated to by default use the original node size and added a function on `ParentLayout` to override this behavior.

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -590,6 +590,138 @@ internal fun mergeStyles(base: ViewStyle, override: ViewStyle): ViewStyle {
...
+        align_self,
+        position_type,
+        transform,
+        relative_transform,
+    )
+}
+
+// Take the external layout data and merge it into this ViewStyle, overriding its values and
+// returning a new ViewStyle.
+internal fun ViewStyle.withExternalLayoutData(
```
iamralpht, line 693: It might make sense to make this a method on the builder instead of on `ViewStyle` directly (but that would make the call site a bit noisier). I don't think it's urgent.

iamralpht:
Discussed in 1:1, and said:

- Let's add a test for inserting a non-DC Composable
- Let's check with users to find out whether we need to offer the option to take the inserted node's size instead of using the slot bounds

rylin8:
Given recent discussions, I'm now leaning toward dropping support to take the replacement component's size. I can't think of a legitimate scenario where this is really needed anymore. Most of the time you want the original size, and when you don't there are ways to make it work without this support.

Dropping support would keep the code a bit cleaner and less confusing. In addition, looking at the change to drop `ParentLayout` propagation (https://github.com/google/automotive-design-compose/issues/843), it would be hard to get that working and support using the replacement component's size because doing so requires the developer to change the parent layout.

Thoughts?

iamralpht:
> Dropping support would keep the code a bit cleaner and less confusing. In addition, looking at the change to drop `ParentLayout` propagation (https://github.com/google/automotive-design-compose/issues/843), it would be hard to get that working and support using the replacement component's size because doing so requires the developer to change the parent layout.
> 
> Thoughts?

I'm happy with dropping -- good to keep it simpler until we're clear on the use case motivating complexity.

rylin8:
> > Dropping support would keep the code a bit cleaner and less confusing. In addition, looking at the change to drop `ParentLayout` propagation (#843), it would be hard to get that working and support using the replacement component's size because doing so requires the developer to change the parent layout.
> > Thoughts?
> 
> I'm happy with dropping -- good to keep it simpler until we're clear on the use case motivating complexity.

Updated.

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -590,6 +590,130 @@ internal fun mergeStyles(base: ViewStyle, override: ViewStyle): ViewStyle {
...
+    builder.cross_axis_item_spacing = cross_axis_item_spacing
+    builder.flex_grow = flex_grow
+    builder.flex_shrink = flex_shrink
+    builder.flex_basis = flex_basis
+    builder.bounding_box = bounding_box
+    builder.horizontal_sizing = horizontal_sizing
+    builder.vertical_sizing = vertical_sizing
+    builder.width = width
+    builder.height = height
+    builder.min_width = min_width
```
timothyfroehlich, line 656: Can you confirm that this is the correct place for this code? Does it make sense for it to be handled back in the Rust code and have the result be serialized into the .dcf file?

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -590,6 +590,130 @@ internal fun mergeStyles(base: ViewStyle, override: ViewStyle): ViewStyle {
...
+    builder.max_width = max_width
+    builder.max_height = max_height
+    builder.aspect_ratio = aspect_ratio
+    builder.pointer_events = pointer_events
+    builder.meter_data = meter_data
+    return builder
+}
+
+// Take the external layout fields of ViewStyle and put them into an ExternalLayoutData object.
+internal fun ViewStyle.externalLayoutData(): ExternalLayoutData {
```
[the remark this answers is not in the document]
iamralpht: We could, but I think it should be a bigger change and would be primarily around splitting out appearance stuff from layout stuff. I filed #869 to track it. I don't think it should block landing this change.

timothyfroehlich (approved):

designcompose/src/main/java/com/android/designcompose/Utils.kt
rylin8: I'll consider this in a follow up.

PR #831, opened 2024-02-29
Update to Android Gradle Plugin 8.3.0 stable

timothyfroehlich:
This corresponds with the release of Android Studio Iguana


rylin8:
Does this mean I have to install Android Studio again, without the update method?

timothyfroehlich:
Yes, though since this is to the stable version it'll be the last time for a while.

rylin8 (approved):

PR #849, opened 2024-03-05
Replace KTFmt Gradle Plugin with Spotless.

timothyfroehlich:
I've been having consistent problems with the ktfmt Gradle plugin and
this time it blocked me from updating to Gradle 8.6. I've been meaning
to replace it with Spotless for a while, which can be expanded to also
handle licenses, markdown, and other formatters. Spotless seems better
supported in general and has allowed me to remove most of the custom
configuration that I had to add to the ktfmt Gradle plugin to get it to
also format our build scripts.

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/InteractionState.kt
+++ b/designcompose/src/main/java/com/android/designcompose/InteractionState.kt
@@ -34,7 +34,7 @@ import com.android.designcompose.serdegen.View
 // to a specific component instance instead of all of them, such as changing to a different variant.
 private fun getInstanceIdWithKey(instanceId: String, key: String?): String {
     var varKey = instanceId
-     if (key != null) varKey += key
+    if (key != null) varKey += key
     return varKey
 }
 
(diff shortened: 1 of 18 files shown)


plugins/settings.gradle.kts
```diff
@@ -30,6 +30,7 @@ pluginManagement {
     includeBuild("../build-logic")
 }
 
+@Suppress("UnstableApiUsage")
```
rylin8, line 33: Why do you need this?

designcompose/src/main/java/com/android/designcompose/InteractionState.kt
```diff
@@ -34,7 +34,7 @@ import com.android.designcompose.serdegen.View
 // to a specific component instance instead of all of them, such as changing to a different variant.
 private fun getInstanceIdWithKey(instanceId: String, key: String?): String {
     var varKey = instanceId
-    if (key != null) varKey += key
+     if (key != null) varKey += key
```
rylin8, line 37: Accidental space

plugins/settings.gradle.kts
timothyfroehlich: The dependencyResolutionManagement block is still marked unstable, though it's been around since at least Gradle 7

designcompose/src/main/java/com/android/designcompose/InteractionState.kt
timothyfroehlich: That's... an indication that I didn't get something right, since spotless/ktfmt should be catching that

rylin8 (approved):

PR #856, opened 2024-03-07
Closes #855: Add support to parse baseline alignment, diamond stroke cap

### LINKED ISSUE #855: Baseline alignment, Diamond stroke cap causing serialization errors
Figma has been continuously adding new features. Some of these features bring new types into the API that we don't yet support, causing serialization errors. Fix a some of these that we have seen in recent designs so that serialization doesn't immediately fail and we can continue to use the Figma file.

rylin8:
This avoid serialization errors.

### DIFF
--- a/crates/figma_import/src/transform_flexbox.rs
+++ b/crates/figma_import/src/transform_flexbox.rs
@@ -143,12 +143,14 @@ fn compute_layout(node: &Node, parent: Option<&Node>) -> ViewStyle {
             LayoutAlignItems::Max => AlignItems::FlexEnd,
             LayoutAlignItems::Min => AlignItems::FlexStart,
             LayoutAlignItems::SpaceBetween => AlignItems::FlexStart, // XXX
+            LayoutAlignItems::Baseline => AlignItems::FlexStart, // Not supported, default to FlexStart
         };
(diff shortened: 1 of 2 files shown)


iamralpht (approved):

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -143,12 +143,14 @@ fn compute_layout(node: &Node, parent: Option<&Node>) -> ViewStyle {
             LayoutAlignItems::Max => AlignItems::FlexEnd,
             LayoutAlignItems::Min => AlignItems::FlexStart,
             LayoutAlignItems::SpaceBetween => AlignItems::FlexStart, // XXX
+            LayoutAlignItems::Baseline => AlignItems::FlexStart, // Not supported, default to FlexStart
```
timothyfroehlich, line 146: I really wish we had some way to throw a non-crashing error here to inform the user that they're using an unsupported Figma feature and that it's silently being replaced with a different behavior.

rylin8: We can print an error (though most people wouldn't see it), or think about a way to make it more clear. For now though, I'm just trying to get something quick in.

PR #859, opened 2024-03-10
squoosh: introduce an (optional, disabled) tree management strategy.

iamralpht:
Squoosh introduces a new approach for managing the view tree to DesignCompose. With the introduction of f/rustlayout, we actually don't need to make a Composable for every node in the design asset, and can instead only emit Composables when we need interactions or accessibility.

Because we consolidate management of the tree of design elements, it's a bit easier to perform layout animations (as we can simply have two trees, apply the relevant state to both, do layout on both, and then create a combined tree that allows transitioning). When Squoosh is enabled some basic "Smart Animate" functionality works.


designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt
```diff
@@ -114,7 +114,7 @@ private interface DesignSwitcher {
 
     @Composable
     fun CustomComponent(
-        modifier: Modifier = Modifier,
+        modifier: Modifier, // = Modifier,
```
iamralpht, line 117: Not sure why I needed this -- Compose compiler was complaining about Composables not being allowed to have defaults -- perhaps I need to upgrade Android Studio? (This is in main and CI is happy).

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -56,11 +56,36 @@ import java.util.Optional
 import kotlin.math.roundToInt
 
 internal data class TextMeasureData(
+    val textHash: Int,
     val paragraph: ParagraphIntrinsics,
     val density: Density,
     val maxLines: Int,
     val autoWidth: Boolean,
-)
+) {
```
iamralpht, line 64: This stuff is all optimization; hashing TextMeasureData was number 1 in the profile for a while.

designcompose/src/main/java/com/android/designcompose/squoosh/DESIGN.md
```diff
@@ -0,0 +1,92 @@
...
+rendered. All of the child Composables have a Modifier that customizes drawing to completely skip
+drawing unless the id in the `SquooshRenderSelector` matches. Yuck! But good enough...
+
+## Animated transitions and enum-driven variant changes
+Squoosh implements animated transitions, and can also animate variant changes caused by enum values
+changing.
+
+This is described above in the "typical flow". When there are animations to be run, they are always
+between states of some kind. We generate two trees, merge parts of them that need to be animated
+between, and then do
```
iamralpht, line 92: 🤔

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -0,0 +1,395 @@
...
+private fun needsStyleTween(a: ViewStyle, b: ViewStyle): Boolean {
+    // Compare some style things and decide if we need to tween the styles.
+    if (a.background != b.background) return true
+    if (a.stroke != b.stroke) return true
+    return false
+}
+
+// XXX: Horrible code to deal with our terrible generated types. Maybe if style moves to proto then
+//      we'll get some more egonomic generated classes.
+internal fun ViewStyle.asBuilder(): ViewStyle.Builder {
```
iamralpht, line 320: @rylin8 I think we have another version of this function somewhere else now, too? I should probably delete this version...

rylin8: Yes I copied yours in my PR which is not merged yet -- https://github.com/google/automotive-design-compose/pull/826

crates/dc_jni/src/jni.rs
```diff
@@ -234,6 +234,11 @@ fn jni_add_nodes<'local>(
                             );
                         }
                     }
+                    for LayoutParentChildren { parent_layout_id, child_layout_ids } in
+                        &node_list.parent_children
+                    {
+                        manager.update_children(*parent_layout_id, child_layout_ids)
```
rylin8, line 240: Can you remind me why this change is needed? Was this to avoid a problem when views were added out of order?

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1338,8 +1338,8 @@ fn visit_node(
             path: fill_paths,
             stroke: stroke_paths,
             stroke_cap: node.stroke_cap.clone(),
-            start_angle_degrees: arc_data.starting_angle,
-            sweep_angle_degrees: arc_data.ending_angle,
+            start_angle_degrees: euclid::Angle::radians(arc_data.starting_angle).to_degrees(),
+            sweep_angle_degrees: euclid::Angle::radians(arc_data.ending_angle).to_degrees(),
```
rylin8, line 1342: What does this fix?

designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt
rylin8: I've seen this when doing inheritance, but yea don't know why this is needed.

crates/dc_jni/src/jni.rs
iamralpht: Yeah, it does a pass to "update the native layout tree" once it's built its full tree in Kotlin, and I couldn't think of an efficient way to visit the subset of the tree that had changed (so I could use the old index-based child update methods) in the right order. So instead, I made it so that the parent would just specify all of it's child indices whenever there's a change, and then the algorithm to figure out how to update only the invalidated layout nodes became simple (it just visits the invalidated nodes in tree order).

This way the caching that the Rust layout engine does gets used, so the rust layout iterations are faster.

crates/figma_import/src/transform_flexbox.rs
iamralpht: I think these values are in radians from the plugin. I noticed this when I tried animating arcs; dials/gauges still works (so either I made a similar change there, or it wasn't needed).

iamralpht:
Ok, looks like a legit bug in the 1px separator tests! I'll have to figure out how to exclude the Compositing Views one (which uses ExoPlayer to make an external surface) from the test.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -952,6 +953,13 @@ fun DesignDoc(
     endSection()
 }
 
+// Enable the "Squoosh" refactor. Squoosh implements its own view tree, rather than using Compose,
+// which brings some performance and flexibility benefits. Squoosh isn't feature complete (no
+// scrolling, no lists, no transformed input), but it does add animations and is likely the
+// direction that DesignCompose will move in to be lighter weight and better integrate with
+// external layout.
+private const val USE_SQUOOSH = false
```
timothyfroehlich, line 961: Since this is hard coded it means that we can't run tests in CI, correct?

iamralpht: Yeah, I should have written this in the PR, but I thought we could:

1. Land the change so that we get build coverage, and don't need to rebase again.
2. Add the ability to turn it on dynamically, enable test cases where it's working...
3. Iterate on fixing the implementation and enabling more test cases.

timothyfroehlich (commented):
I don't know where to start on this, there's almost 4000 lines of code here. I haven't budgeted the time to provide a meaningful review of something like this.

designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt
timothyfroehlich: I may have found your answer, though I don't know why you only just started running into it: https://b.corp.google.com/issues/322121224

Compose Compiler 1.5.8 actually blocks the use of default values for open Composables entirely since they don't have support for it: https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.8

rylin8: But there's no inheritance going on, and all the generated functions from annotations look like this, with the first parameter being `modifier: Modifier = Modifier`.

crates/figma_import/src/toolkit_schema.rs
```diff
@@ -122,6 +124,9 @@ pub enum RenderMethod {
 /// Represents a toolkit View (like a Composable).
 #[derive(Clone, PartialEq, Debug, Serialize, Deserialize)]
 pub struct View {
+    // unique numeric id (not replicated within one DesignCompose file), which is used by the
+    // renderer when building a layout tree, or when mapping from an integer to a View.
+    pub unique_id: u16,
```
rylin8, line 129: I think this warrants a serialized document header version bump.

designcompose/src/main/java/com/android/designcompose/squoosh/DESIGN.md
rylin8: Missing a work at the end of the last sentence?

designcompose/src/main/java/com/android/designcompose/squoosh/DESIGN.md
```diff
@@ -0,0 +1,92 @@
...
+   with the updated context.
+    * This gives us a `SquooshResolvedNode` tree of the world after all transitions have ended.
+    * We use `newSquooshAnimate` to create `SquooshAnimationControl` objects that control progress
+      for each transition. `newSquooshAnimate` identifies similar nodes and different nodes and
+      creates objects that will animate the appropriate property for each. Some nodes from the source
+      tree get moved to the destination tree so that they can be faded out.
+6. If there are transitions, we launch an effect to run an animation clock and update the state of
+   each transition's animation. This is how we can seemingly run many different animations at once
+   -- they all just show different positions between the two trees, and can't be nested.
+7. We emit a `Layout` Composable that occupies the appropriate space and has a `squooshRender`
```
rylin8, line 68: What do you mean by emitting a `Layout` composable? Is this related to the custom layout function?

designcompose/src/main/java/com/android/designcompose/squoosh/DESIGN.md
```diff
@@ -0,0 +1,92 @@
...
+DesignCompose tree that needs input).
+
+Inside of the `Layout` Composable in `SquooshRoot`, we nest any Child Composables from our list.
+
+Painting just one child Composable is tricky. Compose only lets us paint *all* child Composables in
+one go. So we have a horrible hack where a boxed object is shared between the child Composables and
+the `squooshRender` Modifier. When `squooshRender` wants to paint a particular child Composable it
+updates the boxed object (a `SquooshRenderSelector`) to have the id of the node that should be
+rendered. All of the child Composables have a Modifier that customizes drawing to completely skip
+drawing unless the id in the `SquooshRenderSelector` matches. Yuck! But good enough...
```
rylin8, line 84: I don't really follow this. In what scenarios do you paint just one child composable?

crates/figma_import/src/toolkit_schema.rs
iamralpht: Oops, that's right, I'll update the header version.

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -56,11 +56,36 @@ import java.util.Optional
...
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as TextMeasureData
+
+        if (textHash != other.textHash) return false
+        if (density != other.density) return false
+        if (maxLines != other.maxLines) return false
+        if (autoWidth != other.autoWidth) return false
```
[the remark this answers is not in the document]
iamralpht: Good catch! Will fix.

designcompose/src/main/java/com/android/designcompose/squoosh/DESIGN.md
iamralpht: I meant the basic [Compose Layout composable function](https://developer.android.com/reference/kotlin/androidx/compose/ui/layout/package-summary#Layout(kotlin.collections.List,androidx.compose.ui.Modifier,androidx.compose.ui.layout.MultiContentMeasurePolicy)). Currently Squoosh makes one of those that does drawing (via the `squoohRender` Modifier), and then some other child ones that handle input events and wrap nested Composables (via replacement, etc).

I think that Squoosh should probably be changed so that instead of the root Layout doing all rendering itself, the children also do some rendering. The disadvantage of moving to this approach is that it's harder to support multiple masks and things like that over multiple children.

designcompose/src/main/java/com/android/designcompose/squoosh/DESIGN.md
iamralpht: We can only paint all children at once (via `drawContent`), but `squooshRender` wants to render all of the content in one go -- including the stuff before an external child, between two external children, and after the last external child.

I added this thing as a mechanism so that `squooshRender` could call `drawContent` any time it got to a node with an external Composable, but only one of the external Composables would actually render. So as `squooshRender` iterates the tree it will call `drawContent` N times, but each individual external Composable will only get painted once.

I'd really like to be able to have a `drawChild(idx)` method in the `drawWithContent` modifier, but I couldn't figure out how to achieve this, because Compose works quite hard to prevent any modifier from enumerating its children. Perhaps there's a way to do this with the `Modifier.Node` API...

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -477,7 +479,304 @@ internal fun ContentDrawScope.render(
         drawContent()
     }
 
-    if (useBlendMode) {
+    if (useBlendMode || opacity < 1.0f) {
```
rylin8, line 482: Why is the `opacity < 1.0f` added here? If this is added, doesn't a `canvas.save()` need to be called somewhere earlier also?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -0,0 +1,703 @@
...
+    }
+    val switchDocId: (String) -> Unit = { newDocId: String ->
+        run { DocumentSwitcher.switch(originalDocId, newDocId) }
+    }
+
+    val designSwitcher: @Composable () -> Unit = {
+        if (showDesignSwitcher) {
+            val branchHash = DocServer.branches(docId)
+            Box(Modifier.fillMaxSize(), contentAlignment = Alignment.TopEnd) {
+                // DesignSwitcher(doc, docId, branchHash, switchDocId)
```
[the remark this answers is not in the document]
iamralpht: I think it mostly works; basic interactions and overlays are supported, so if you uncomment this line then it should come up.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
```diff
@@ -0,0 +1,288 @@
...
+    animationValues: State<Map<Int, Float>>
+): Modifier =
+    this.then(
+        Modifier.drawWithContent {
+            val animValues = animationValues.value
+            for ((id, transition) in animations) {
+                val animationOffset = animValues[id]
+                if (animationOffset == null) {
+                    // This happens the first time through, because we manage to render before our
+                    // LaunchedEffect has run. It doesn't seem OK to drop a frame at the start of
```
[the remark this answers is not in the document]
iamralpht: Oh, I will update this comment. Initially I was using `LaunchedEffect` to manage the coroutine that updates the animation values, but then at some point I needed more control and switched to using `InteractionScope.launch` (SquooshRoot.kt ~L550) instead. That method had some flags to control if the coroutine would be executed immediately or lazily, and that eliminated the one frame latency for starting new animations.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -477,7 +479,304 @@ internal fun ContentDrawScope.render(
...
+        drawContext.canvas.restore()
+    }
+    drawContext.canvas.restore()
+}
+
+internal fun squooshShapeRender(
```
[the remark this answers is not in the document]
iamralpht: It's really similar to the existing logic in that file, eventually we should only have one version of it...

designcompose/src/main/java/com/android/designcompose/Layout.kt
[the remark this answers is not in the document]
iamralpht: re-reading, hashCode does take density into account, but I'll reorder the equals comparisons so they're in the same order as the hashCode calculation for easier reading.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
iamralpht: Your PR has merged, but I haven't updated to use that implementation yet.

crates/figma_import/src/toolkit_schema.rs
iamralpht: Introducing DCF v17...

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
iamralpht: I removed this, I applied it in error -- we handle `opacity < 1.0f` via `graphicsLayer` in non-squoosh, so it was a mistake to add this condition (and it was indeed causing the 1px separator test to crash).

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt
```diff
@@ -82,6 +82,10 @@ val EXAMPLES: ArrayList<Triple<String, @Composable () -> Unit, String?>> =
         Triple("Variable Borders", { VariableBorderTest() }, VariableBorderTestDoc.javaClass.name),
         Triple("Layout Tests", { LayoutTests() }, LayoutTestsDoc.javaClass.name),
         Triple("Custom Brush", { CustomBrushTest() }, CustomBrushTestDoc.javaClass.name),
+        Triple("SA", { SmartAnimateTest() }, SmartAnimateTestDoc.javaClass.name),
+        Triple("SA: Variant", { VariantAnimationTest() }, VariantAnimationTestDoc.javaClass.name),
```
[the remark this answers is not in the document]
iamralpht: Yep, removed.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/CompositingViewsTest.kt
```diff
@@ -0,0 +1,90 @@
...
+interface CompositingViewsTest {
+    @DesignComponent(node = "#MainFrame")
+    fun MainFrame(
+        @Design(node = "#Child1") child1: @Composable (ComponentReplacementContext) -> Unit,
+        @Design(node = "#Child2") child2: @Composable (ComponentReplacementContext) -> Unit,
+    )
+}
+
+@Composable
+fun CompositingViewsTest() {
```
iamralpht, line 48: This test needs to be skipped for Roborazzi, because it can't run ExoPlayer.

iamralpht: Excluded manually for now.

designcompose/src/main/java/com/android/designcompose/squoosh/DESIGN.md
iamralpht: I improved the doc in this PR, but I also put it in [Docs](https://docs.google.com/document/d/1tCNgkqCUrb-D1w13nBueeEmSB-ElroriLTyribkuDiE/edit?tab=t.0) so we can iterate on it a bit more easily.

designcompose/src/main/java/com/android/designcompose/squoosh/DESIGN.md
iamralpht: Cleaned this up.

designcompose/src/main/java/com/android/designcompose/squoosh/DESIGN.md
iamralpht: I updated the doc, it's a bit clearer now, and we can do more iteration on it in Docs.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
```diff
@@ -0,0 +1,288 @@
...
+    computedLayout: Layout,
+    customizations: CustomizationContext,
+    nodeName: String,
+) {
+    val paragraph =
+        Paragraph(
+            paragraphIntrinsics = textInfo.paragraph,
+            width = computedLayout.width * density.density,
+            maxLines = textInfo.maxLines,
+            ellipsis = style.text_overflow is TextOverflow.Ellipsis
```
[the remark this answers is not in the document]
iamralpht: I've got a local fix for this (it looks at whether a line exceeds the layout allocated bounds after layout is complete and computes a new `maxLines`) but I implemented it in the wrong place (in rendering, rather than in a post-layout pass) and I'd rather just land Squoosh as-is now and then fix bugs.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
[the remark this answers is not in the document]
iamralpht: I tried uncommenting, and it works (except for the stuff using content replacement, which squoosh doesn't do yet, and needs some design work). I'll check in an uncommented variation once we've got this change merged, tho.

rylin8:
There is a conflict with the component replace serialized file since I merged a change to fix the crash. Can I restore the nodes that I removed from the Figma document for that test now, since I merged? Then you should be able to take my serialized file (or redownload it), along with the screenshot for it, to resolve the conflict.

iamralpht:
> There is a conflict with the component replace serialized file since I merged a change to fix the crash. Can I restore the nodes that I removed from the Figma document for that test now, since I merged? Then you should be able to take my serialized file (or redownload it), along with the screenshot for it, to resolve the conflict.

Yes, please go ahead.

rylin8:
> Yes, please go ahead.

Done

iamralpht:
> > Yes, please go ahead.
> 
> Done

Ok, I've rebased and updated the serialized file and the screenshot.

timothyfroehlich (commented):
+1, The non squoosh changes look fine, but I can't review the squoosh changes themselves.

integration-tests/validation/build.gradle.kts
```diff
@@ -111,6 +111,9 @@ dependencies {
     implementation(libs.androidx.compose.material)
     implementation(libs.androidx.compose.ui.tooling.preview)
     implementation(libs.material)
+    implementation("androidx.media3:media3-exoplayer:1.1.1")
```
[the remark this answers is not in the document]
iamralpht: I actually am not sure how to do this - would you be able to do it? (otherwise I can have a look tomorrow evening).

designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt
iamralpht: I deleted the commented code, since it's an internal implementation detail and not an externally exposed API.

rylin8 (approved):

integration-tests/validation/build.gradle.kts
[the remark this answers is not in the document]
iamralpht: Captured in #890

PR #863, opened 2024-03-12
Parentlayout WIP

rylin8:


designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
```diff
@@ -89,24 +89,23 @@ internal fun DesignFrame(
...
-                    content()
-                }
-
-                override val textStyle: TextStyle? = null
+        val replacementParentLayout =
+            parentLayout?.withReplacementLayoutData(viewStyle.externalLayoutData())
+        DesignParentLayout(replacementParentLayout) {
+            replacementComponent(
+                object : ComponentReplacementContext {
+                    override val layoutModifier = layoutInfo.selfModifier
```
iamralpht, line 97: Do we need a ticket to look at what's in this Modifier post Rust-layout? (I'm not sure what we have in here anymore).

rylin8: I started looking into this crash more https://github.com/google/automotive-design-compose/issues/864 and I'm in this area of the code so I was going to take a pass at refactoring/removing bits of this that aren't needed anymore.

iamralpht (approved):

designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
iamralpht: sg, thanks Rob.

Issue #860, opened 2024-03-11
Cannot upgrade to Compose Compiler 1.5.8+


timothyfroehlich:
[Compose Compiler 1.5.8](https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.8) adds the following:

>Add a compilation error for default parameters in open functions (previously only it was restricted only for abstract functions). Compose does not support substitution of default parameters from overridden functions at the moment, and overriding a composable function with default parameters was failing at runtime. Use @Suppress("ABSTRACT_COMPOSABLE_DEFAULT_PARAMETER_VALUE") to suppress, but note that overriding such function without providing a default value for parameter will result in a crash at runtime.([b/317490247](https://issuetracker.corp.google.com/issues/317490247))

timothyfroehlich:
```
> Task :designcompose:compileDebugKotlin FAILED
e: file:///data/DesignCompose/primary/designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt:117:30 Overridable Composable functions with default values are not currently supported
e: file:///data/DesignCompose/primary/designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt:147:30 Overridable Composable functions with default values are not currently supported
```

timothyfroehlich:
The result of this is that users of DesignCompose will have blocking build failures if they attempt to update to the latest Compose Compiler / Kotlin

PR #870, opened 2024-03-13
Update to Kotlin 1.9.23, Compose Compiler 1.5.11, KSP 1.9.23-1.0.20

timothyfroehlich:


rylin8:
The comment that this is blocked by https://github.com/google/automotive-design-compose/issues/860 no longer applies right? And is that issue still an issue?

timothyfroehlich:
> The comment that this is blocked by #860 no longer applies right? And is that issue still an issue?

Yep, I updated the title but not the description.

rylin8 (approved):

PR #873, opened 2024-03-15
Closes #864: Fix possible crash when using component replacement with a non-DesignCompose generated function

### LINKED ISSUE #864: Component replacement using a non-DesignCompose generated function can cause a crash
When doing a component replacement, DesignCompose passes in a `ParentLayoutInfo` object that contains the child index of the replacement component with respect to its parent. However if the replacement component is not a DesignCompose generated function, it will never subscribe to layout with that child index, causing the siblings of the original node being replaced to have a gap in their child indices.

For example, if node A has children B, C, and D, their indices might be 0, 1, and 2 respectively. If C has a component replacement customization and is replaced by a non-DesignCompose composable, then B will subscribe to layout with index 0, then D will subscribe with index 2. The Rust layout system currently doesn't support this and will try to insert a child at index 2, causing an assertion.

rylin8:
When using a component replacement customization with a non-DesignCompose composable, the replacement composable would not subscribe to layout changes. If the original component had any siblings, those siblings would subscribe to layout with a missing child index, causing a crash in Rust layout when trying to insert a style node at an invalid position. For example if nodes A, B, and C were siblings and B was the replacement node, A and C would subscribe to layout with indicies 0 and 2 respectively, and Rust would fail trying to insert a style node at index 2 when the list was only had a lenght of 1.

This change fixes the crash by detecting when a component replacement does not call a DesignCompose composable, and in that scenario will subscribe to layout anyway with the original node's style.

Secondly, the Modifiers passed into ComponentReplacementContext have been changed. The appearance modifier has been removed, and the layout modifier has been changed to be a modifier that holds layout data for the original node. When using DesignCompose composables as replacement content, this modifier is not needed, but when using non-DesignCompose composables, the layout modifier should be passed into the composable in order to use the original node's size and position.


designcompose/src/main/java/com/android/designcompose/CustomizationContext.kt
```diff
@@ -254,16 +254,13 @@ fun CustomizationContext.setKey(key: String?) {
...
-    // This modifier doesn't contain any appearance information, and should be the first
-    // in the modifier chain.
+    // Return the custom layout modifier that this component would have used so that the layout
+    // function can retrieve the component's layout properties. When replacing a node with a
+    // composable that is not a DesignCompose generated function, such as a simple Box() or an
+    // AndroidView, this modifier should be used as a modifier for that component in order for it to
+    // retain the original node's layout (size, position) properties.
     val layoutModifier: Modifier
 
-    // Return the appearance modifier. This causes the view to be presented as specified
```
iamralpht, line 262: hooray! (I saw this is also gone in the docs)

designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
```diff
@@ -84,6 +84,60 @@ internal fun DesignFrame(
         m = m.then(customModifier)
     }
 
+    // Keep track of the layout state, which changes whenever this view's layout changes
+    val (layoutState, setLayoutState) = remember { mutableStateOf(0) }
+    var rootLayoutId = parentLayout?.rootLayoutId ?: -1
+    if (rootLayoutId == -1) rootLayoutId = layoutId
+
+    val subscribeLayout =
```
iamralpht, line 92: I like that this logic is getting better encapsulated, tho I think we need to revisit the implementation soon (squoosh is one alternative since only the root runs layout, I was also looking at the `Modifier.Node` API which allows traversal of children -- that might give a way to gather all of the child layout properties after composition but before layout, but I'm not sure...).

iamralpht (approved):

designcompose/src/main/java/com/android/designcompose/DesignFrame.kt
rylin8: Interesting I haven't seen `Modiifer.Node`, worth checking out.

PR #874, opened 2024-03-15
Closes #844: Add documentation for component replacement

### LINKED ISSUE #844: Add documentation for component replacement
Replacing nodes is a necessary and useful feature of DesignCompose, but it's a bit confusing. Add additional documentation with examples on how to do this, including descriptions of what features the replacement composable takes from the original node vs it's own, and what context is passed in when invoking your replacement composable.

rylin8:
Explain what is in the updated ComponentReplacementContext class and how to use with non-DesignCompose replacement composables.


rylin8:
Here's what the updated section looks like
Screenshot 2024-03-15 at 9 56 31 AM

docs/_docs/modifiers/modifiers.md
```diff
@@ -270,42 +270,60 @@ interface HelloWorld {
...
 
-The example below passes a `Composable` `fun` that accepts
-`software.vehicle.design.ComponentReplacementContext`. This object provides
-methods to access the style of the replaced element and its descendants.
-Component Replacements are provided with information on the component they are
-replacing including the modifiers that would be applied for the layout style,
-visual style, and (if applicable) the text style.
-
-Doing so allows a replacement component to take on the designer specified
+The `software.vehicle.design.ComponentReplacementContext` object provides
```
iamralpht, line 278: Is it still called `software.vehicle`? I thought it was `com.android.something` now...

rylin8: Whoops, good catch will fix.

rylin8: Done

docs/_docs/modifiers/modifiers.md
```diff
@@ -270,42 +270,60 @@ interface HelloWorld {
...
     val layoutModifier: Modifier
 
-    /// Return the appearance modifier. This causes the view to be presented as specified
-    /// by the designer. It also includes any modifier customizations that may have been
-    /// specified elsewhere in the program.
-    val appearanceModifier: Modifier
-
-    /// Render the content of this replaced component, if any.
+    // Render the content of this replaced component, if any.
     @Composable fun Content(): Unit
```
timothyfroehlich, line 296: What does someone do with the `Content()` Composable? Where, how and why would they use it in their code?

timothyfroehlich (changes requested):
I think Component replacement deserves more of a write-up with a more clear use case and images of the Figma File and what the rendered result would be, but we can get away with what we have here. 

Generally, the documentation has gotten garbled in all of the transitions from when we originally wrote it. We need a bit more of a rewrite, and as I mentioned in the comments, it should start more simply, without mentioning the ComponentReplacementContext. Describe the CRC and why you'd need it after the simple example.

timothyfroehlich:
@Gurupreet Could you take a look at this as well?

docs/_docs/modifiers/modifiers.md
rylin8: Ooh good point. I forgot about what `Content()` does. It actually gives the developer a way to compose the children of the original node, but it currently doesn't work. I forgot about this completely. @iamralpht what is the use case for this?

Since it doesn't work I'm wondering if we need to support it at this time. I did a quick test and I believe it doesn't work because of the way I fixed the component replacement crash by subscribing to layout after the fact. In doing so, if the replacement composable also called `Content()`, then `Content()` would subscribe to layout before its parent, and the Rust layout manager wouldn't be able to find the parent. It should probably error but right now it just think that it's a root node and adds it to a list of root nodes, but won't be used.

iamralpht: Let's remove it. It used to be "I want to add interactivity to something by interposing an interactive view which takes the original children", but with Compose you can just use a Modifier for this use case (and we do still have the Modifier customization). So I don't think that we need this.

docs/_docs/modifiers/modifiers.md
```diff
@@ -270,42 +270,60 @@ interface HelloWorld {
...
 
-The example below passes a `Composable` `fun` that accepts
-`software.vehicle.design.ComponentReplacementContext`. This object provides
-methods to access the style of the replaced element and its descendants.
-Component Replacements are provided with information on the component they are
-replacing including the modifiers that would be applied for the layout style,
-visual style, and (if applicable) the text style.
-
-Doing so allows a replacement component to take on the designer specified
+The `com.android.designcompose.ComponentReplacementContext` object provides
```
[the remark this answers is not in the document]
rylin8: done

docs/_docs/modifiers/modifiers.md
```diff
@@ -270,42 +270,60 @@ interface HelloWorld {
...
+When replacing a node with another DesignCompose generated composable, the `layoutModifier` from the
+`ComponentReplacementContext` can be ignored. DesignCompose will automatically use the original
+node's layout properties, but compose the replacement node instead. The example below describes how
+to do the replacement:
+
+```kotlin
+HelloWorldDoc.MainFrame(
+    button = { HelloWorldDoc.MyButton() }
+)
+```
```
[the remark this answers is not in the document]
rylin8: done

docs/_docs/modifiers/modifiers.md
```diff
@@ -270,42 +270,60 @@ interface HelloWorld {
...
+
+When replacing a node with a composable that was not generated by DesignCompose, that composable
+does not call into any DesignCompose code, and thus has no knowledge of the original node's layout.
+In order to use the original node's size and position, the replacement composable should use the
+`layoutModifier` from `ComponentReplacementContext`. The example below describes how to do this
+with a `Box` as the replacement composable:
+
+```kotlin
+HelloWorldDoc.MainFrame(
+    button = { Box(Modifier.background(Color.Green).then(it.layoutModifier)) }
```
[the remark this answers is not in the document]
rylin8: done

docs/_docs/modifiers/modifiers.md
```diff
@@ -270,42 +270,60 @@ interface HelloWorld {
     fun MainFrame(
         @Design(node = "#button") button: @Composable (ComponentReplacementContext) -> Unit,
```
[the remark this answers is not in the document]
rylin8: done

docs/_docs/modifiers/modifiers.md
rylin8: done

docs/_docs/modifiers/modifiers.md
```diff
@@ -270,42 +270,60 @@ interface HelloWorld {
     fun MainFrame(
```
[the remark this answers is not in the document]
rylin8: done

iamralpht (approved):

docs/_docs/modifiers/modifiers.md
```diff
@@ -256,56 +256,86 @@ VariantTestDoc.Square(
 
 ### Component replacement {#comp-replace}
 
-With a [`Composable`][2] of your choice, an element of a design can be replaced
-while retaining the style information provided by the designer. The difference
-between replacing versus adding descendants (as shown in [Creating lists][5] is
-that a replaced Figma design element isn't rendered.
+An node in a design can be replaced with a [`Composable`][2] of your choice while
```
timothyfroehlich, line 259: s/An/A

docs/_docs/modifiers/modifiers.md
```diff
@@ -256,56 +256,86 @@ VariantTestDoc.Square(
...
+retaining the style information provided by the designer. The difference between replacing
```
timothyfroehlich, line 260: It's an odd sentence, I feel like it would confuse more than help.

timothyfroehlich (approved):

docs/_docs/modifiers/modifiers.md
rylin8: Done

docs/_docs/modifiers/modifiers.md
rylin8: It explains the difference between replacing a node and replacing the children of a node. It's not the best sentence I agree but I don't want to hold up the release for this. Feel free to improve on it after the RC is made.

PR #875, opened 2024-03-18
Linux Only: Add method for reading JNI library for host unit tests.

timothyfroehlich:
This allows unit tests running with Roborazzi to use the layout JNI library and successfully run.

When compiling DesignCompose from source the JNI libraries will be compiled for your host system and unit tests will work. 

The published test libraries are only compatible with Linux systems.


timothyfroehlich:
@iamralpht How would you prefer I handle the Mac and Windows support? I could try to add something to the DesignCompose plugin to disable unit tests. I could update the [known issues page in our documentation](https://google.github.io/automotive-design-compose/docs/known-issues), but I'm hesitant to update it without confirming the status of the other issues listed there.

I would also like your recommendation on what to do about the Mac and Windows reference apps checks, which are now failing. They're marked as required so it blocks the merge.

iamralpht:
> @iamralpht How would you prefer I handle the Mac and Windows support? I could try to add something to the DesignCompose plugin to disable unit tests. I could update the [known issues page in our documentation](https://google.github.io/automotive-design-compose/docs/known-issues), but I'm hesitant to update it without confirming the status of the other issues listed there.
> 
> I would also like your recommendation on what to do about the Mac and Windows reference apps checks, which are now failing. They're marked as required so it blocks the merge.

Let's disable the tests for Mac and Windows right now, and file a ticket to fix these tests on these targets in the future. Linux is where most of our internal & partner developers are (and Win/Mac can still use DC prebuilts through Maven and test on device/emulator).

iamralpht (approved):
Cargo.lock changes also look OK, but not sure if you meant to include them.

PR #906, opened 2024-03-30
Closes #916: Support customization flattening into modules

### LINKED ISSUE #916: Support customization flattening into modules
Support grouping of common customizations into objects that can be used in other DesignCompose customizations instead of having to redeclare all of the same annotations. This would mean that common customizations go into a generated class that can be reused instead of an interface.

https://docs.google.com/document/d/1RilypCuxR7-UqHGrqjskTF8z4rcqI-vRDbUAurzcBJ4/edit?resourcekey=0-J0APk_JV2wDZ5n8kcAcY7A&tab=t.0#heading=h.ug6pvzkp8f66

rylin8:
Support a new @DesignModuleClass annotation to declare a class that holds customizations. These classes can be used to consolidate customizations into a single parameter in a @DesignComponent function, as well as nested in another @DesignModuleClass.

When declaring a @DesignModuleClass, use the @DesignProperty and @DesignVariantProperty annotations instead of the normal @Design and @DesignVariant properties. This is needed because of a bug in KSP that prevents annotations from being used with multiple targets. Once fixed, we can remove @DesignProperty and @DesignVariantProperty.

A @DesignModuleClass parameter can be specified in a @DesignComponent function with the @DesignModule annotation. A @DesignModuleClass class property can be specified in a @DesignModuleClass class with the @DesignModuleProperty annotation. Once the KSP bug is fixed, we can remove the @DesignModuleProperty annotation.

This change creates a new source file BuildModuleClass.kt to process @DesignModuleClass classes. Each @DesignModuleClass will generate a new file associated with that class that has an extension function for the class. The extension function creates a CustomizationContext object, populates customizations for it, and then returns it. Node queries and ignored images are also tracked so that any @DesignComponent function that uses modules will generate the appropriate queries() and ignoredImages() functions.

This change also cleans up some of the code in BuildProcessor.kt and moves common functionality into a separate file.


annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
```diff
@@ -50,15 +50,39 @@ annotation class DesignComponent(
  *
  * @param node the name of the Figma node
  */
+// TODO adding AnnotationTarget.PROPERTY to the list in @Target doesn't work. The annotation will
```
iamralpht, line 53: (Oops, I commented in the wrong place!). Please add the link to the KSP ticket in these comments.

codegen/src/main/kotlin/com/android/designcompose/codegen/BuildModuleClass.kt
```diff
@@ -0,0 +1,237 @@
...
+    private fun addModuleNodeQueries(className: String, allQueries: HashSet<String>) {
+        // Add queries for the specified class
+        val classQueries = queries[className]
+        classQueries?.let { allQueries.addAll(it) }
+        // Recurse on any modules that className references
+        nestedModules[className]?.forEach { addModuleNodeQueries(it, allQueries) }
+    }
+
+    // Return a set of ignored image node names for the module named className. This includes any
+    // n ode names from nested modules as well.
```
iamralpht, line 58: "n ode" -> "node"

iamralpht:
Looks good so far, only minor feedback from reviewing.

iamralpht:
Oh, plus need documentation, of course ;).

codegen/src/main/kotlin/com/android/designcompose/codegen/BuildModuleClass.kt
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    // Add node queries associated with className, then recurse on any classes that className
+    // references in its properties.
+    private fun addModuleNodeQueries(className: String, allQueries: HashSet<String>) {
+        // Add queries for the specified class
+        val classQueries = queries[className]
+        classQueries?.let { allQueries.addAll(it) }
+        // Recurse on any modules that className references
+        nestedModules[className]?.forEach { addModuleNodeQueries(it, allQueries) }
```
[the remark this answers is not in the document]
rylin8: Yep it will most likely stack overflow. I thought about it and it was going to be a stretch goal for my initial PR since I wanted to get this out before I'm out.

codegen/src/main/kotlin/com/android/designcompose/codegen/BuildModuleClass.kt
rylin8: done

annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
rylin8: Created a KSP bug since it didn't exist and reference it here.

codegen/src/main/kotlin/com/android/designcompose/codegen/BuildModuleClass.kt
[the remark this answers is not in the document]
rylin8: Added detection for circular recursion. Is that the correct term for it? Detecting when a class includes itself or includes another class that directly or indirectly includes the original class.

rylin8:
Added documentation task https://github.com/google/automotive-design-compose/issues/921

codegen/src/main/kotlin/com/android/designcompose/codegen/BuildModuleClass.kt
```diff
@@ -0,0 +1,251 @@
...
+    internal fun addIgnoredImage(className: String, nodeName: String) {
+        if (!ignoredImages.containsKey(className)) ignoredImages[className] = HashSet()
+        ignoredImages[className]?.add(nodeName)
+    }
+
+    internal fun addNestedModule(className: String, moduleName: String, logger: KSPLogger) {
+        if (!nestedModules.containsKey(className)) nestedModules[className] = HashSet()
+
+        // Check that there is no recursive nesting of module classes
+        if (hasRecursivNesting(className, moduleName)) {
```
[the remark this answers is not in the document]
rylin8: Oof spelling is hard. Fixed.

iamralpht (approved):
Looks good - let's get it merged now at the start of the cycle, and we can get a couple of weeks early testing before RC. We **must** get the docs ticket in for this release, too, but that can wait until you're back.

PR #927, opened 2024-04-02
Reduce the frequency of Dependabot version updates by only scheduling it to run on Fridays.

timothyfroehlich:
This will make it easer to batch approve them and reduce the time spent on updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -33,10 +33,10 @@ updates:
       github-workflow-dependency-patches:
         update-types:
         - "patch"
-    open-pull-requests-limit: 3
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -117,12 +120,11 @@ updates:
     assignees: [timothyfroehlich]
     reviewers: [timothyfroehlich]
     target-branch: "main"
-    # Dependabot groups seem to be busted for Gradle. Uncommenting this will prevent
```
timothyfroehlich, line 120: I believe this is fixed, but we'll see.

Issue #915, opened 2024-04-01
Refactor `DesignDocOverride` into more generic `DesignSettings`, and make Squoosh runtime enable-able


iamralpht:
Let's make Squoosh runtime-enable-able instead of a build-time constant. I think we should refactor `DesignDocOverride` to take multiple optional settings, e.g.:

```kotlin
DesignSettings(
  overrideDocId? = null,
  useSquoosh = false,
  openLinkCallback? = null,
  dcCallbacks? = null,
  content: @Composable () -> Unit
)
```

Then you could specify any of these settings without us extending the arguments to the generated Composables. Additionally, we could remove some of the extra args (like `openLinkCallback` and friends) from the generated Composables and clean up the type signatures.

Once we can turn on Squoosh at runtime, let's turn it on for DriverUI at least.

timothyfroehlich:
Questions:

DesignDocOverride affected all Figma documents in the Compose tree. Do we want to fix that here and make DesignSettings be specific to instances of the `*Gen` interfaces?

What do we want to do about openLinkCallback? I don't have a good definition of how it's supposed to work right now, so other than making sure that the one Validation test still works, I don't know how to fix any issues with it.

What do we want to do about backwards compatibility with 0.25? Should I try to make this a seamless change to the main composable?

iamralpht:
Good questions!

1. I think it's OK for squoosh/openLink, etc to propagate all the way. The doc override is the only one that's dubious, and we'd upgrade that by saying "everything from this @DesignDoc/interface should now come from this other docId" as a map, rather than by limiting how far the settings propagates into the tree.
2. Let's ensure that validation passes for openLinkCallback, we can expand test cases too if you see the opportunity once you get into the change.
3. Don't worry about backwards compatibility. We're not API stable yet, so this should be OK ;)

timothyfroehlich:
Next question: Why a CompositionLocal? Each of these settings, except for Squoosh, only apply to the root composable. I think it makes more sense to just provide this object as a parameter for the root composable, which would not pass it to the child composables.

Going along with that, why is DesignDocOverride a CompositionLocal? Why not have the docID just be an argument to the root composable? The docID should only be relevant to the root one, the one that could have the DesignSwitcher and therefore could change the ID, so there's no reason for it to be a CompositionLocal which could be accessed by everything below it.

iamralpht:
`CompositionLocal` is nice because:

[...]

PR #955, opened 2024-04-03
Allow Squoosh to be enabled at runtime, add DesignDocSettings

timothyfroehlich:
This is implemented as a class that can be applied to a DesignCompose composable via a localComposition.


timothyfroehlich (commented):
First step for #915. Upon approval I'll continue with the other changes that were requested.

designcompose/src/main/java/com/android/designcompose/DesignDocSettings.kt
```diff
@@ -0,0 +1,42 @@
...
+ *   better integrate with external layout.
+ */
+class DesignDocSettings(
+    val useSquoosh: Boolean = false,
+)
+
+internal val LocalDesignDocSettingsContext = compositionLocalOf { DesignDocSettings() }
+
+@Composable
+fun WithDesignDocSettings(settings: DesignDocSettings, content: @Composable () -> Unit) {
```
iamralpht, line 40: Is the `With` a naming convention in Compose? I don't mind it but I'm not sure if I've seen it before...

timothyfroehlich: Yeah, actually when I thought about it I realized I was just dumping syntactic sugar on top of a normal Compose thing, so I just removed the function altogether. I also cleaned up the compositionLocal's name based on the [style guide](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#naming-compositionlocals).

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/HelloSquoosh.kt
```diff
@@ -0,0 +1,31 @@
...
+package com.android.designcompose.testapp.validation.examples
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import com.android.designcompose.DesignDocSettings
+import com.android.designcompose.LocalDesignDocSettings
+
+@Composable
+fun HelloSquoosh() {
+    CompositionLocalProvider(LocalDesignDocSettings provides DesignDocSettings(useSquoosh = true)) {
```
iamralpht, line 26: Sorry, I think I gave bad feedback. This should look like:

```kotlin
DesignDocSettings(useSquoosh = true) {
  HelloWorldDoc.Main(name = "Squoosh")
}
```

iamralpht: I think, anyway, it was the "With" thing I wasn't sure about. Perhaps other CompositionLocal users provide a "With" composable.

timothyfroehlich: I just came to the conclusion that it was better to be more explicit about the mechanism that we were using, rather than hiding the implementation detail. This gives the user more flexibility to change any of the settings (like a docOverride?) at runtime. [Material's docs ](https://developer.android.com/develop/ui/compose/designsystems/material#emphasis) directly use CompositionLocal so I don't think it should be an issue for us to use it here.

iamralpht: Oh, cool -- doing what Material does is good for me! Thanks for finding this, Guru!

PR #971, opened 2024-04-09
Fix #970: add initial automated test for variant animation implementation

### LINKED ISSUE #970: Test animation implementation
Squoosh provides an animation implementation, but we don't have any automated testing of animations. This ticket is to add some automated tests for the basic variant auto-animation logic; subsequent animation work can then bring in new tests.

iamralpht:
Use ComposeTestRule's support for manual clock advancement to control the timeline of one of the basic "variant animation" cases. If CI is able to run this OK, then we can expand with coverage of animation interruption and other nice features.


iamralpht:
@timothyfroehlich will the screenshots generated by the new test case that I added get compared automatically? (I think so, since they get generated by the `recordRoborazziDebug` target).

Adding this test case was great, btw -- really easy to figure out what to do based on the existing special case for interaction and the generic AllExamples one.

integration-tests/validation/src/testDebug/roborazzi/RenderAllExamples/CrossAxis-Fill.png
rylin8, line 1: What changed in this image and other ones that look the same?

iamralpht: There doesn't seem to be any difference on the layout files; I ended up with a diff when I ran the gradle verifyAndRecord step, but the CI thing doesn't indicate any difference.

For the two tests that I switched over to use squoosh, there's some tiny difference in antialiasing.

I can try reverting the changes to the two layout images if you like (I'm not sure why only these two ended up with new files, either).

timothyfroehlich: Yep, any pixel difference will show up so it's not surprising that rendering via Squoosh will cause an image diff.

timothyfroehlich (commented):
> @timothyfroehlich will the screenshots generated by the new test case that I added get compared automatically? (I think so, since they get generated by the `recordRoborazziDebug` target).

Yep, they'll show up the next time a difference is detected.

timothyfroehlich (commented):
> Oh, also, really nice that this all runs host side. Somehow my emulator setup got corrupted (can't install -- no space -- but lots of space on the device) so I switched to acid/CF... but this is even faster! 

It's so nice :) Just imagine if we built up full Compose Desktop support and we could run a whole mock app on your desktop!

timothyfroehlich (approved):
Looks good! Just a minor nit.

rylin8 (approved):

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/AnimationMidpoints.kt
```diff
@@ -0,0 +1,99 @@
...
+import org.robolectric.annotation.GraphicsMode
+
+@RunWith(AndroidJUnit4::class)
+@GraphicsMode(GraphicsMode.Mode.NATIVE)
+@Config(qualifiers = RobolectricDeviceQualifiers.MediumTablet, sdk = [34])
+class AnimationMidpoints {
+    @get:Rule val composeTestRule = createAndroidComposeRule<ComponentActivity>()
+    @get:Rule val roborazziRule = designComposeRoborazziRule(javaClass.simpleName)
+
+    @Before fun setup() {}
```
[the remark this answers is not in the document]
iamralpht: Good spotting - removed.

Issue #978, opened 2024-04-10
Layout calculation logging spams debug builds


johnoneil:
This is pretty simple. I'll try to provide a PR for this.

We've got a basic integration using the `v0.25.0` layout support, but when using it we get some log spam in debug builds:

```
2024-04-10T20:48:35.096Z INFO  [layout::layout] compute_node_layout 0
2024-04-10T20:48:35.429Z INFO  [layout::layout] compute_node_layout 0
2024-04-10T20:48:35.763Z INFO  [layout::layout] compute_node_layout 0
2024-04-10T20:48:36.096Z INFO  [layout::layout] compute_node_layout 0
2024-04-10T20:48:36.429Z INFO  [layout::layout] compute_node_layout 0
2024-04-10T20:48:36.439Z INFO  [layout::layout] compute_node_layout 0
2024-04-10T20:48:36.762Z INFO  [layout::layout] compute_node_layout 0
```

I'm going to suggest we move that to `trace!` level logging.

rylin8:
Yea that should be fine, sorry there used to be a lot more logging and I reduced most of it but missed a couple I think.

rylin8:
@johnoneil can we close this now?

rylin8:
Closing as pretty sure this is fixed. Please reopen if not.

PR #979, opened 2024-04-10
#978 Reduce layout log spam

johnoneil:
See #978 

Reduce log spam when using new layout code.


iamralpht (approved):
Looks good to me, we do have a lot of log spam (I started filtering to just see errors recently -- would have been better to have made this change instead). @rylin8 do you need these messages at all, or would you rather remove them entirely?

johnoneil:
Merging. Thank everyone for your help on this. We may try to pick this up in the release tomorrow (pending discussions).

PR #1006, opened 2024-04-15
Refactor the layout and figma_import crates

timothyfroehlich:
As a result of this change the `figma_import` crate now depends on the
`layout` crate, instead of the `layout` crate depending on
`figma_import`.

As a result, the `layout` crate now has no dependencies within the
workspace.

`layout`'s tests had to be moved to `figma_import` due to the tests heavy reliance on `figma_import`.


rylin8:
Not sure why there is both a `styles.rs` and `types.rs`?

timothyfroehlich:
> Not sure why there is both a `styles.rs` and `types.rs`?

Personal preference for smaller files. "Styles" was the best name I could think of for the enums, which is what they correspond to in Taffy. I would have called it `layout_styles.rs` but that's too close to `layout_style`.

rylin8:
> > Not sure why there is both a `styles.rs` and `types.rs`?
> 
> Personal preference for smaller files. "Styles" was the best name I could think of for the enums, which is what they correspond to in Taffy. I would have called it `layout_styles.rs` but that's too close to `layout_style`.

Sure but what goes in one vs the other? Like I don't get why `Dimension` is in one but `JustifyContent` is in the other.

timothyfroehlich:
> > > Not sure why there is both a `styles.rs` and `types.rs`?
> > 
> > 
> > Personal preference for smaller files. "Styles" was the best name I could think of for the enums, which is what they correspond to in Taffy. I would have called it `layout_styles.rs` but that's too close to `layout_style`.
> 
> Sure but what goes in one vs the other? Like I don't get why `Dimension` is in one but `JustifyContent` is in the other.

Because `Dimension` is a basic type and gives you positioning and `JustifyContent` is just one way you can lay out the content, just like `AlignItems` and so on. 

It could be a bit more clear to rename `types.rs to `geometry.rs`, I suppose.

rylin8:
Gave this a test and seems fine, so I'm good with it.

rylin8 (approved):

PR #1007, opened 2024-04-15
Closes #948: Remove support for interface inheritance

### LINKED ISSUE #948: Remove support for overriding DesignDoc generated interfaces
This was a bit of a hack and the use cases it supported are now supported by DesignModules. This is preventing us from updating the Compose Compiler past 1.5.8 ( #860 )

Blocked by #928

rylin8:
With design modules available, there should be no need for interface inheritance in @DesignDoc interfaces. Remove support for them and show an error if inheritance is attempted. In order to support upgrading the Compose compiler, which doesn't support default args in open functions, change the generated code to create a class instead of an interface.

### DIFF
--- a/codegen/src/main/kotlin/com/android/designcompose/codegen/BuilderProcessor.kt
+++ b/codegen/src/main/kotlin/com/android/designcompose/codegen/BuilderProcessor.kt
@@ -166,8 +166,6 @@ class BuilderProcessor(private val codeGenerator: CodeGenerator, val logger: KSP
         private var queriesNameSet: HashSet<String> = HashSet()
         private var ignoredImages: HashMap<String, HashSet<String>> = HashMap()
 
-        private var overrideInterface: String = ""
-
         private lateinit var out: OutputStream
         private var currentJsonStream: OutputStream? = null
         private var designDocJson: JsonObject = JsonObject()
@@ -193,18 +191,14 @@ class BuilderProcessor(private val codeGenerator: CodeGenerator, val logger: KSP
 
             currentJsonStream = jsonStreams[packageName]
 
-            // If the interface inherits from another interface, get the name of it
+            // We don't support inheritance from another interface. Check for this and emit an
+            // error if we find a superclass.
             classDeclaration.superTypes.forEach {
                 val superType = it.resolve()
                 // All classes that don't declare a superclass inherit from "Any", so ignore it
                 if (superType.toString() != "Any") {
-                    if (superType.isError)
-                        logger.error("Invalid supertype for interface hg $classDeclaration")
-                    overrideInterface = superType.declaration.qualifiedName?.asString() ?: ""
-                    if (overrideInterface.endsWith("Gen"))
-                        logger.error(
-                            "Extending a generated interface $overrideInterface not supported"
-                        )
+                    val overrideInterface = superType.declaration.qualifiedName?.asString() ?: ""
+                    logger.error("Extending a generated interface $overrideInterface not supported")
                 }
             }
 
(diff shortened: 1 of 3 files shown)


codegen/src/main/kotlin/com/android/designcompose/codegen/BuilderProcessor.kt
```diff
@@ -353,7 +340,7 @@ class BuilderProcessor(private val codeGenerator: CodeGenerator, val logger: KSP
             out.appendText("}\n\n")
 
             val objectName = className + "Doc"
-            out.appendText("object $objectName: $interfaceName {}\n\n")
+            out.appendText("val $objectName = $interfaceName()\n\n")
```
timothyfroehlich, line 343: I wonder if we still need this. Removing it would be a breaking change, but I can't think why statically instantiating an instance of it is a good thing.

I suppose the same question could have been asked about the original line too.

codegen/src/main/kotlin/com/android/designcompose/codegen/BuilderProcessor.kt
```diff
@@ -599,30 +578,28 @@ class BuilderProcessor(private val codeGenerator: CodeGenerator, val logger: KSP
...
-            // Add a modifier for the root item. Default params not allowed with override
-            val defaultModifier = if (override) "" else " = Modifier"
+            // Add a modifier for the root item.
+            val defaultModifier = " = Modifier"
             args.add(Pair("modifier", "Modifier$defaultModifier"))
 
             // Add an option to register a function to handle open link callbacks
-            val defaultOpenLink = if (override) "" else " = null"
+            val defaultOpenLink = " = null"
             args.add(Pair("openLinkCallback", "OpenLinkCallback?$defaultOpenLink"))
```
timothyfroehlich, line 595: Can we just combine these two lines (and the others like it?)

codegen/src/main/kotlin/com/android/designcompose/codegen/BuilderProcessor.kt
```diff
@@ -312,12 +303,8 @@ class BuilderProcessor(private val codeGenerator: CodeGenerator, val logger: KSP
...
             // contains all the variant properties and values
             out.appendText("    @Composable\n")
-            // Add an override declaration if this interface inherits from another
-            val overrideDecl = if (overrideInterface.isNotEmpty()) "override " else ""
-            out.appendText("    ${overrideDecl}fun CustomComponent(\n")
-            // Default parameters are not allowed when overriding a function
-            val defaultModifierDecl = if (overrideInterface.isEmpty()) " = Modifier" else ""
-            out.appendText("        modifier: Modifier$defaultModifierDecl,\n")
+            out.appendText("    final fun CustomComponent(\n")
+            out.appendText("        modifier: Modifier,\n")
```
timothyfroehlich, line 307: Seems like you lose the ` = Modifier` bit here.

timothyfroehlich (changes requested):
Mostly looks good, and this does resolve the compile errors in #870

codegen/src/main/kotlin/com/android/designcompose/codegen/BuilderProcessor.kt
rylin8: yep, meant to do that.

codegen/src/main/kotlin/com/android/designcompose/codegen/BuilderProcessor.kt
rylin8: Yea I don't think that's needed. This function is called from generated code and a modifier is always passed in.

codegen/src/main/kotlin/com/android/designcompose/codegen/BuilderProcessor.kt
rylin8: It would mean all existing code would have to instantiate the object manually then wouldn't it? It'd break everything.

timothyfroehlich (approved):

timothyfroehlich (approved):
LGTM

timothyfroehlich:
I know we don't do release notes, but if we did this would be a good thing to mention @iamralpht.

PR #1010, opened 2024-04-17
Closes #921: Add documentation for flattening customizations into modules

### LINKED ISSUE #921: Add documentation for flattening customizations into modules

rylin8:


yiqunw700 (approved):

docs/_docs/modules.md
```diff
@@ -0,0 +1,64 @@
...
+## Usage in a function
+Module classes by themselves are not useful until used as a parameter in a `@DesignComponent` function. Use the `@DesignModule` annotation for parameters in such a function. The generated function will combine any customizations `@Design` and `@DesignVariant` parameters in the function with all the embedded customizations from parameters that are modules.
+
+### Example:
+```kotlin
+interface ModuleExample {
+    @DesignComponent(node = "#stage")
+    fun Main(
+        @Design(node = "#header") header: String,
+        @DesignModule moduleCustomizations: TextModuleCombined,
```
timothyfroehlich, line 59: I think we need usage example of this, showing the generated composable and how the user would set values for the customizations.

timothyfroehlich (changes requested):

docs/_docs/modules.md
```diff
@@ -0,0 +1,64 @@
...
+Module classes are declared with the `@DesignModuleClass` annotation. The annotation takes no arguments and simply annotates the class so that a customizations() extension function will be generated for the class. Within the class, customizations can be declared in the same way that customizations are declared in a `@DesignComponent` function. However, due to a [bug in KSP][1], different annotations to specify customizations need to be used when declaring them within a module class vs a function. Here is a table of what annotation to use in a function vs a class:
+
+| Function              | Class                         |
+| --------------------- | ----------------------------- |
+| @Design               | @DesignProperty               |
+| @DesignVariant        | @DesignVariantProperty        |
+| @DesignContentTypes   | @DesignContentTypesProperty   |
+| @DesignPreviewContent | @DesignPreviewContentProperty |
+
+Note that once this bug has been fixed, all of the `@Design*Property` annotations will be removed in favor of the analogous annotations used for functions.
```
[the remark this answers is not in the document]
rylin8: done

docs/_docs/modules.md
rylin8: done

docs/_docs/modules.md
```diff
@@ -0,0 +1,90 @@
...
+interface ModuleExample {
+    @DesignComponent(node = "#stage")
+    fun Main(
+        @Design(node = "#header") header: String,
+        @DesignModule moduleCustomizations: TextModuleCombined,
+    )
+}
+```
+
+To call the generated function, an instance of TextModuleCombined must be passed into the `moduleCustomizations` parameter. This can be constructed normally since it uses the developer defined class directly, not a generated one.
```
timothyfroehlich, line 65: Keep this line generic and if you feel you need a bit more clarification you can add comments  into the code block below.

```suggestion
To call the generated function, an instance of the design module class must be passed into the design module parameter. This can be constructed normally since it uses the developer-defined class directly, not a generated one.
```

docs/_docs/modules.md
```diff
@@ -0,0 +1,90 @@
...
+            replaceNode = { ModuleExample.Button("My Button") },
+            textStyle = TextStyle.LargeBold,
+            content = {
+                ListContentData(count = 3) { index ->
+                    ModuleExample.Button("Button $index")
+                }
+            },
+        ),
+    )
+}
```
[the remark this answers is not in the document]
rylin8: Because that's how I would imagine it would often work when using modules, since a module would likely be shared and you'd have some common code to construct one. I don't really feel strongly about this either way though, so I've removed it.

docs/_docs/modules.md
timothyfroehlich: This is a clarity issue. Throughout the rest of the file you're describing how to use the annotations, but in this line you describe how to use the examples that you'd created. Someone reading the documentation but not paying close attention to the examples (or were reading through the main text first) would then need to go see what `TextModuleCombined` and `moduleCustomizations` are to understand that you're saying to add an instance of the design module class to the parameter.

rylin8: It's describing the examples because it is under the `Example` subheader. Do you want me to just move this into the code as a comment?

docs/_docs/modules.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+To call the generated function, an instance of TextModuleCombined must be passed into the `moduleCustomizations` parameter. This can be constructed normally since it uses the developer defined class directly, not a generated one.
+
+```kotlin
+ModuleExampleDoc.Main(
+    headerText = "Module Example",
+    moduleCustomizations = TextModuleCombined(
```
timothyfroehlich, line 70:
```suggestion
    moduleCustomizations = TextModuleCombined( // passing in the original `@DesignModuleClass`
```

docs/_docs/modules.md
timothyfroehlich: Ok, I didn't realize that. This is the only place that you mention how to use the generated function so I assumed that this line was supposed to be part of the general documentation, not something explaining the example.

So here's the underlying problem, and this is what I was getting at in my first comment saying "please add an example of how to use it". The "Defining a module class" section tells you how to create a module and "Nested Modules" shows you how to add them to another `DesignModule`. Then you have "Usage in a function". The first sentance doesn't actually tell you how to use it in a function, it tells you how you can add the `DesignModule` to a `DesignComponent`, and this is what the first code block shows. Then you say "the generated function will combine any customizations...", which describes what the generated code will look like, but not how you would use the "use a DesignModule" in a generated function. It won't be obvious what this means, it certainly isn't obvious to me.


So what I expect from a section called "Usage in a function" is "how do write the correct syntax that writes the values for the `DesignModule` in the `DesignComponent` that uses it". This is what I thought the "to call the generated function" line was doing, explaining how to write the code, so I assumed you had made a mistake by using the names from the example code.


So let me sum that all up and rephrase the review comments with some additional context: The "Usage in a function" section doesn't describe how to use it in a function. Please rename it to "Adding to a `DesignComponent`". Then add a new section named "Usage in a DesignCompose function" that explains how to assign values to the `DesignModule` customizations in the generated DesignCompose functions in generic terms, just like how the other sections are written using generic terms. You will probably want to include the last example code block in this new section. Also, make sure that all text that refers to example code clearly says so by prefacing the text with something like "in the below example" or by including the text as comments within the example code.

rylin8: I called it "Usage in a function" because the `@DesignComponent` annotation is for a function, so you are using it for a function. It was supposed to encapsulate both the declaration of the function and the calling of the generated function. I've updated per your suggestion... I think? Though I have changed the section titles.

timothyfroehlich (approved):

PR #1036, opened 2024-04-24
Check size of stops of gradient background

yiqunw700:
Figma is returning empty stops for gradients. Gradients can also have a single stop set.
Also add logging when size is smaller than 2.

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/Utils.kt
+++ b/designcompose/src/main/java/com/android/designcompose/Utils.kt
@@ -1230,42 +1230,84 @@ internal fun Background.asBrush(document: DocContent, density: Float): Pair<Brus
         }
         is Background.LinearGradient -> {
             val linearGradient = this
-            return Pair(
-                RelativeLinearGradient(
-                    linearGradient.color_stops.map { convertColor(it.field1) }.toList(),
-                    linearGradient.color_stops.map { it.field0 }.toList(),
-                    start = Offset(linearGradient.start_x, linearGradient.start_y),
-                    end = Offset(linearGradient.end_x, linearGradient.end_y)
-                ),
-                1.0f
-            )
+            when (linearGradient.color_stops.size) {
+                0 -> {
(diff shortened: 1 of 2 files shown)


designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -1230,42 +1230,84 @@ internal fun Background.asBrush(document: DocContent, density: Float): Pair<Brus
...
-                    linearGradient.color_stops.map { it.field0 }.toList(),
-                    start = Offset(linearGradient.start_x, linearGradient.start_y),
-                    end = Offset(linearGradient.end_x, linearGradient.end_y)
-                ),
-                1.0f
-            )
+            when (linearGradient.color_stops.size) {
+                0 -> {
+                    Log.e(TAG, "No stops found for the linear gradient")
+                    return null
```
rylin8, line 1236: This should never happen because Rust will return `Background::None` right? (good to have it still, just checking)

rylin8:
Looks good! Thanks for fixing this quickly! I would test it but my machine is not setup fully yet.

designcompose/src/main/java/com/android/designcompose/Utils.kt
iamralpht: Yeah, I think you get here if you have an old serialized doc.

iamralpht (approved):

designcompose/src/main/java/com/android/designcompose/Utils.kt
rylin8: 👍

rylin8 (approved):

rylin8:
Finally have a semi-working machine. Built this, and found that even HelloWorld's gradient is broken :(. Hope Figma fixes this soon. If we were to update our .dcf files, I think all tests with gradients would change.
Screenshot 2024-04-24 at 11 22 26 PM

PR #1053, opened 2024-04-28
Fix #1052: rewrite squoosh animation tree building to not mutate trees.

### LINKED ISSUE #1052: Squoosh: animation: root frames of variants aren't included in animations
If the root of a component has a different style or color, then we don't animate it. It just snaps from one appearance to the other. We should animate roots.

iamralpht:
While attempting to fix a bug in squoosh's animation support where the root frames of components were not considered for animation, I kept running into layout assertions, which ultimately were because the animation implementation performs destructive mutations on the base layout tree AND on the "transition layout tree".

This change means that all `SquooshResolvedNode` trees are now immutable after construction (except for layout and style values which can change after the tree was built).

Finally, because the source trees are immutable, we can run layout on the root and target root trees whenever layout needs to change. This means animations are uninterrupted by relayouts, and we can relayout as many times as we like without generating new trees.


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -132,173 +145,240 @@ internal class SquooshAnimationControl(
...
+            // Clone the target node, then look at its siblings, then at its child.
+            Pair(from.cloneSelf(parent), false)
+        }
+
+    // First recurse along siblings
+    val nextFrom = from.nextSibling
+    if (nextFrom != null) {
+        val nextCloned = createMergedAnimationTree(nextFrom, to, requestedAnimations, parent)
+        // Insert at the end; does this reorder? Hope not
+        var c: SquooshResolvedNode? = cloned
```
rylin8, line 204: Odd that this needs `?`

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -132,173 +145,240 @@ internal class SquooshAnimationControl(
...
+    SquooshResolvedNode(
+        view = this.view,
+        style = this.style,
+        layoutId = this.layoutId,
+        textInfo = this.textInfo,
+        unresolvedNodeId = this.unresolvedNodeId,
+        layoutNode = this,
+        parent = parent,
+        computedLayout = this.computedLayout,
+        needsChildRender = this.needsChildRender
```
rylin8, line 231: Do you want to set `firstChild` and `nextSibling` to `null`?

iamralpht:
I noticed (a few months ago, when last I worked on this!) a regression in the "Same Name" animation test. I've made that test into a unit test as well as an interactive one, and fixed the bug. I think this is ready to merge, but I haven't gone through @rylin8 's most recent comments.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
iamralpht: It's because I do `c = c.nextSibling` even though it's guarded. I think this linked list code is really quite hard going, although it seems to run fairly fast. Maybe if we move it to Rust we can use the same thing `expanse` uses for its tree and it'll be fast enough without being so hard to read?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
iamralpht: No, I think they're always initialized to `null` anyway

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -704,6 +694,6 @@ fun SquooshRoot(
                 )
             }
         )
-        designSwitcher()
+        // designSwitcher()
```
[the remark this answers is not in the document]
iamralpht: Oops, will fix.

iamralpht:
The midpoint diff should be a bit more visible when it gets rendered -- I changed the midpoint time we sample to get something that was more "middle" than before, so that test is expected to be quite different (but I manually inspected the new expectation and believe it's correct).

rylin8 (approved):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshLayout.kt
```diff
@@ -218,3 +218,11 @@ internal fun layoutTree(
     layoutValueCache.putAll(updatedLayouts)
     populateComputedLayout(root, layoutValueCache)
 }
+
+/// Go over a tree that layout can't be applied to, and copy the
+/// layout values from whatever the source tree is.
+internal fun updateDerivedLayout(n: SquooshResolvedNode) {
+    if (n.layoutNode != null) n.computedLayout = n.layoutNode.computedLayout
+    if (n.firstChild != null) updateDerivedLayout(n.firstChild!!)
```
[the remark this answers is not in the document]
iamralpht: I left it how it was because I have a few other lines of 'if not null', and I couldn't make them so clear (so I felt it was better to be clunky but similar from line to line...)

PR #1055, opened 2024-05-01
Closes #951: Initial support for variables and design tokens

### LINKED ISSUE #951: Support Design Tokens for Color Values
We'd like to support Design Tokens, starting with tokens for colors. The goal is for a DesignCompose app to be able to override the colors of different design elements at runtime, based on token values that are computed at runtime.

There are a few elements to this, which should be covered in a design doc ahead of implementation:

 * How are design tokens specified in Figma? We've previously looked at Figma Variables and at the Material Design Token Plugin (which uses Figma Styles, but is a bit hard to work with because you can only see one set of token values in a doc at a time). We need to decide on a supported approach or approaches.
 * How are design token values communicated in Kotlin? Most likely, we will have some low level interface that proves a `Map<String, Color>` to go from the token name to color value, and then a layer on top that can go from AAOS OEM Design Token definitions into that map.
 * How does the DC renderer look up token values from the map when it renders?
 * (Do we want to only allow `Color`, or just say `Brush`, and be more generic to what can be used?).

rylin8:
Use the Figma API to parse variables. There are three main areas we now parse:
1. boundVariables from nodes that use variables. This change supports background color, stroke color, text color, and rounded corner sizes.
2. explicitVariableModes from nodes where Figma set a mode to be used on that node and children.
3. All the variable data (collections, modes, variables) from a new variables API call.

boundVariables and explicitVariableModes are stored in the view generated for each node. The rest of the variable data is stored in a variable map data structure stored in the serialized file.

Three new functions are introduced to override the current theme or mode.
1. DesignVariableCollection(themeName) declares a CompositionLocalProvider object that changes the current theme (collection in Figma) to the specified theme name. This theme name must exist as a collection in the Figma file, or it will do nothing.
2. DesignVariableModeValues(modeValues) declares a CompositionLocalProvider hash map that maps collection names to mode names. This changes the mode for the specified collection. For example if the default mode in Figma is set to 'light', this can be used to change the mode to 'dark'.
3. DesignMaterialThemeProvider(Boolean) declares a CompositionLocalProvider boolean to specify whether the material theme on the device should be used to override the theme in Figma. This only works for variables that were created from the Material Theme Builder plugin, or were made in a way to match -- e.g. the collection name is "material-theme" and the variable names match the Material Theme names such as "Schemes/Primary".

When it is time to render a node that uses variables, we need to translate the variable into a value. This is achieved by using the helper class VariableManager, which takes the current VariableState containing the current collection and mode, along with the variable map in the serialized file, and looks up the correct value for the variable.


integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariableModesTest.kt
```diff
@@ -0,0 +1,272 @@
...
+    Material("material-theme"),
+    MyTheme("my-theme"),
+}
+
+@DesignDoc(id = "HhGxvL4aHhP8ALsLNz56TP")
+interface VariablesTest {
+    @DesignComponent(node = "#stage")
+    fun Main(
+        // @DesignVariableCollection(node = "#stage") theme: Theme?,
+        @DesignVariableMode(node = "#Left", collection = "material-theme")
```
rylin8, line 70: This method of setting a mode uses an annotation to set the mode on a node. Note that you must also specify the collection (i.e. "theme", but called a colleciton in Figma), since each collection has a set of modes that can have the same name. We could alternatively remove this annotation in favor of allowing the developer to wrap their code around a compositionLocal object specifying the mode.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariableModesTest.kt
```diff
@@ -0,0 +1,272 @@
...
+@Composable
+fun VariableModesTest() {
+    val theme = remember { mutableStateOf<Theme?>(null) }
+    val leftModeMaterial = remember { mutableStateOf(LightDarkMode.Light) }
+    val rightModeMaterial = remember { mutableStateOf(LightDarkMode.Dark) }
+    val leftModeMyTheme = remember { mutableStateOf(LightDarkMode.Light) }
+    val rightModeMyTheme = remember { mutableStateOf(LightDarkMode.Dark) }
+
+    println("### Theme ${theme.value?.name}")
+    DesignVariableCollectionProvider(theme.value?.themeName) {
```
rylin8, line 182: This method of setting a theme/collection uses a compositionLocal object. The developer wraps their composable around this, saying that all composables under this should use the specified theme. We could alternatively remove this function in favor of allowing the developer to use an annotation to specify the collection, copying how a mode is set. I've provided different implementations of setting the mode and collection to demonstrate what it would look like, but ultimately I think they should be the same.

rylin8 (commented):
This initial implementation supports variables for colors (background, stroke, and text color), and rounded corner sizes.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariableModesTest.kt
timothyfroehlich: So does this hard-code the collection in?

timothyfroehlich (commented):
I'm leaning into the CompositionLocal method. It lines up with how Material already works which is a big plus. I think the downside of it is that it makes it harder to change the collection or mode partway down the tree, but that seems like that can just be handled by doing a ComponentReplacement with a node that handles that part. It'd be ugly, since you need to change your Figma design to use/replace the placeholder, but it also (probably?) won't be that common.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariableModesTest.kt
rylin8: No, this is just saying, "for the collection called 'material-theme', set the mode to the value in the parameter".

timothyfroehlich: But what if you wanted to change the collection at runtime?

rylin8: Sorry I should have been more clear. In this implementation, the way to set the mode is to use this annotation, and the way to set the theme is to use `DesignVariableCollectionProvider`. If we were to decide to go with the annotation method, I would add another annotation like `@DesignVariableCollection` to set the collection. If we were to decide to go with the composition local method, I would add another one like `DesignVariableModeProvider` to set the mode.

crates/figma_import/src/toolkit_schema.rs
```diff
@@ -19,13 +19,55 @@ use std::sync::atomic::AtomicU16;
...
+}
+impl ColorOrVar {
+    pub(crate) fn from_var(
+        bound_variables: &figma_schema::BoundVariables,
+        var_name: &str,
+        color: &FigmaColor,
+    ) -> ColorOrVar {
+        let var = bound_variables.get_variable(var_name);
+        if let Some(var) = var {
+            ColorOrVar::Var(var)
```
timothyfroehlich, line 65: This is largely duplicated code. Can we make a generic Variable<T> type?

crates/figma_import/src/toolkit_schema.rs
```diff
@@ -230,3 +278,140 @@ impl View {
...
+            figma_schema::Variable::Color { common, values_by_mode } => Variable {
+                id: common.id.clone(),
+                name: common.name.clone(),
+                remote: common.remote,
+                key: common.key.clone(),
+                variable_collection_id: common.variable_collection_id.clone(),
+                var_type: VariableType::Color,
+                values_by_mode: VariableValueMap::from_figma_map(values_by_mode),
+            },
+        }
```
timothyfroehlich, line 400: This is also very duplicated, can this also be generic, and can this be combined with the "XorVar" structs?

timothyfroehlich (changes requested):

designcompose/src/main/java/com/android/designcompose/MaterialThemeValues.kt
```diff
@@ -0,0 +1,113 @@
...
+
+package com.android.designcompose
+
+import androidx.compose.ui.graphics.Color
+
+// Collection name for material theme generated from Material Theme Builder plugin
+const val MATERIAL_THEME_COLLECTION_NAME = "material-theme"
+
+// Material theme variable names for variables generated from Material Theme Builder plugin
+object MaterialThemeConstants {
```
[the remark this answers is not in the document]
rylin8: There are no strings in there. I need strings to match with the variable names created in Figma.

crates/figma_import/src/toolkit_schema.rs
```diff
@@ -19,13 +19,55 @@ use std::sync::atomic::AtomicU16;
...
+use crate::figma_schema;
 use crate::reaction_schema::FrameExtras;
 use crate::reaction_schema::Reaction;
 use crate::toolkit_style::{StyledTextRun, ViewStyle};
+use std::collections::HashMap;
 
-pub use crate::figma_schema::OverflowDirection;
-pub use crate::figma_schema::StrokeCap;
+pub use crate::figma_schema::{FigmaColor, OverflowDirection, Rectangle, StrokeCap, VariableAlias};
+
```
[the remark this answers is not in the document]
rylin8: sure we are, but this initial change doesn't. See the [design doc](https://docs.google.com/document/d/1G1te8eKqw8cmaiZSIE5Px37rtjIbT67H4ZVYBfR3Cr4/edit?tab=t.0#heading=h.k90u9y7ul7d1)

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariableModesTest.kt
```diff
@@ -0,0 +1,249 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.android.designcompose.ComponentReplacementContext
+import com.android.designcompose.DesignMaterialThemeProvider
+import com.android.designcompose.DesignVariableCollection
+import com.android.designcompose.DesignVariableModeValues
+import com.android.designcompose.annotation.Design
```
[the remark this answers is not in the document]
rylin8: Why do you need another?

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1263,25 +1270,47 @@ fn visit_node(
         Vec::new()
     };
```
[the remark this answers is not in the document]
rylin8: What specifically are you referring to?

crates/figma_import/src/toolkit_schema.rs
rylin8: This Variable type contains data about the variable such as its id, collection, and all its values for each mode. It's different than the "XorVar" types, so I think it should be separate, but I will clean this code up to use a common initialization function.

crates/figma_import/src/toolkit_schema.rs
rylin8: The idea of the "XorVar" types is that Figma nodes that use variables are typically either set to a fixed value, such as the color #FF0000, or a variable ID, such as "VariableId:12345". Hence I came up with "ColorOrVar", since the background of the node is one of the two values in the enum. So Figma nodes that use variables use these "XorVar" values, whereas the large variable table that holds all variables in a Figma file use the `Variable` struct below.

gradle/libs.versions.toml
```diff
@@ -58,6 +58,7 @@ robolectric = "4.12.1"
 roborazzi = "1.15.0"
 coreKtx = "1.13.1"
 appcompat = "1.6.1"
+material3Android = "1.2.1"
```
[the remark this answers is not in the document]
rylin8: Fixed

designcompose/src/main/java/com/android/designcompose/MaterialThemeValues.kt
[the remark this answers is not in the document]
timothyfroehlich: I feel like we could use something like https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Class.html#getDeclaredFields() to parse them out automatically, though that would be a little hacky.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariableModesTest.kt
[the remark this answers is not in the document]
timothyfroehlich: Because the current test is static and doesn't change the collections or modes at runtime.

crates/figma_import/src/transform_flexbox.rs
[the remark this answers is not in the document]
timothyfroehlich: I thought I'd selectecd a range, but I guess not. I was referring to the section of code below this comment, lines 1287 to 1321, handling the corner radiii.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariableModesTest.kt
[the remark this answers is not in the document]
timothyfroehlich: I think it's worth having a test to make sure that changing the mode (like daytime to night time) works properly.

crates/figma_import/src/toolkit_schema.rs
timothyfroehlich: Makes sense to keep the things that are set separate from the entire list, but I was specifically referring to creating a generic XorVar<T> type so that we don't have four sets of largely duplicated code.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariableModesTest.kt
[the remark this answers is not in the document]
rylin8: There are buttons to change the theme, mode, and use the runtime material theme. I was going to add a test to click these buttons and generate screenshots but wanted to get some initial feedback. I'll add that now.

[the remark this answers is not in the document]
rylin8: Added a test

crates/figma_import/src/toolkit_schema.rs
rylin8: I looked into this but not sure if that would really be worth it. I could make something like:
```
pub enum BoundVariable {
    Num(f32),
    Color(Color),
    Var(String),
}
```
to encapsulate all the different variable types, but I'd still have to make separate functions `from_num_var()`, `from_color_var()`, etc. to initialize them from different types of bound variable data. So the code savings would be minimal. Plus by keeping it the way it is, we know that a background color is a `ColorOrVar`, which is either a color value or a variable, not just a more generic `BoundVariable` that can be any type.

designcompose/src/main/java/com/android/designcompose/MaterialThemeValues.kt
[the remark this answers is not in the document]
iamralpht: Yeah, better not to use reflection if we can avoid it -- it's more surprising than a static table of strings since it can change from a dependency update, but doesn't necessarily show up as a reference.

rylin8:
@timothyfroehlich any idea why roborazzi test is showing this diff?

crates/figma_import/src/toolkit_schema.rs
timothyfroehlich: An `XorVar<T>` (or `BoundVariable<T>`) type wouldn't work? That's what I meant by [generic](https://google.github.io/comprehensive-rust/generics/generic-data.html).

yiqunw700:
> @timothyfroehlich any idea why roborazzi test is showing this diff?

It seems only the tests from the same design file have the diffs?

crates/figma_import/src/figma_schema.rs
```diff
@@ -1070,3 +1080,164 @@ pub struct ProjectFilesResponse {
...
+    Boolean,
+    Float,
+    String,
+    Color,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct VariableAlias {
+    pub r#type: String,
```
yiqunw700, line 1114: What does the "r#type" do in rust?

timothyfroehlich:
> > @timothyfroehlich any idea why roborazzi test is showing this diff?
> 
> It seems only the tests from the same design file have the diffs?

I'm really not sure. I tried to recreate this with John's PR but his isn't currently building. 

I'll try to look into it more, but I'd say that if it's the only thing holding up this PR then it's not worth waiting for. If it happens again then I'll dig into it more and hopefully it'll happen in a smaller PR where it'll be easier to isolate the changes.

crates/figma_import/src/figma_schema.rs
timothyfroehlich: "type" is a reserved keyword so using "r#type" lets you get around that.

crates/figma_import/src/figma_schema.rs
```diff
@@ -1070,3 +1080,164 @@ pub struct ProjectFilesResponse {
...
+    pub name: String,
+    pub remote: bool,
+    pub key: String,
+    pub variable_collection_id: String,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+#[serde(rename_all = "SCREAMING_SNAKE_CASE")] // Because "resolvedType" is in this format
+#[serde(tag = "resolvedType")] // Maps to different enum values based on "resolvedType"
+pub enum Variable {
```
yiqunw700, line 1202: Is this enum having Boolean, Float, String, Color mainly for the resolved type? I mean all of them have 
```
        common: VariableCommon,
        values_by_mode: HashMap<String, VariableValue>,
```
which are the same. So I am wondering why have four enum instead of just `struct Variable` with common, values_by_mode and resolved_type?

yiqunw700: I also see below in tookit_schema.rs defines Variable again.
So here the definition has to be this way to make the figma json response be able to convert to this Variable and later we can convert this to the toolkit defined Variable? Do I get it right?

crates/figma_import/src/toolkit_schema.rs
rylin8: I looked into this, and although it is doable, having a generic struct/enum makes serde reflection harder. In order to generate XorVar<f32>, XorVar<Color>, etc, I would need to pull in the serde-name crate, then override the serialize and deserialize functions to generate different named classes for each different generic parameter. At this point I don't think it's worth the work. Plus, this might change during your protobuf work.

timothyfroehlich: Ok, yeah, we'll look at it again with the Protobuf work. I'd been using the pattern of having individual messages (such as `DimensionRect`) and doing some simple mapping to the generic Rust native structs (`Rect<Dimension>`) to handle generics and I think that'll work here too.

designcompose/src/main/java/com/android/designcompose/VariableManager.kt
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// Return the value of a ColorOrVar enum
+internal fun ColorOrVar.getValue(variableState: VariableState): Color? {
+    return when (this) {
+        is ColorOrVar.Color -> value.toColor()
+        is ColorOrVar.Var -> VariableManager.getColor(value, variableState)
+        else -> null
+    }
+}
```
timothyfroehlich, line 292: Is there a reason why this code cannot be rewritten with generics? I notice that the color version checks the MaterialThemeValues, but MaterialThemeValues could also support generic types, even if the Google Material Theme only supports colors; it could just return null.  Since we'll also need to implement String and Boolean this seems worth doing, otherwise we'll have four copies of the same functions.

timothyfroehlich (changes requested):

timothyfroehlich:
I've finished my review. It mostly looks good, but I think deduplicating the code will make implementation of String and Boolean variables much easier, and will help prevent bugs that stem from changes to the code for only one of the variable types.

timothyfroehlich:
Also, a thought, it might be helpful to use the name "DesignVariable" (or similar) instead of "Variable", to help with discussion of the code and with documentation in general.

designcompose/src/main/java/com/android/designcompose/VariableManager.kt
```diff
@@ -0,0 +1,292 @@
...
+fun DesignMaterialThemeProvider(useMaterialTheme: Boolean = true, content: @Composable () -> Unit) =
+    CompositionLocalProvider(LocalVariableUseMaterialTheme provides useMaterialTheme) { content() }
+
+// VariableManager holds the VariableMap retrieved from the design file and has functions to find
+// the value of a variable from the variable ID.
+internal object VariableManager {
+    // Since we can have variables from multiple documents, store a variable map for each document
+    private val docVarMap: HashMap<String, VariableMap> = HashMap()
+    // A global variable map containing entries from all documents. We currently don't support
+    // duplicate variable names across multiple documents.
```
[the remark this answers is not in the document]
rylin8: https://github.com/google/automotive-design-compose/issues/1169

designcompose/src/main/java/com/android/designcompose/VariableManager.kt
```diff
@@ -0,0 +1,292 @@
...
+@Composable
+fun DesignMaterialThemeProvider(useMaterialTheme: Boolean = true, content: @Composable () -> Unit) =
+    CompositionLocalProvider(LocalVariableUseMaterialTheme provides useMaterialTheme) { content() }
+
+// VariableManager holds the VariableMap retrieved from the design file and has functions to find
+// the value of a variable from the variable ID.
+internal object VariableManager {
+    // Since we can have variables from multiple documents, store a variable map for each document
+    private val docVarMap: HashMap<String, VariableMap> = HashMap()
+    // A global variable map containing entries from all documents. We currently don't support
```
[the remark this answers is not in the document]
rylin8: We can store it that way, but looking up variable values would require passing in the Figma document ID as well. That is a larger endeavor that would require piping that value through many layers that currently don't know anything about the document ID.

designcompose/src/main/java/com/android/designcompose/VariableManager.kt
rylin8: I can make a common function to share some code between the getColor() and getNumber() functions, but I don't otherwise see much benefit in trying to use generics. There are only a few types that variables support, and while using generics might save a small bit of code duplication, I don't really think it's worth sacrificing the code readability and simplicity here.

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -54,6 +54,7 @@ import com.android.designcompose.serdegen.BlendMode
 import com.android.designcompose.serdegen.Dimension
 import com.android.designcompose.serdegen.Display
 import com.android.designcompose.serdegen.Easing
+import com.android.designcompose.serdegen.FigmaColor
```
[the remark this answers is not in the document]
rylin8: Oops, `FigmaColor` is a type used in `figma_schema.rs`. It represents the structure of a color when parsed from the Figma response. The color we use in `toolkit_schema.rs` should be a different `Color` struct and I'll fix that so that we don't output it into code that Kotlin uses.

crates/figma_import/src/toolkit_schema.rs
```diff
@@ -230,3 +279,130 @@ impl View {
...
+    pub id: String,
+    pub name: String,
+    pub remote: bool,
+    pub key: String,
+    pub variable_collection_id: String,
+    pub var_type: VariableType,
+    pub values_by_mode: VariableValueMap,
+}
+impl Variable {
+    fn create(
```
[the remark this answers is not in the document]
rylin8: done

iamralpht (approved):

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1320,7 +1355,7 @@ fn visit_node(
         NodeData::Rectangle { vector } => ViewShape::VectorRect {
             path: fill_paths,
             stroke: stroke_paths,
-            corner_radius,
+            corner_radius: corner_radius,
```
yiqunw700, line 1358: Wondering if here should use corner_radius.clone() as above?

rylin8: It's not needed here. It's needed above because of how Rust memory management works. Without the clone above, the local `corner_radius` variable will have been moved, or consumed, into the `ViewShape::RoundRect` enum, and there would be a "corner_radius has been moved" compilation error. Since nothing below this uses `corner_radius`, it doesn't need a clone.

timothyfroehlich (approved):

PR #1074, opened 2024-05-06
Migrate the Layout Nodes structs to the Layout crate

timothyfroehlich:
While working on the Protobuf code I realized that we never finished changing the LayoutNode structs that are sent across the JNI to use LayoutStyle instead of the full ViewStyle. In order for me to finish the first set of Protobuf changes I'll need to get these moved to the proper crate.


rylin8 (approved):

iamralpht:
@timothyfroehlich LP can you make a ticket for this change and put it in the milestone?

PR #1076, opened 2024-05-07
Closes #1056: Fix view finding for @DesignComponent functions on a component set.

### LINKED ISSUE #1056: showing/hiding a layer in figma causes replacement context to fail
I've created a component with two variants and an icon.  I am replacing the component on the stage with the variant component, and replacing the icon within the variant with the desired icon.  All of the replacements work as intended, until I unhide a second instance of the component on the stage.  When i do that, the working icon is always rendered in which ever state the second icon is set two in figma.  Below is a video of the issue:

https://github.com/google/automotive-design-compose/assets/64603080/b47511b1-29d8-477f-bb03-300b98f47292

Here is a zip file with the source code and figma file
[Ex4Google-variant-icon.zip](https://github.com/google/automotive-design-compose/files/15179663/Ex4Google-variant-icon.zip)

rylin8:
Historically when a @DesignComponent is declared with a component set, the component set name must match the property name in a @DesignVariant parameter in order for the view matching algorithm to use the correct variant at runtime. This has always been confusing and undocumented, so this change allows the property name to differ from the component set name. The view searching algorithm now checks at runtime of a node is actually a variant of a component set, and uses that if so. Then, in order to support TapCallback using the name of the component set, populate variantParentName if we have the component set info to do so.

The VariantProperties test has been updated to test this scenario.


yiqunw700 (approved):

iamralpht:
@kovmarci86 FYI - we will need to make the accompanying change in the Rust variant resolution implementation.

timothyfroehlich (approved):

PR #1077, opened 2024-05-08
Support fetching figma file version

yiqunw700:


crates/figma_import/src/document.rs
```diff
@@ -153,18 +157,23 @@ impl Document {
...
         proxy_config: &ProxyConfig,
         image_session: Option<ImageContextSession>,
     ) -> Result<Document, Error> {
         // Fetch the document...
-        let document_url = format!(
+        let mut document_url = format!(
             "{}{}?plugin_data=shared&geometry=paths&branch_data=true",
             BASE_FILE_URL, document_id,
         );
+        if version_id.is_empty() == false {
```
rylin8, line 169: minor: !version_id.is_empty()

crates/figma_import/src/document.rs
```diff
@@ -153,18 +157,23 @@ impl Document {
...
+            document_url.push_str("&version=");
+            document_url.push_str(&version_id);
+        }
         let document_root: FileResponse =
             serde_json::from_str(http_fetch(api_key, document_url, proxy_config)?.as_str())?;
 
-        // ...and the mapping from imageRef to URL
+        // ...and the mapping from imageRef to URL. It returns images from all versions.
```
rylin8, line 176: Is this just an old comment where the beginning got cut off?  So if this request returns all images, do we just possibly store some extra images that aren't used?

crates/figma_import/src/document.rs
```diff
@@ -193,20 +203,26 @@ impl Document {
...
+        let mut document_head_url = format!("{}{}?depth=1", BASE_FILE_URL, document_id);
+        if requested_version_id.is_empty() == false {
+            document_head_url.push_str("&version=");
+            document_head_url.push_str(&requested_version_id);
+        }
         let document_head: FileHeadResponse =
             serde_json::from_str(http_fetch(api_key, document_head_url, proxy_config)?.as_str())?;
 
-        if document_head.last_modified == last_modified && document_head.version == version {
+        if document_head.last_modified == last_modified && document_head.version == last_version {
```
rylin8, line 220: If we request a specific version of the Figma file, when would last_modified be different, or document_head.version be different than last_version? I'm a little confused here.

rylin8 (commented):
Looks good! Just a couple of minor comments. I didn't realize this would have to touch so much code, but hopefully it helped with understanding the flow of data in DesignCompose. @iamralpht do you want to rename `FigmaDocId` to something without Figma in it?

crates/figma_import/src/bin/fetch_layout.rs
```diff
@@ -74,6 +74,9 @@ struct Args {
     /// Figma Document ID to fetch and convert
     #[arg(short, long)]
     doc_id: String,
+    /// Figma Document ID to fetch and convert
+    #[arg(short, long)]
+    version_id: String,
```
rylin8, line 79: Can we make this parameter optional?

crates/figma_import/src/bin/fetch.rs
```diff
@@ -50,6 +50,9 @@ struct Args {
     /// Figma Document ID to fetch and convert
     #[arg(short, long)]
     doc_id: String,
+    /// Figma Document Version ID to fetch and convert
+    #[arg(short, long)]
+    version_id: String,
```
rylin8, line 55: This one also -- optional

yiqunw700: Done

crates/figma_import/src/document.rs
yiqunw700: Done

crates/figma_import/src/document.rs
yiqunw700: Yeah. I didn't make changes here and just added a comment.

Actually [this api](https://www.figma.com/developers/api#get-image-fills-endpoint) would return images from all branches and all versions.

crates/figma_import/src/document.rs
yiqunw700: Actually it would not. But there could be cases that no local files saved and app loads the figma doc with a version for the first time.

designcompose/src/main/java/com/android/designcompose/DocServer.kt
```diff
@@ -530,15 +537,14 @@ internal fun DocServer.doc(
...
-            "5n0LhOQ6wOiDxrH0YUVhJS" to "Media 4",
-            "dui99iAKZ273s7RN11Z9Ak" to "Media 5",
-            "2DQtQOf6U26mA8dqBie3gT" to "Media Nova",
+            FigmaDocId("7rvM6aVWe0jZCm7jhO9ITx") to "Media 1",
+            FigmaDocId("S3n4mhNgoHzNxCCHhmrVcR") to "Media 2",
+            FigmaDocId("5n0LhOQ6wOiDxrH0YUVhJS") to "Media 4",
+            FigmaDocId("dui99iAKZ273s7RN11Z9Ak") to "Media 5",
+            FigmaDocId("2DQtQOf6U26mA8dqBie3gT") to "Media Nova",
         )
     return mediaDocs
```
timothyfroehlich, line 549: This... never should have been checked in. But this should be fixed in a seperate PR.

timothyfroehlich (changes requested):
Looks good, other then the minor changes.

annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
```diff
@@ -20,11 +20,16 @@ package com.android.designcompose.annotation
  * Generate an interface that contains functions to render various nodes in a Figma document
  *
  * @param id the id of the Figma document. This can be found in the url, e.g. figma.com/file/<id>
+ * @param docVersionId version id of the Figma document.
  * @param version a version string that gets written to a generated JSON file used for the Design
```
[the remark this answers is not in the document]
rylin8: Not sure what to name it. jsonVersion?

crates/figma_import/src/document.rs
yiqunw700: Yes, that is right.

annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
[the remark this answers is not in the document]
timothyfroehlich: :/ What about figmaVersion and designDocVersion? @iamralpht and @rylin8 could you weigh in here?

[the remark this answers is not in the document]
timothyfroehlich: :/ hm, what about figmaVersion and designDocVersion? @rylin8 @iamralpht could you weigh in here?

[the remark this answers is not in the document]
rylin8: For the new version field, I was ok with `docVersionId`. I'm also ok with `designVersion` but it's quite similar to `designDocVersion`. I think @iamralpht wants to stay away from using the name Figma in the code, which is why I wanted his input on the new class used everywhere. For the old version argument, I think @timothyfroehlich's suggestion is good. So I would suggest: `docVersionId` and `designDocVersion`.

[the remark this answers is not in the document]
timothyfroehlich: > I think @iamralpht wants to stay away from using the name Figma in the code, which is why I wanted his input on the new class used everywhere.

Yeah, though my thinking was that this is a Figma specific thing. But then again, it's not really.

The problem is we don't have great names for things, we've just been calling the JSON file "the JSON file". 

I've got a new idea, what about `designSourceVersion` and `designIntegrationVersion`? Or `sourceVersion` and `integrationSpecVersion`? Like, we could name the JSON file the "Integration Spec"?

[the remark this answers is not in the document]
yiqunw700: I like the `spec` name. What about **DesignDoc(`id`, `versionId`, `integrationSpecVersion`)**?

As the new class name `FigmaDocId`, I can rename it to `DocId` or `DesignDocId`.

Let me know your thoughts. Thanks!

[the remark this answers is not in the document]
timothyfroehlich: I don't think `versionId` works, it's a version number, not an identifier. maybe just `sourceVersion` and `integrationSpecVersion`.

And yeah, `DesignDocId` works better than `FigmaDocId`.

annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
```diff
@@ -20,11 +20,16 @@ package com.android.designcompose.annotation
...
+ * @param designVersion version id of the Figma document.
+ * @param composeVersion a version string that gets written to a generated JSON file used for the
+ *   Design Compose Figma plugin
  */
 @Target(AnnotationTarget.CLASS)
-annotation class DesignDoc(val id: String, val version: String = "0")
+annotation class DesignDoc(
+    val id: String,
+    val designVersion: String = "",
+    val composeVersion: String = "0"
```
iamralpht, line 31: I might change this to `interfaceVersion`, because I think the `@DesignDoc` stuff is defining the interface between the code and the design doc.

common/src/main/java/com/android/designcompose/common/FigmaDocId.kt
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.android.designcompose.common
+
+/** When version id is not specified, it will load head of the doc. */
+data class FigmaDocId(var id: String, var versionId: String = "") {
```
iamralpht, line 20: Would it be possible to change this to `DesignDocId`? We should one day support tools other than Figma (or support hand written DCF files), and it would be good if we just added an enum or something to this type when we do that, rather than needing to rename it.

annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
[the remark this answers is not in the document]
iamralpht: We discussed live and think `customizationInterfaceVersion` rather than `composeVersion`.

annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
iamralpht: Added a comment to the discussion above.

timothyfroehlich (approved):

rylin8:
Looks like mediacompose needs to be fixed

yiqunw700:
> Looks like mediacompose needs to be fixed

Done. Could you please to take another look before I merge? @rylin8  Thanks!

rylin8 (approved):

PR #1096, opened 2024-05-11
Generate additional static case test images for rendering parity tests

johnoneil:
Generating a PR just to make this easier to discuss.


integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/RenderStaticExamples.kt
```diff
@@ -0,0 +1,91 @@
...
+ import androidx.test.core.app.ApplicationProvider
+ import com.android.designcompose.DesignSettings
+ import com.android.designcompose.DocRenderStatus
+ import com.android.designcompose.TestUtils
+ import com.android.designcompose.docClassSemanticsKey
+ import com.android.designcompose.test.assertRenderStatus
+ import com.android.designcompose.test.internal.captureRootRoboImage
+ import com.android.designcompose.test.internal.designComposeRoborazziRule
+ import com.android.designcompose.testapp.common.interFont
+ //import com.android.designcompose.testapp.validation.examples.EXAMPLES
```
johnoneil, line 33: Remove

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/StaticExamples.kt
```diff
@@ -0,0 +1,87 @@
...
+ // To run these tests in a local development env, run `./gradlew recordRoborazziDebug`.file
+ // Test outputs should be a series of images at:
+ // `integration-tests/validation/src/testDebug/roborazzi/RenderStaticExamples/*.png`
+ val STATIC_EXAMPLES: ArrayList<Triple<String, @Composable () -> Unit, String?>> =
+     arrayListOf(
+         // TODO: There's a problem with code generation so that I have to keep these tets in place or
+         // nodes can't be found for subsequent tests.
+         Triple("Layout Tests", { LayoutTests() }, LayoutTestsDoc.javaClass.name),
+         Triple("Alignment", { AlignmentTest() }, AlignmentTestDoc.javaClass.name),
+         Triple("Blend Modes", { BlendModeTest() }, BlendModeTestDoc.javaClass.name),
```
johnoneil, line 76: @timothyfroehlich Tim, This code currently runs and provides images. But only if I bring in these tests from `AllExamples.kt`.

I believe I'm not correctly hooking the tests defined in this module into the code generation system, requiring these external tests to be brought in.
Any idea what hook I'm missing here?

timothyfroehlich: Check out the commit that I just pushed. The `@DesignDoc` annotation creates a composables named after each `@DesignComponent` in the `@DesignDoc` (such as `MainFrame()`). We can use these directly in the tests, we don't need to create a separate composable that contains them.

Also, you need to fetch the Figma files before running screenshot tests. You can run `./gradlew val:fAUFF` (short for validation-app:fetchAndUpdateFigmaFiles) to fetch them all. That task is a fairly dumb one and I don't know if there's a way to filter it to only fetch some, so I'd recommend commenting out the `testRender` test in `RenderAllExamples` so you don't have to wait for all of the Validation examples to re-fetch.

johnoneil: I've updated to include your commit. Thanks Tim.
Was able to include a fair number of additional tests, but some (commented out) are still failing, sometimes at the "fAUFF" pass, sometimes in the "recordRoborazziDebug" stage.
I'll look at those next week.

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/RenderStaticExamples.kt
johnoneil: fixed.

PR #1111, opened 2024-05-14
Create a custom error type for dc_jni so that it can define it's own, separate from figma_import

### LINKED ISSUE #1103: Create a custom error type for dc_jni so that it can define it's own, separate from figma_import
`dc_jni` currently just uses the `Error` enum from figma_import. With the protobuf changes it'll need to add some extra variants to the enum, so it's best to create a specific `Error` enum.

timothyfroehlich:
Fixes #1103


crates/dc_jni/src/error.rs
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use jni::JNIEnv;
+use log::error;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("Error: {0}")]
```
rylin8, line 21: What does this "Error: {0}" mean?

crates/dc_jni/src/jni.rs
```diff
@@ -200,58 +203,66 @@ fn jni_add_nodes<'local>(
...
                         }
                     }
-                    for LayoutParentChildren { parent_layout_id, child_layout_ids } in
-                        &node_list.parent_children
-                    {
-                        manager.update_children(*parent_layout_id, child_layout_ids)
+                    Err(e) => {
+                        throw_basic_exception(
+                            &mut env,
+                            &Error::GenericError(format!("Internal JNI Error: {}", e)),
```
rylin8, line 249: Isn't this a deserialization error?

crates/dc_jni/src/jni.rs
```diff
@@ -151,15 +149,17 @@ fn layout_response_to_bytearray<'local>(
...
         Ok(bytes) => env.byte_array_from_slice(&bytes).unwrap_or_else(|_| {
-            throw_basic_exception(&mut env, "Internal JNI Error".to_string());
+            throw_basic_exception(&mut env, &Error::GenericError("Internal JNI Error".to_string()));
             JObject::null().into()
         }),
         Err(_err) => {
-            throw_basic_exception(&mut env, "Layout serialization error".to_string());
+            throw_basic_exception(
+                &mut env,
+                &Error::GenericError("Layout serialization error".to_string()),
```
rylin8, line 161: not bincode serialization error?

crates/dc_jni/src/error.rs
timothyfroehlich: GenericError's a [tuple struct](https://google.github.io/comprehensive-rust/user-defined-types/tuple-structs.html) so that's how you get at it to print it.

crates/dc_jni/src/jni.rs
timothyfroehlich: Should just use the error itself.

rylin8 (approved):

PR #1115, opened 2024-05-14
Move the LayoutManager Rust code to it's own file Fixes #1106

### LINKED ISSUE #1104: Move the layout-related code in `jni` to it's own file/module.
This simplifies some of the changes for the protobuf code, since the first phase is only converting the layout structures.

### LINKED ISSUE #1106: Move the LayoutManager Kotlin code to it's own file
https://github.com/google/automotive-design-compose/blob/2dd23fb95d07c54f22d1a297b9cc2fa021882eb4/designcompose/src/main/java/com/android/designcompose/Layout.kt

Helps to better define the split between the Compose Kotlin code and the code that interacts with the Rust lib.

timothyfroehlich:
Fixes #1104

### DIFF
--- a/crates/dc_jni/src/jni.rs
+++ b/crates/dc_jni/src/jni.rs
@@ -12,51 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashMap;
 use std::ffi::c_void;
-use std::sync::atomic::{AtomicI32, Ordering};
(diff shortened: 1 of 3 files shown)


rylin8:
The functions you changed from private to `pub` should probably be `pub(crate)`

timothyfroehlich (commented):
> The functions you changed from private to `pub` should probably be `pub(crate)` 

Makes sense

rylin8 (approved):

PR #1146, opened 2024-05-23
Add default Android Studio Settings

timothyfroehlich:
# Based on PR #1142

## Full chain of PRs as of 2024-05-24

* PR #1146: `wb/froeht/android_studio_settings` ➔ `wb/froeht/vscode-settings`
* PR #1142: `wb/froeht/vscode-settings` ➔ `main`

Please give this a shot and let me know what you think.

This **WILL** conflict with your existing files. These are the files that GitHub and IntelliJ recommend you check into version control. (See the additions to .gitignore)

I've checked each file to make sure that the settings are appropriate and enforce our style.

The formatting defined here should NOT conflict with the formatting options defined in the VS Code config that's been committed. If it does then try to figure it out and upload a fix.

This includes proper formatting rules for yaml and markdown files, so you'll see those get updated if you modify them (such as the github-pages.yml file that I used to make sure Android Studio and VSCode were formatting it the same)

As a bonus, I've also added a bit of configuration to have Android Studio automatically run the generate sources tasks (KSP and Serdegen) when the project opens. This will give you a head start.

Also, I found the setting to make Android Studio use Git's file staging by default, so now it and VS Code's source control views will work the same!

### DIFF
--- a/.idea/externalDependencies.xml
+++ b/.idea/externalDependencies.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ExternalDependencies">
+    <plugin id="com.facebook.ktfmt_idea_plugin" />
+    <plugin id="com.intellij.gradle" />
+    <plugin id="com.yoheimuta.intellij.plugin.protolint" />
+    <plugin id="idea.plugin.protoeditor" />
+    <plugin id="org.intellij.plugins.markdown" />
+    <plugin id="org.jetbrains.android" />
+    <plugin id="org.jetbrains.kotlin" />
+    <plugin id="org.jetbrains.plugins.yaml" />
+    <plugin id="org.toml.lang" />
+  </component>
+</project>
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,0 +1,6 @@
+<project version="4">
+  <component name="ExternalStorageConfigurationManager" enabled="true" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" default="true" project-jdk-name="jbr-17" project-jdk-type="JavaSDK">
+    <output url="file://$PROJECT_DIR$/build/classes" />
+  </component>
+</project>
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -13,7 +13,8 @@
         "plex.vscode-protolint", // Protobuf linting
         "timonwong.shellcheck", // Shell script linting
         "zxh404.vscode-proto3", // Needed for Protobuf linting
-        "redhat.vscode-yaml", // YAML formatting
+        "redhat.vscode-yaml", // YAML
         "github.vscode-github-actions", // Github Action formatting and linting
+        "bluebrown.yamlfmt", // YAML formatting
     ]
 }
--- a/.github/workflows/giithub-pages.yml
+++ b/.github/workflows/giithub-pages.yml
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 name: GitHub Pages
-
 on:
   push:
     paths:
@@ -26,21 +25,17 @@ on:
       - main
       - 'wb/**'
   merge_group:
-
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
-
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:
   contents: read
   pages: write
   id-token: write
-
 # Allow one concurrent deployment
 concurrency:
   group: "pages"
   cancel-in-progress: true
-
 jobs:
   # Build job
   build:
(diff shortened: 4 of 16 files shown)


timothyfroehlich:
> Do all of the files in `.idea` need to be in here? I'm not sure how they are generated, but for the most part, mine are already the same as the ones in this change, 

That's the idea, many of these will have been generated by Android Studio and will be identical. But having them checked in will make sure that they stay consistent.

>but there are some small differences. For example my `.idea/deploymentTargetSelector.xml` file contains some extra info on which emulator to use for the different targets, and I'd like to keep that.

That's exactly the conversation that I want to have, some of those files are really up to the team. Yeah, deploymentTargetSelector sounds like something that we should exclude since we each have our own sets of targets. I'll remove it. 

What else?

.idea/misc.xml
rylin8, line 1: Mine has some more:
```
<project version="4">
  <component name="ExternalStorageConfigurationManager" enabled="true" />
  <component name="ProjectRootManager" version="2" languageLevel="JDK_17" default="true" project-jdk-name="jbr-17" project-jdk-type="JavaSDK">
    <output url="file://$PROJECT_DIR$/build/classes" />
  </component>
```

README.md
```diff
@@ -124,6 +124,54 @@ compiler. Compilation will happen automatically for the packages that need it, b
...
 
+### Formatters
+
+- Yaml: `yamlfmt`
+- Kotlin: `ktfmt`
+- Rust: `rustfmt`
+- Markdown: `markdownlint-cli2`
+- Protobuf: `protolint`
+
+### Install everything at once please
```
rylin8, line 135: I'm confused what this section is. You have instructions on things like installing protobuf below, when you already have it above in the protobuf section. You also mention using brew, but my understanding is that that is usually for Mac, not linux. Above you have use `apt` to install protobuf.

.vscode/extensions.json
```diff
@@ -14,6 +14,7 @@
         "timonwong.shellcheck", // Shell script linting
         "zxh404.vscode-proto3", // Needed for Protobuf linting
         "redhat.vscode-yaml", // YAML formatting
-        "github.vscode-github-actions", // Github Action formatting and linting
+        "github.vscode-github-actions",
+        "bluebrown.yamlfmt", // Github Action formatting and linting
```
rylin8, line 18: Is the comment for this on the wrong line? I think it will conflict with the VSCode settings.

README.md
timothyfroehlich: Brew works for Linux too. This is my attempt to make it easier to install everything. What I'll do is add links for each of the formatters so that the reader can follow the instructions for installing each of them if needed.

timothyfroehlich: Moving this to the Protolint PR

.vscode/extensions.json
timothyfroehlich: That might have been a bug with VSCode's "Add to recommended extensions" command. I should file that. Fixed.

.idea/misc.xml
timothyfroehlich: Doesn't seem to cause any issues on my end, I'll include it.

.idea/externalDependencies.xml
```diff
@@ -0,0 +1,34 @@
...
+    <plugin id="com.faendir.intellij.gradle-version-catalogs-plugin" />
+    <plugin id="com.intellij.gradle" />
+    <plugin id="com.yoheimuta.intellij.plugin.protolint" />
+    <plugin id="idea.plugin.protoeditor" />
+    <plugin id="org.intellij.plugins.markdown" />
+    <plugin id="org.jetbrains.android" />
+    <plugin id="org.jetbrains.kotlin" />
+    <plugin id="org.jetbrains.plugins.github" />
+    <plugin id="org.jetbrains.plugins.yaml" />
+    <plugin id="org.toml.lang" />
```
rylin8, line 32: Ok now that I don't have conflicts, I tried this out and have a red popup in Android Studio that "Required plugins have not been loaded". In VS Code at least it was just "recommended", can we do that here? I was able to work perfectly fine without these plugins.

timothyfroehlich: Doesn't look like I can set them to "Recommended", unfortunately. I removed the ones that aren't necessary like GitHub, but I think I'll leave in the formatting ones like `toml` and `yaml`. This is so we can make sure we have our formatting correct, after all.

rylin8:
Seems ok on my end, but I don't know if this would conflict with other people's configs. I don't know what a lot of those configs do or if it's a good idea to check them in. Since nobody else has said anything I'm ok trying this out, but if there are issues please chime in.

rylin8 (approved):

timothyfroehlich:
> Seems ok on my end, but I don't know if this would conflict with other people's configs. I don't know what a lot of those configs do or if it's a good idea to check them in. Since nobody else has said anything I'm ok trying this out, but if there are issues please chime in.

Yep, we can always make adjustments later.

PR #1184, opened 2024-05-31
Closes #1173: Variable support for font size, weight, and gradient color stops

### LINKED ISSUE #1173: Add variable support for font properties
Support font size and weight.
Font family would be nice, but we currently don't use the font from Figma anyway, so we need to support that before supporting variables for font family.

rylin8:


rylin8:
Hmm, roborazzi is failing for not only the variant properties test, but for the eight variable test images that I updated in this PR.

timothyfroehlich:
> Hmm, roborazzi is failing for not only the variant properties test, but for the eight variable test images that I updated in this PR.

Logged in #1186. We'll have to just ignore it until I can implement a fix.

timothyfroehlich (approved):
LGTM

PR #1189, opened 2024-06-03
Create Proto messages for SerializedDesignDoc 

timothyfroehlich:
This is another step towards eliminating Serdegen. This should be the full set of messages required for SerializedDesignDoc.

The messages have been largely generated by Gemini, so there's likely some inaccuracies in there. Those will become evident when we start integrating the generated messages into the code, so I'm not too concerned about that.

What I would like review on right now is the layout of the files in the proto/ directory. Since we're trying to create a documented, stable format it made sense to split the messages into different packages. They are:

design: Everything needed for a DesignDoc
design/element: Basic elements of a design. Color, Rectangle, etc.
design/interaction: Animations and reactions
design/layout: Grid, alignment, etc.
design/modifier: Blends, filters, transforms
design/plugin: Figma Plugin specific, Dials/Guages data
toolkit: "Buisiness logic" types, currently SerializedDesignDoc.
android_interface: Messaging types that are only used to communicate across the JNI.


proto/design/layout_node.proto
```diff
@@ -14,31 +14,12 @@
```
rylin8, line 14: Not sure why this is not in the `layout` folder.

timothyfroehlich:
Ok, it's ready for review now. I've cleaned up the formatting, fixed some obviously bad messages, copied comments over and generated comments for the rest. I also noticed that instead of creating empty messages you can instead use the `google.protobuf.Empty` type, which seems to be a bit cleaner.

rylin8:
The description still says that this is a draft.

timothyfroehlich:
> The description still says that this is a draft.

Fixed.

timothyfroehlich:
To be clear, the goal here is to get these messages checked in on their own, to keep the PR relatively small.

crates/layout/src/proto/layout_node.rs
```diff
@@ -12,12 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::layout_node::{self};
+use dc_proto::{
```
rylin8, line 15: I'm confused because the crate `dc_proto` doesn't have much in it -- like it doesn't have any of the dependencies below in it. Can you explain somewhere how this all works?

proto/design/component.proto
```diff
@@ -0,0 +1,47 @@
...
+// Figma component properties can be "overridden" in the UI. These overrides
+// are applied in the node tree we get back from Figma, so we don't have to
+// worry about them, unless we present a variant of a component as a result of
+// an interaction. In that case, we need to figure out what has been overridden
+// and apply it to the variant. This struct keeps track of those overrides that
+// need to be applied to variant instances.
+message ComponentOverrides {
+  ViewStyle style = 1;
+  oneof component_content_override {
+    bool none = 2;
```
rylin8, line 33: Are you able to use google.protobuf.Empty here instead of bool?

proto/design/element/variable.proto
```diff
@@ -0,0 +1,94 @@
...
+  VariableValueMap values_by_mode = 7;
+}
+
+// We redeclare VariableValue instead of using the one from figma_schema because
+// the "untagged" attribute there prevents serde_reflection from being able to
+// run properly.
+message VariableValue {
+  oneof value {
+    bool boolean = 1;
+    float float = 2;
```
rylin8, line 50: I'd call this "number"

proto/design/element/variable.proto
```diff
@@ -0,0 +1,94 @@
...
+// variable given a mode.
+message VariableValueMap {
+  map<string, VariableValue> values_by_mode = 1;
+}
+
+// a reference to a design variable (e.g., color, spacing). Think of it as a
+// symbolic name for a value that can change throughout your design.
+message VariableAlias {
+  string id = 1;
+  string name = 2;
```
rylin8, line 68: I don't think a VariableAlias has a name in it. I think it's just the ID, and we have to find the name based off the ID. Though I'm not sure how you're using this, maybe you've already resolved the name.

proto/design/text_style.proto
```diff
@@ -0,0 +1,49 @@
...
+// These are the style properties that apply to text, so we can use them on
+// subsections of a longer string. We then assume that every style transition is
+// a potential line break (and also run the linebreaking algorithm on the
+// content of every style for the normal case where we need to break text that's
+// all in one style).
+message TextStyle {
+  modifier.Background text_color = 1;
+  float font_size = 2;
+  optional string font_family = 3;
+  element.FontWeight font_weight = 4;
```
rylin8, line 32: font size and weight can be variables now

proto/design/element/variable.proto
rylin8: How was this resolved?

proto/design/text_style.proto
rylin8: How was this resolved?

timothyfroehlich (commented):
> How was this resolved? 

Sorry, I misinterpeted your comment as saying that they can be floats, but I see that you've merged in the font variables now.

proto/design/element/variable.proto
rylin8: `VariableAlias` is structured the way it is because that is how it is parsed from Figma. The 'type' corresponds to the string "VARIABLE_ALIAS", such as:
```
            "boundVariables": {
              "itemSpacing": {
                "type": "VARIABLE_ALIAS",
                "id": "VariableID:2039:72"
              },
            }
```

timothyfroehlich: Wow, I'm sorry, Rob, I was not reading properly. I don't know why I looked at the VariableAlias message and thought that it had the same fields as the rust struct.

Ok, so you're saying that all we actually need is the ID. That makes sense. I also checked and BoundVariables doesn't get serialized (it's not in reflection.rs) so I shouldn't have a proto message for it anyways. I'll just remove the VariableAlias type since it's nothing more than a choice of the VariableValue enum/oneof.

proto/design/text_style.proto
timothyfroehlich: Ok. That's been fixed. I just set it to directly use a NumOrVar instead of a seperate message type, and we'll see if we want to change that when we work it into the code

proto/design/element/font.proto
```diff
@@ -0,0 +1,54 @@
...
+  // A form that is generally cursive in nature.
+  FONT_STYLE_ITALIC = 2;
+  // A typically-sloped version of the regular face.
+  FONT_STYLE_OBLIQUE = 3;
+}
+
+// The width of a font as an approximate fraction of the normal width.
+// Widths range from 0.5 to 2.0 inclusive, with 1.0 as the normal width.
+message FontStretch {
+  NumOrVar value = 1;
```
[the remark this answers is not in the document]
timothyfroehlich: Clearly I wasn't thinking straight when I was addressing the last set of review comments.

rylin8 (approved):
Ok I think you've resolved all of my initial concerns, mainly with incorrect types in the proto files. I'm not sure I fully understand the organizational structure of where the proto files belong. For example `variable.proto` is in `element`, but `variable_collection.proto` is in `design`. There are a few others I'm not sure about, like the other proto fields directly in `design` but not in a subfolder. If you want to reorganize a bit later I'm ok with that too.

timothyfroehlich:
> Ok I think you've resolved all of my initial concerns, mainly with incorrect types in the proto files. I'm not sure I fully understand the organizational structure of where the proto files belong. For example `variable.proto` is in `element`, but `variable_collection.proto` is in `design`. There are a few others I'm not sure about, like the other proto fields directly in `design` but not in a subfolder. If you want to reorganize a bit later I'm ok with that too.

Yeah, I had the things in 'design' in a directory for a bit but couldn't figure out what to call it. I'll look at variable collection again. What else?

rylin8:
Maybe put the stuff in `design` into `design/view`. Also, the fact that there are two `layout_style` protos is confusing.

PR #1191, opened 2024-06-03
Closes #1188: Fix a bug where explicit variable modes set in Figma were not honored

### LINKED ISSUE #1188: Variable modes set on Figma nodes are not always honored
When multiple nodes in the same hierarchy have variable modes set on them, only the highest level node's mode is used. Any additional modes set down in the UI tree are ignored.

rylin8:


yiqunw700 (approved):

iamralpht (approved):

rylin8:
Forgot to update the variables test serialized file -- just updated now.

Issue #275, opened 2023-08-01
Create figmaFileCommit and figmaFileFetch andCommit tasks, create developer script to update all serialized files in the repository.


timothyfroehlich:
Also update documentation to document the two new tasks.

PR #1219, opened 2024-06-12
Create a script to update dcf files for design switcher, tutorial and helloworld app

yiqunw700:


rylin8:
This looks good. One small concern is that by getting the dcf files with the `fetch` tool, we don't have the benefit of passing in the `ignoredImages` hash that DesignCompose uses with live update. This hash allows some images in the Figma file to be ignored (not downloaded) because we know that the image will be replaced with some kind of customization. For some reason, our `fetch` tool doesn't support passing in ignored images as part of the command line parameters. The result is that while the downloaded `.dcf` file will function correctly, it might have extra images in there that would cause the file to bloat in size.

I think we can live with that for now, but it would be nice to fix the `fetch` tool to support ignored images and pass them in here so that we get the same result as from live update in DesignCompose.

@timothyfroehlich it would be great to have one script/task to update all of the `.dcf` files in one go.

timothyfroehlich:
> @timothyfroehlich it would be great to have one script/task to update all of the `.dcf` files in one go.

We already have that request recorded in #275. But @iamralpht is the one who should decide if I switch to a different task.

timothyfroehlich (approved):

rylin8 (approved):

rylin8:
Should make a ticket for this and reference it

PR #1220, opened 2024-06-12
Create test util methods to clear interaction states

yiqunw700:


designcompose/src/testFixtures/java/com/android/designcompose/TestUtils.java
```diff
@@ -39,4 +43,17 @@ public static void triggerLiveUpdate() {
 
         testOnlyTriggerLiveUpdate(actualFigmaToken);
     }
+
+    public static void clearInteractionStates() {
+        InteractionStateManager.INSTANCE.getStates().clear();
```
rylin8, line 48: where does `INSTANCE` come from?

yiqunw700: yeah this is java, not kt. Would you like it converted to kt?

timothyfroehlich: When I wrote this there actually wasn't support for Kotlin test fixtures. Looks like it's coming in AGP 8.5: https://b.corp.google.com/issues/259523353?pli=1

rylin8: Oh -- didn't notice. I guess that's fine... I don't know why this file was made in java to begin with -- @timothyfroehlich ?

timothyfroehlich: I replied right before you did :)

timothyfroehlich (commented):
This looks awesome. I didn't realize something like this was so simple, I haven't dug into Junit rules that much!

rylin8 (approved):

Issue #1186, opened 2024-06-01
Consistent Roborazzi screenshot differences in the VariantMatching test


timothyfroehlich:
This has been happening since #1055. Since we couldn't find a difference in the files we decided to just merge it, but now all of our PRs are reporting differences in screenshots for the VariantMatching tests, even if the PR changes nothing that could affect it.

I've tried regenerating the screenshots in #1185 but that doesn't seem to have helped. My best guess is that there's some new small difference between our Linux environments and the GitHub runners environments. Implementing a CI scheme where the screenshots are recorded and stored by the runners, similar to what's suggested in https://github.com/takahirom/roborazzi/issues/351 may solve the problem.

Until then we might have to ignore the screenshot differences and validation test failures.

PR #1223, opened 2024-06-13
Set inter font for all validation tests

yiqunw700:
This should fix issue #1186 that github workflow generates consistent screenshot diffs for variant tests. The verifyRoborazzi check are not failing any more.

I believe this is similar to the interaction states not get cleared. On the local machine, it seems tests from different kt files are running on different processes but the github flow somewhat runs different kt test files on the same process.

Right now some of the tests have the font set and others not. So I am setting font to all tests so the screenshot are generated using the same font no matter if it shares a process with others or not.

But further more, we might want to look into why and how to isolate tests?? I haven't found anything how to do that for Robolectric though.


timothyfroehlich:
Looks good to me, but could you update the description with the reasoning for this change?

integration-tests/validation/src/testFixtures/java/com/android/designcompose/testapp/validation/InterFontTestRule.java
timothyfroehlich, line 1: Could you move this to be part of the app-common package? I think it's likely we'll re-use this rule.

timothyfroehlich (approved):
Great catch that it's the font! And yeah, it's super odd that the test process seems to be shared. Perhaps there's some environment variable that's set on the runners to try to combine the processes to make them run faster.

integration-tests/validation/src/testFixtures/java/com/android/designcompose/testapp/validation/InterFontTestRule.java
timothyfroehlich: If you can and it works then go ahead and push that change and submit.

rylin8:
Nice find! Not sure what triggered it to behave differently but seems this should be done anyway.

rylin8 (approved):

integration-tests/validation/src/testFixtures/java/com/android/designcompose/testapp/validation/InterFontTestRule.java
yiqunw700: Done. Move to app common package.

PR #1248, opened 2024-06-18
wb/froeht/migrate-styles-to-bundle

timothyfroehlich:


rylin8 (approved):
Same comment as before -- I'd prefer reducing the number of these PRs as it takes up a lot of time to look at them individually, and it's difficult to keep track of the big picture.

Issue #1213, opened 2024-06-08
Add squoosh support for component replacement


rylin8:
Component replacement in squoosh with a DesignCompose generated composable partially works, but the replaced node should take the size, position, and layout properties of the original node.

rylin8:
@iamralpht this is done correct?

PR #1249, opened 2024-06-19
Towards #1213: improve component replacement implementation.

iamralpht:
This change improves basic component replacement so that more of the component replacement tests pass, and a baseline image for component replacement with squoosh has been added.

There is no real connection in the layout trees in this implementation. The parent decides the size of the children and the children get laid out in that size. If a child needs more space, then it has no way of requesting it.

We could make it so that the parent uses the Compose measure method first, but it's not clear what we need to pass in as constraints (it could be the style-based min/max width/height if available...).


iamralpht:
Rotation isn't working here, and I don't think our test exercises opacity, blend modes, or masking...

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -692,18 +692,39 @@ fun SquooshRoot(
...
+                                    object : ComponentReplacementContext {
+                                        override val layoutModifier: Modifier = Modifier
+                                        override val textStyle: TextStyle? = null
+                                    }
+                                )
+                            } else if (child.content != null) {
+                                Log.d(TAG, "Unimplemented: child.content")
+                            }
+                        },
+                        measurePolicy = { measurables, constraints ->
```
rylin8, line 712: Hmm I don't understand how the constraints here are calculated. I tried something very similar to this and the constraints here were those calculated when, a few lines above, `child.component.invoke()` eventually calls `SquooshRoot()`, which collects the view hierarchy of the replacement component and performs a full layout on that tree. This layout didn't take into account the layout from the previous (original) view tree, so it didn't for example stretch if the original node was stretched in layout due to contraints.

Your implementation does seem to take the original node size into account though, because the test with left/right constraints works as expected.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
```diff
@@ -121,93 +121,101 @@ internal fun Modifier.squooshRender(
                             offsetFromRoot.y * density
                         )
                         childRenderSelector.selectedRenderChild = null
-                    }
+                    } else {
```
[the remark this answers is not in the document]
iamralpht: Yep, exactly.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
iamralpht: What's happening here (I think) is:

- The outer `SquooshRoot` runs layout and generates `Constraints` which are passed to the child composables at L661 of SquooshRoot.kt. This logic is a bit dubious because it never consults the child measurables on their ideal size -- so if a replaced thing is bigger/smaller than the space the parent allocates for it then it will be ignored and won't fit.
- The outer `SquooshRoot` used to create a Compose `Box` object here, and that ate the layout constraints. Now it's creating this `Layout` thing which just passes the Constraints directly through to the child.
- If the child is also a DesignCompose generated Composable, then its `SquooshRoot` will receive the `Constraints` generated at L661 at L502 (and apply them to its root at L505).

So all of this stuff is just glue to get the constraints generated at L661 passed to children at L505.

To make the list-based content replacement work, I think I need to fix the issue where we don't consult child measurables on the space they need (otherwise we won't make enough space for the list items in the design switcher, for example). I'm looking at implementing the intrinsic size accessors of the layout MeasurePolicy for SquooshRoot (and also making a synthetic layout entry for every child item).

rylin8: Ok, that makes sense, though I don't know why when I attempted this, the constraints weren't passed to children at L505. There's a lot less code here than what I had which is great! I have an implementation of fixing rotations and rotation animations in review now, but I don't think component replacement with a rotated node will work still because that transformation is lost. We'd have to do something like what the non-squoosh branch does, where we copy the original node's external layout data to the replacement node.

rylin8: https://github.com/google/automotive-design-compose/pull/1263

rylin8 (approved):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
iamralpht: Ok, filed #1279 to track the rotation issue.

I agree we will probably need something else to pass through this information: some stuff gets propagated via the graphics context, but if the embedded Composable uses a GraphicsLayer then I suspect that info gets lost, too...

PR #1259, opened 2024-06-20
Fixes #1251: Move Figma/Live Update related messages to a proto live_update package

### LINKED ISSUE #1251: Move Figma/Live Update related proto messages to a live_update package.
Keep them separated from the core Serialized Design Doc / Design Definition. This will help us keep live_update and Figma-specific code out of the core design definition.

timothyfroehlich:


proto/live_update/figma/figma_node.proto
```diff
@@ -0,0 +1,37 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package com.android.designcompose.proto.live_update.figma;
```
rylin8, line 17: I'm a little confused here, is this file just for live update or is this related to the NodeQuery that we use a bunch in the Kotlin code?

timothyfroehlich: I assumed NodeQuery was part of live update. Is it still used even when we don't have live update enabled?

rylin8: Yea, if you look at the generated kotlin code for a designcompose function, it creates a nodequery of the root frame name. That's how we look for views.

timothyfroehlich: Ok, so it's not part of the design definition, and it doesn't look like it's sent across the JNI, right? So it's just a struct that we use in both the kotlin and rust code?

rylin8: Yea it's part of the design definition:
```
pub struct SerializedDesignDoc {
    pub last_modified: String,
    pub views: HashMap<NodeQuery, toolkit_schema::View>,
    ...
}
```

timothyfroehlich: Ah, that's the issue, the message for definition is incorrect. 
```
message SerializedDesignDoc {
  string last_modified = 1;
  map<string, definition.view.View> views = 2;
```

Fixing!

timothyfroehlich: .... And it's not that simple. I can't use a message as a key for a proto map. I'll file a separate ticket for this.

rylin8 (approved):

PR #1263, opened 2024-06-20
Closes #917: Support rotations in squoosh animations

### LINKED ISSUE #917: Squoosh: animation: support rotation
Currently the Squoosh animation implementation doesn't support rotation, because matrix decomposition hasn't been implemented in our Kotlin code (and rotation is encoded into the transformation matrix). We should implement support for animations that rotate an object between variants.

rylin8:
First, this change fixes rendering of rotated nodes in Squoosh by rendering with the original node size instead of the layout-calculated node size. To prevent animation regressions for nodes whose size changes, a new 'overrideLayoutSize' field has been added to SquooshResolvedNode. When this field is set to true, which is during a layout animation, the renderer always uses the layout sizes computed by the animation code. When this field is set to false, the renderer tries to use the original node size, if it was fixed.

To support rotation animations, we simply interpolate the transform matrix.


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -74,11 +93,30 @@ internal class SquooshAnimatedLayout(
...
+            for (i in 0..15) {
+                transform[i] = toTransform[i] * value + fromTransform[i] * iv
+            }
+            target.style = target.style.withNodeStyle { s -> s.transform = Optional.of(transform) }
+        }
+
+        // When doing a layout animation, set overrideLayoutSize to true so that the rendering code
+        // uses this computed size as opposed to the size of the node in ViewStyle. This ensures
+        // that size change animations and rotation animations render at the correct size.
+        target.overrideLayoutSize = true
```
iamralpht, line 114: Wouldn't this make a discontinuity at the start of an animation where we need to use the style size (instead of the layout size)?

I really dislike how Figma's rotation implementation causes so much complexity in our runtime, so I am wondering if there's another way of handling it during doc conversion so that the runtime doesn't have to deal with it (or if there's a different way to express it in the runtime?). We shouldn't hold up this change on a new design for rotation, tho. 

PTAL at any discontinuous cases, we should fix those because they'll definitely come up in partner designs.

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -753,6 +753,21 @@ internal fun ViewStyle.withExternalLayoutData(data: ExternalLayoutData): ViewSty
     return overrideStyle.build()
 }
 
+// Get the raw width in a view style from the width property if it is a fixed size, or from the
+// node_size property if not.
+internal fun ViewStyle.fixedWidth(density: Float): Float {
```
iamralpht, line 758: These two methods seem a bit weird in sometimes considering density and sometimes not.

I'm not sure where node_style comes from, but if it is out of the DCF/DCB then it doesn't know about density, right?

Can you either move these back into the calculation for rotation of a dial, or into a file where we put the other rotation tricks?

iamralpht (changes requested):
Overall looks really good (and super nice to see rotation get animated!!), just minor stuff:

- let's try to isolate the rotation layout stuff a bit, in case we can think of a better design in the future (or can move some of the complexity to the conversion phase?)
- please do matrix decomposition and recomposition, and test with skew matrices, too. 
- We should add an animation midpoints screenshot test for rotation (can be in a future change -- I think I should redo the animation test page so that all of the tests are variants, and then I can generate a big midpoints test without writing a lot of code)

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
rylin8: There are (at least) two scenarios to consider here:
1. Without rotations: without any rotations, the code here uses the computed layout sizes. Setting `overrideLayoutSize` to true forces the renderer to use this size instead of the Figma node size, which is what it was doing before anyway because you had changed the `rectSize` parameter to always use the layout size.
2. With rotations: Instead of the layout size, we use the raw size from Figma. This means that if layout changed that size, then yes there would be a discontinuity. However I don't think we previously supported rotated nodes with contraints that would stretch or shrink them, and we still don't, so I thought that would be ok here.

designcompose/src/main/java/com/android/designcompose/Utils.kt
rylin8: That's a good point, I copied the code in these functions so it might be an existing bug. I should multiply the node_size by density also. And yes node_size comes from the dcf file, and originally came from Figma as the raw size of the node.

rylin8: I got the code from `FrameRender.kt` and wanted a common function for it. It didn't make sense to keep it in FrameRender since the animation code was using it. I guess I could put it in `MathUtils.kt` but it doesn't seem like a math function -- it just retrieves data from `ViewStyle` and most other functions that do that are in `Utils.kt`.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
iamralpht: Oh, right, I had forgotten the "rotated with constraints layout" case :(. We don't support it, and Figma is doing so much UI stuff it's a little hard to figure out exactly what's happening behind the scenes (or how to make another layout engine support the same concept -- CSS does rotation strictly as a non-layout changing render effect, which is conceptually a lot easier to understand and implement, but perhaps a bit harder for content authoring).

Makes sense to keep it as you have it...

designcompose/src/main/java/com/android/designcompose/Utils.kt
iamralpht: Ok - I guess I was thinking if we could find a way to do without knowing the pre-determined size then it would make sense to hide these methods so we don't use them in more places. I'm not sure how to get there given the other complexities in rotation, so we may as well have one implementation and make sure it's fully tested etc :)

iamralpht:
Ok, after the discussion, I think I'm OK with the change, but I do think we should add a midpoints animation test, because the code around node size management has gotten regressed in the past (luckily only affecting animation, before we added an API to enable experimental animation).

rylin8:
> Ok, after the discussion, I think I'm OK with the change, but I do think we should add a midpoints animation test, because the code around node size management has gotten regressed in the past (luckily only affecting animation, before we added an API to enable experimental animation).

Ok, I'm in the process of adding this test and a skew test, and figuring out how to decompose skew + reconstruct the matrix.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -74,11 +93,30 @@ internal class SquooshAnimatedLayout(
...
+        val fromHeight =
+            if (hasRotation) from.view.style.fixedHeight(density) else fromLayout.height
+        val toWidth = if (hasRotation) to.view.style.fixedWidth(density) else toLayout.width
+        val toHeight = if (hasRotation) to.view.style.fixedHeight(density) else toLayout.height
+
+        if (hasRotation) {
+            // Interpolate the transformation matrix
+            val transform: MutableList<Float> = MutableList(16) { 0F }
+            for (i in 0..15) {
+                transform[i] = toTransform[i] * value + fromTransform[i] * iv
```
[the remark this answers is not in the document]
rylin8: I added some code to extract the skew values out of the transform matrices and they look correct. However, if I take the rotation, scale, translate and skew values and construct a matrix out of it with the formula:
matrix = translation * rotation * scale * skew, the resulting matrix seems to be missing some information from the original, so that the skew animation pops the skewed rectangle to be taller in the Y axis during the animation. It's possible there's a bug in the decomposition function, or I have a bug in reconstruction of the matrix. I don't know matrices enough to know what might go wrong with a simple interpolation of all the values, but in my tests so far that is what looks the best.

integration-tests/validation/src/testDebug/roborazzi/AnimationMidpoints/AllSameNameAnimation-MidPoint.png
rylin8, line 1: I've broken something here. The midpoint animation test images all have some element that has doubled in size. Here it's the background.

integration-tests/validation/src/testDebug/roborazzi/AnimationMidpoints/RotationAnimation-MidPoint.png
rylin8, line 1: Here both squares have doubled in size.

integration-tests/validation/src/testDebug/roborazzi/AnimationMidpoints/VariantAnimation-MidPoint.png
rylin8, line 1: Here the square at the top left has doubled in size.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -74,11 +98,52 @@ internal class SquooshAnimatedLayout(
...
+            target.skewXY = toDecomposed.skewXY * value + fromDecomposed.skewXY * iv
+            target.skewYX = toDecomposed.skewYX * value + fromDecomposed.skewYX * iv
+            this.target.style = this.target.style.withNodeStyle { s -> s.transform = Optional.of(target.toMatrix().toFloatList()) }
+            println("### From ${fromTransform}")
+            println("###   Angle ${fromDecomposed.angle} scaleX ${fromDecomposed.scaleX} scaleY ${fromDecomposed.scaleY}")
+            println("### To ${toTransform}")
+            println("###   Angle ${toDecomposed.angle} scaleX ${toDecomposed.scaleX} scaleY ${toDecomposed.scaleY}")
+            println("### Skew ${this.target.style.node_style.transform.get()}")
+
+             */
```
rylin8, line 127: Haven't been able to get the matrix reconstruction with skew to work.

integration-tests/validation/src/testDebug/roborazzi/AnimationMidpoints/AllSameNameAnimation-MidPoint.png
rylin8: Looks like the images generated by tests here don't have the scaling problem that mine have for some reason, so perhaps something with my machine?

rylin8: Fixed an issue with density.

integration-tests/validation/src/testDebug/roborazzi/AnimationMidpoints/VariantAnimation-MidPoint.png
rylin8: Fixed the issue with density

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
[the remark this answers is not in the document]
rylin8: @iamralpht let me know if you have more thoughts on interpolating the transforms vs decomposition/recomposition. Then I'll clean up the code a bit.

rylin8:
This fixes https://github.com/google/automotive-design-compose/issues/1279 also.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
iamralpht: It looks like WebKit doesn't... but I'm not sure if the `decompose2`/`recompose2` methods handle skew or not. I think since the skew terms are included in `m11`, `m12`, `m21`, `m22` that skew is incorporated, but this blending function doesn't include those terms. 

* [recompose2](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/platform/graphics/transforms/TransformationMatrix.cpp#L1804)
* [blend2](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/platform/graphics/transforms/TransformationMatrix.cpp#L1804)

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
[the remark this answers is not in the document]
iamralpht: Ok, do you want to put this in as-is and make a ticket to fix the decomposition? I linked some other examples in the other comment, so perhaps there's something there you didn't already try, otherwise the examples look good.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
rylin8: Oh thanks, this is the code I was looking for earlier but couldn't find. Let me try it.

rylin8: Updated to use decomposition, works now 👍

integration-tests/validation/src/testDebug/roborazzi/AnimationMidpoints/RotationAnimation-MidPoint.png
rylin8: Fixed density issue.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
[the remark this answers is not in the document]
rylin8: Fixed decomposition.

iamralpht (approved):

PR #1281, opened 2024-06-21
Add description and charlimit support in the localization plugin

yiqunw700:


timothyfroehlich:
I'm not sure I can be useful on this review, I haven't ever worked with the plugin.

yiqunw700:
Rebased and ready for review now.

support-figma/extended-layout-plugin/src/localization-module.ts
```diff
@@ -199,6 +207,17 @@ function saveToPluginData(node: TextNode, stringResName: string) {
   );
 }
 
+function saveExtras(node: TextNode, extras?: StringResourceExtras) {
+  if (extras) {
+    node.setPluginData(
```
rylin8, line 212: Can you remind me what the reason is that the other data is saved in shared plugin data, but this one is in private data?

integration-tests/validation/src/main/res/values/gen_strings.xml
```diff
@@ -1,15 +1,42 @@
...
-  <string name="label_goodbye_1" translatable="true">Goodbye</string>
-  <string name="label_parked" translatable="true">Parked</string>
-  <string name="label_reverse" translatable="true">Reverse</string>
-  <string name="label_charging" translatable="true">Charging</string>
-  <string name="label_driving" translatable="true">Driving</string>
-  <string name="label_regen" translatable="true">Regen</string>
-  <string-array name="label_styled_text" translatable="true">
+  <string name="label_hello"
+    description="Hello with a comma"
+    charlimit="20"
```
rylin8, line 6: Are `description` and `charlimit` official keys, or are they just made up but Android is ok with it?

support-figma/extended-layout-plugin/src/localization-module.ts
```diff
@@ -314,3 +334,18 @@ async function clearNodeAsync(node: SceneNode) {
     }
   }
 }
+
+////////////////// Extras //////////////////
+function estimateCharlimit(text: string): number | "NONE" {
+  if (text.length > 40) return "NONE";
+  return Math.ceil(text.length * 0.3) * 5;
```
rylin8, line 341: How did you come up with this formula?

support-figma/extended-layout-plugin/src/ui.html
```diff
@@ -759,63 +798,141 @@
...
+      td.appendChild(iconButton);
+
+      const form = document.createElement("form");
+      form.style.display="none";
+
+      const charlimitInput = document.createElement("input");
+      charlimitInput.type = "number";
+      charlimitInput.min = "0";
+      charlimitInput.max = "200";
+      charlimitInput.step = "5";
```
rylin8, line 843: I don't know much about Android localization, but is there any reason this is 5? What if I want to set the limit to something not divisible by 5?

support-figma/extended-layout-plugin/src/ui.html
```diff
@@ -759,63 +798,141 @@
...
+      charlimitInput.name = "charlimit";
+      charlimitInput.placeholder = "Character Limit";
+      charlimitInput.value = extras["charlimit"] == "NONE" ? 0 : extras["charlimit"];
```
rylin8, line 846: Do you think we should have a warning of some sort if the char limit is manually set to something shorter than the text?

support-figma/extended-layout-plugin/src/ui.html
```diff
@@ -759,63 +798,141 @@
...
+          }}, '*');
+      });
+      iconButton.addEventListener("click", () => {
+        form.style.display = "block";
+        iconButton.style.display = "none";
+      });
+    }
+    return td;
+  }
+
```
rylin8, line 890: I think this is ok for now, but it would be really nice to have some design support for this UI. @iamralpht, Babak could you help get some? I think the "Extras" column is a little clunky with the way it displays values as JSON and lets you edit the values in a text field. I personally probably would just have a char limit column with a text field.

support-figma/extended-layout-plugin/src/localization-module.ts
yiqunw700: This one is not used by the DesignCompose, only used by the plugin. It does not have to be in the shared plugin data so when we fetch the file it can be faster/save some storage too.

integration-tests/validation/src/main/res/values/gen_strings.xml
yiqunw700: Kind of made up but Android is ok with it. It compiles without error. And there are some other apps using description the same way: https://source.corp.google.com/piper///depot/google3/java/com/google/android/gmscore/integ/modules/tapandpay/res/res/values/strings.xml;l=17?q=strings.xml%20description

Using `description` and `charlimit` will be easier for me to parse the input xml file.

support-figma/extended-layout-plugin/src/localization-module.ts
yiqunw700: go/android-localizable-strings#design-the-ui-layout-to-accommodate-longer-translations So I put estimate 1.5 of original length here.

support-figma/extended-layout-plugin/src/ui.html
yiqunw700: Per discussion offline. fine with 5s or 10s.

support-figma/extended-layout-plugin/src/ui.html
```diff
@@ -676,21 +689,42 @@
...
       }
 
-      createInputStringTable("inputStrings", inputStringData);
+      createInputStringTable("inputStrings", inputStringArray);
     })
     reader.readAsText(selectedFile);
     showOutputsButton.textContent = "Next";
   }
 
+  function getDescriptionExtras(strElement, translatable) {
```
rylin8, line 703: this `translatable` parameter isn't used.

support-figma/extended-layout-plugin/src/ui.html
yiqunw700: Done.
I changed the input range from 0 to 200 to a calculated range: around textLength to 5*textLength.

support-figma/extended-layout-plugin/src/ui.html
yiqunw700: Done.

support-figma/extended-layout-plugin/src/ui.html
yiqunw700: Filed ticket #1295 to track this.

I updated the description to align with what currently google uses: `description [CHAR_LIMIT=number|NONE]`. Keep it in one column but it now only shows the input boxes.
Screenshot 2024-06-26 at 10 32 48

PR #1298, opened 2024-06-26
Closes #1139: Add support for a customizations that use a State<Boolean>

### LINKED ISSUE #1139: Add support for Design Customizations that use a Composable that returns a boolean.
Will require Codegen changes.

rylin8:
This allows nodes that change visibility to recompose or redraw without requiring the whole screen to recompose also.


rylin8:
While looking at this, I realized that squoosh does not support the `@Composable () -> Text`, `@Composable () -> Meter`, and now the `@Composable () -> Boolean` customizations. Any thoughts on how to implement that? We could recurse down the tree in the `SquootRoot()` composable function and call all the composable customization functions, then set some data and pass that into `resolveVariantsRecursively()`. However, wouldn't that recompose everything every time?

iamralpht:
Hm, I didn't think about the Squoosh case. I wonder if we can retire all of the `@Composable` variations and provide `State<>` versions instead. `State<>` is nice because you can read it from some Composable, and then when the state is changed, that Composable is marked dirty for the next recomposition. So it allows similar optimizations in scoping of invalidations with a bit more flexibility (e.g.: we can read `State<>` in a render function).

You can update `State<>` in `LaunchedEffect`s or in other coroutines.

rylin8:
> Hm, I didn't think about the Squoosh case. I wonder if we can retire all of the `@Composable` variations and provide `State<>` versions instead. `State<>` is nice because you can read it from some Composable, and then when the state is changed, that Composable is marked dirty for the next recomposition. So it allows similar optimizations in scoping of invalidations with a bit more flexibility (e.g.: we can read `State<>` in a render function).
> 
> You can update `State<>` in `LaunchedEffect`s or in other coroutines.

Even if we changed to State<>, let's say I have a State<Boolean> to control visibility. Every time this changed values, wouldn't entire `SquooshRoot` tree be recomposed every time?

iamralpht:
> Even if we changed to State<>, let's say I have a State<Boolean> to control visibility. Every time this changed values, wouldn't entire `SquooshRoot` tree be recomposed every time?

Yes, if we wanted to change the layout tree because we only edit that during recomposition (we could make it so some values are consulted during layout, then we'd just redo layout and not recomposition).

No, if it was just a rendering thing (in which case it would just redo the rendering).

Most Squoosh property changes will result in the recomposition of everything because it centralizes all of the invalidations, so I think for Squoosh State allows a bit more efficiency than Composables would.

rylin8:
Interesting, I thought access the value out of a State<> needed to be done in a composable function. That's quite helpful -- I updated to use State<Boolean> instead. Created https://github.com/google/automotive-design-compose/issues/1301 to take care of the other customizations.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshTreeBuilder.kt
```diff
@@ -142,6 +143,7 @@ internal fun resolveVariantsRecursively(
     useLocalStringRes: Boolean?,
 ): SquooshResolvedNode? {
     if (!customizations.getVisible(v.name)) return null
+    customizations.getVisibleState(v.name)?.let { if (!it.value) return null }
```
iamralpht, line 146: This will cause a full recomposition, but that's probably OK until we have performance data showing otherwise.

For text, we could do something clever and allow text changes to only invalidate rendering if we know they can't change layout (ie the text node size doesn't change based on text content).

iamralpht (approved):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshTreeBuilder.kt
rylin8: Yea that makes sense. For visibility we would need to determine if changing visibility changes the layout due to its parent being autolayout, but I wanted to keep this change simple.

PR #1320, opened 2024-06-29
Read json customization to exclude text customizations

yiqunw700:
- Move design spec interfaces into a single module which is now shared by localization plugin and keywords plugin.
- Visit the reaction chain to find all the text node customizations.

### DIFF
--- a/support-figma/extended-layout-plugin/src/design-spec-module.ts
+++ b/support-figma/extended-layout-plugin/src/design-spec-module.ts
@@ -1,0 +1,309 @@
+/**
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as Utils from "./utils";
+
+/** See: com.android.designcompose.codegen.CustomizationType */
+export enum DesignCustomizationKind {
+  Text = "Text",
+  TextFunction = "TextFunction",
+  TextStyle = "TextStyle",
+  Image = "Image",
+  ImageWithContext = "ImageWithContext",
+  Visibility = "Visibility",
+  VariantProperty = "VariantProperty",
+  ComponentReplacement = "ComponentReplacement",
+  ContentReplacement = "ContentReplacement",
+  Modifier = "Modifier",
+  // Not covering "placeholder"; we should remove it.
+}
+
+export interface DesignCustomization {
+  kind: DesignCustomizationKind;
+  name: string;
+  node: string;
+}
+
+export interface DesignCustomizationVariantProperty
+  extends DesignCustomization {
+  kind: DesignCustomizationKind.VariantProperty;
+  values: string[];
+}
+
+export interface DesignCustomizationContentReplacement
+  extends DesignCustomization {
+  kind: DesignCustomizationKind.ContentReplacement;
+  content: string[]; // XXX: this won't work for children that have variants.
+}
+
+export interface DesignComponentSpec {
+  name: string;
+  node: string;
+  isRoot?: boolean; // is this the root component? We must know this since we pass customizations down to children.
+  customizations: DesignCustomization[];
+}
+
+export interface DesignDocSpec {
+  name: string;
+  version: string;
+  components: DesignComponentSpec[];
+}
+
+type TopLevelComponentMaps = {
+  topLevelComponents: Map<string, BaseNode[]>;
+  topLevelComponentDefns: Map<string, DesignComponentSpec>;
+  topLevelReactionComponents: Map<string, BaseNode[]>;
+};
+
+export function initTopLevelComponentMaps(
+  json: DesignDocSpec
+): TopLevelComponentMaps {
+  // Recurse over all nodes in the doc, from the root down. We populate this
+  // `topLevelComponents` map to identify duplicates and identify missing
+  // top level components (a top-level component can appear anywhere in a doc;
+  // it is "top-level" because of how it is used in the code).
+  let topLevelComponents: Map<string, BaseNode[]> = new Map();
+  let topLevelComponentDefns: Map<string, DesignComponentSpec> = new Map();
+  // A map of the top level component node name to the aggregation of the reaction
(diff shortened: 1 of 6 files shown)


support-figma/extended-layout-plugin/src/design-spec-module.ts
```diff
@@ -0,0 +1,292 @@
...
+      }
+    }
+  }
+
+  var reactions = null;
+  let maybeReactions = node as ReactionMixin;
+  if (maybeReactions.reactions) {
+    reactions = maybeReactions.reactions;
+  }
+  if (reactions != null && reactions.length > 0) {
```
rylin8, line 203: Can you add some comments to describe what is going on here, and in the handleReactionNodAsync function above? Are you following all reaction types (navigate, change, overlay, etc) and adding their destination nodes to the list of nodes to look for text customizations?

support-figma/extended-layout-plugin/src/design-spec-module.ts
```diff
@@ -0,0 +1,292 @@
...
+    for (const customization of insideComponentDefn.customizations) {
+      // We should also look for nodes that start with "#" but don't match a customization
+      // or a top-level component name.
+      if (node.name != customization.node) continue;
+      if (
+        customization.kind == DesignCustomizationKind.Text ||
+        customization.kind == DesignCustomizationKind.TextFunction ||
+        customization.kind == DesignCustomizationKind.ComponentReplacement
+      ) {
+        if (node.type == "TEXT") {
```
rylin8, line 266: Why not TextFunction?

support-figma/extended-layout-plugin/src/design-spec-module.ts
```diff
@@ -0,0 +1,292 @@
...
+  } else if (insideComponentDefn) {
+    // Check to see if our name is one of the customizations, and if so, save the node if it is a text node.
+    for (const customization of insideComponentDefn.customizations) {
+      // We should also look for nodes that start with "#" but don't match a customization
+      // or a top-level component name.
+      if (node.name != customization.node) continue;
+      if (
+        customization.kind == DesignCustomizationKind.Text ||
+        customization.kind == DesignCustomizationKind.TextFunction ||
+        customization.kind == DesignCustomizationKind.ComponentReplacement
```
rylin8, line 264: Why ComponentReplacement?

support-figma/extended-layout-plugin/src/ui.html
```diff
@@ -255,36 +255,46 @@
...
+        <label for="uploadStringsXmlOption">Upload an existing gen_strings.xml that is generated by this plugin?</label>
+      </div>
+      <div id="stringInputPanel" style="margin-left: 18px;">
+        <div class="container" style="margin-top: 8px; margin-bottom: 8px;">
+          <button class="styled-button">
+            <label for="stringsXmlInputFile" style="width: inherit; height: inherit; line-height: 36px;">Browse</label>
+            <input type="file" id="stringsXmlInputFile" name="strings_xml_input_file" accept=".xml" class="hidden" />
+          </button>
+        </div>
+        <div id="inputStringsTableContainer" class="table-container" style="display: none; max-height: 320px;">
```
rylin8, line 284: Is this the embedded table inside the UI? It looks a little odd to me to have a table with horizontal and vertical scrollbars inside the main UI, which also has a vertical scrollbar. Is it possible to not have the inner table and make the main UI resizeable?

support-figma/extended-layout-plugin/src/design-spec-module.ts
```diff
@@ -0,0 +1,292 @@
...
+  TextStyle = "TextStyle",
+  Image = "Image",
+  ImageWithContext = "ImageWithContext",
+  Visibility = "Visibility",
+  VariantProperty = "VariantProperty",
+  ComponentReplacement = "ComponentReplacement",
+  ContentReplacement = "ContentReplacement",
+  Modifier = "Modifier",
+  // Not covering "placeholder"; we should remove it.
+}
```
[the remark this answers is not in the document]
yiqunw700: The snake case doesn't work for the clippy check of mismatched type. The mismatch warnings now show up when I use the camel case here.

support-figma/extended-layout-plugin/src/design-spec-module.ts
yiqunw700: This is the figma node type, it doesn't have a TextFunction type..

support-figma/extended-layout-plugin/src/design-spec-module.ts
yiqunw700: When I test with Media, there is a customization for the search field which is a ComponentReplacement. See here: https://github.com/google/automotive-design-compose/blob/031ee2816bc5cece5ab721176b15e9436dcfdcc6/reference-apps/aaos-unbundled/mediacompose/src/main/java/com/android/designcompose/reference/mediacompose/Modules.kt#L217

support-figma/extended-layout-plugin/src/localization-module.ts
```diff
@@ -356,3 +381,41 @@ function getSavedExtras(node: TextNode): StringResourceExtras {
...
+    let customizedNodeArray: BaseNode[] = [];
+    await DesignSpecs.findTextCustomizationNodesAsync(
+      idToDesignComponentSpecMap,
+      figma.root,
+      customizedNodeArray
+    );
+    return customizedNodeArray;
+  } else {
+    figma.notify(
+      'Keywords document not found or corrupted. Run the "Check keyword definitions" plugin to upload a new keyword document.'
```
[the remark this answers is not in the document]
yiqunw700: Done.

support-figma/extended-layout-plugin/src/ui.html
yiqunw700: I removed height limit of the table so it won't have a vertical scrollbar. But I don't know how to make the main UI resizable that it seems figma's api ShowUIOptions doesn't have a way to make it resizable.

support-figma/extended-layout-plugin/src/design-spec-module.ts
yiqunw700: Done.

support-figma/extended-layout-plugin/src/ui.html
```diff
@@ -255,36 +255,46 @@
...
+        </div>
+      </div>
     </div>
+  
+    <label id="showOutputsButton"
+      style="margin-top: 16px; position: fixed; bottom: 20px; right: 20px; text-decoration: underline; color: blue; line-height: 36px;"
+      onclick="showOutputStrings()"> Next >> </label>
```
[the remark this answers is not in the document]
yiqunw700: I broke it into two pages. The first page only has the exclude options and the second page remain the same where it has a "Skip" button and can change to "Next" when it uploads a string xml file.

rylin8:
Not really related to this change, but the "Submit" button is a little confusing as to whether it applies to the "Char Limit" and "Description", or just "Description".

rylin8 (approved):

PR #1322, opened 2024-07-02
Closes #1321: Support custom variant animations in code

### LINKED ISSUE #1321: Support custom variant animations in code
Animations between code-driven variant changes are currently hardcoded to one type of transition. Support custom transitions by supplying a function that takes the transition info and returns a transition.

rylin8:
Add a customVariantTransition parameter to DesignDocSettings that can be populated to customize variant transitions.


rylin8:
Ping

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariantAnimationTest.kt
```diff
@@ -35,6 +35,10 @@ import com.android.designcompose.annotation.Design
 import com.android.designcompose.annotation.DesignComponent
 import com.android.designcompose.annotation.DesignDoc
 import com.android.designcompose.annotation.DesignVariant
+import com.android.designcompose.serdegen.Bezier
```
iamralpht, line 38: I'm not sure if we should expose these types directly (partly because we're going to change to proto serialization, partly because they're currently clunky because serdegen makes ugly interfaces).

I know Compose has "AnimationSpec" which we could use instead of the serdegen curve definitions. I think we'd have to define a Kotlin type for whether it's smart animate or something else. Then if we wanted to do a very custom thing with different curves for different elements, or additional property changes that are only in the transition (like an extra rotation or blink or something), there are types to extend to do that.

Does that make sense, or it's better to use just the DCF types? Otherwise I like this change because it's pretty simple but can give a lot of options.

rylin8: Does "SmartAnimate" vs the other types of animation actually animate differently? It looked to me like they all did the same thing.

rylin8: Ok I updated so that the custom animation function returns a `AnimationTransition` interface that lets you access an `AnimationSpec<Float>`. Right now there is a `SmartAnimationTransition` that implements the interface. We could add more later but not sure that's the approach you were thinking of... and adding more transition implementations would require a little more plumbing.

designcompose/src/main/java/com/android/designcompose/InteractionState.kt
```diff
@@ -435,7 +437,9 @@ internal fun InteractionState.dispatch(
                                     key,
                                     action.destination_id.get(),
                                     undoInstanceId,
-                                    action.transition.get(),
+                                    SmartAnimateTransition(
```
iamralpht, line 440: Oh - this is a nice way to convey the kind of transition the animation system should do. Thanks for aligning both paths here!

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshVariantTransition.kt
```diff
@@ -212,6 +260,12 @@ internal class SquooshVariantTransition {
             // This will wipe out any existing transition; we probably want to make a smooth
             // function here, like we do for the interaction-based transitions.
             if (toId != fromId) {
+                val transition =
+                    customVariantTransition?.let { it(VariantTransitionContext(view, variantView)) }
+                        ?: SmartAnimateTransition(
```
iamralpht, line 265: minor nit: do you want to pull this out as a const somewhere?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariantAnimationTest.kt
```diff
@@ -68,7 +74,38 @@ fun VariantAnimationTest() {
         }
     }
 
-    CompositionLocalProvider(LocalDesignDocSettings provides DesignDocSettings(useSquoosh = true)) {
+    CompositionLocalProvider(
+        LocalDesignDocSettings provides
+            DesignDocSettings(
+                useSquoosh = true,
+                customVariantTransition = { context ->
+                    if (context.fromComponentSet("RightPanelAndroid")) {
```
iamralpht, line 82: Eventually we'll have some tooling to have the transitions come from the DCF file (and presumably from Figma eventually too, to close the loop), so this might be a short term API... I think this will be the first time we've exposed node ids directly to user code, and I'm wondering if we should have a warning annotation on these methods, or wrap the strings in a "UnstableNodeId" type or something like that so that it's clear what's going on, and that any design change can break the animation.

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/AnimationMidpoints.kt
```diff
@@ -76,6 +79,41 @@ class AnimationMidpoints {
...
+        val text = mutableStateOf("X")
+
+        composeTestRule.setContent {
+            CompositionLocalProvider(
+                LocalDesignDocSettings provides
+                    DesignDocSettings(
+                        useSquoosh = true,
+                        customVariantTransition = {
+                            SmartAnimateTransition(
+                                tween(
```
iamralpht, line 97: For some reason the midpoint snapshot looks like it hasn't moved from the start of the animation. Does it show something more mid-pointy if you change the curve?

iamralpht:
Minor feedback - the indication that node ids are unstable, and the midpoints tests are the only changes I think are really needed.

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/AnimationMidpoints.kt
rylin8: Yea the cubic bezier curves start out slow and the midpoint test only progresses time by 0.1 before taking the midpoint screenshot. I'll change back to a spring.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshVariantTransition.kt
rylin8: Done

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariantAnimationTest.kt
rylin8: Done

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/VariantAnimationTest.kt
iamralpht: Yeah, I think this is good. Currently only "SmartAnimate" is implemented. Other animation types that Figma has are for the most part simpler because they are just simple translate + opacity done on the incoming and outgoing root views.

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/AnimationMidpoints.kt
iamralpht: Sounds good - otherwise you might be able to use some bigger constants in the first two terms of the cubic bezier to make it do something right way.

iamralpht (approved):

iamralpht:
Looks good - thanks for getting this in shape!

PR #1324, opened 2024-07-04
Fixes  #1301: Replace customizations functions with State<>

### LINKED ISSUE #1301: Update customizations to use State<>
Update the `@Composable () -> String` function to use a `State<String>`, and the `@Composable () -> Meter` function to use a `State<Meter>`. Update all examples to use these instead as well. Make sure these customizations also work in the squoosh branch.

yiqunw700:
- Replace text function customization with `State<String>`
- Replace meter function customization with `FloatState`
- Update MediaCompose usages to the new state customizations
- Create a new validation test for  text state and meter state customizations


timothyfroehlich:
@yiqunw700 Looks like you forgot to add a description or issue link.

yiqunw700:
> @yiqunw700 Looks like you forgot to add a description or issue link.

Updated. Please take a look, thanks!

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt
```diff
@@ -99,6 +101,11 @@ val EXAMPLES: ArrayList<Triple<String, @Composable () -> Unit, String?>> =
         Triple("Text Inval", { TextResizingTest() }, TextResizingTestDoc.javaClass.name),
         Triple("Shared Customization", { ModuleExample() }, ModuleExampleDoc.javaClass.name),
         Triple("Variable Modes", { VariableModesTest() }, VariablesTestDoc.javaClass.name),
+        Triple(
+            "State Customizations",
+            @TargetApi(28) { StateCustomizationsTest(Clock.systemDefaultZone()) },
```
timothyfroehlich, line 106: What's going on here? Why the TargetApi?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/StateCustomizations.kt
```diff
@@ -0,0 +1,67 @@
...
+}
+
+@Composable
+@TargetApi(26)
+fun StateCustomizationsTest(clock: Clock) {
+    val time = remember { mutableStateOf("00:00:00") }
+    val firstHand = remember { mutableFloatStateOf(0F) }
+    val secondHand = remember { mutableFloatStateOf(0F) }
+
+    val handler = Handler(Looper.getMainLooper())
```
timothyfroehlich, line 51: Any reason you used Looper instead of a coroutine?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt
timothyfroehlich, line 106: Does this mean that the test will behave differently depending on your time zone?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/StateCustomizations.kt
```diff
@@ -0,0 +1,67 @@
...
+    val runnable =
+        object : Runnable {
+            override fun run() {
+                val currentTime = LocalTime.now(clock)
```
timothyfroehlich, line 55: This means the example will show different results depending on the time, correct? I think it'd be better to have it use a specified time so that you get the same results whenever you run it.

timothyfroehlich (changes requested):
The biggest thing is that the new clock test in Validation should just behave the same no matter what time it is.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt
yiqunw700: systemDefaultZone() is available on api level 28. The app min api level is 30 so I think @TargetApi is good enough here. Also add @TargetApi then Android Studio won't show the warnings/errors.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/StateCustomizations.kt
yiqunw700: It depends on the Clock passed to the StateCustomizationsTest. The one in validation app will show the device time while the roborazzi test uses a fixed time clock that we get the same results whenever we run it.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt
yiqunw700: Yes. Cause using a different time zone rather than the system default, it will show a different time from the device time.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/StateCustomizations.kt
yiqunw700: I haven't used much coroutine. It looks like coroutine is mostly for asynchronization.
Here the task is very very light to update the UI. So I used handler here to update the UI every second.

timothyfroehlich: Since it's just for a test it's fine, but library code should definitely use coroutines that use the Composable's scopes. It'll be worth looking into :)

yiqunw700: Sure. I will take a look into how to use coroutines.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt
timothyfroehlich: I'm a bit confused why it's showing a warning there, but I dunno, it's probably not causing any issues.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/StateCustomizations.kt
timothyfroehlich: Ok, I see it now. Looks good.

yiqunw700:
> The biggest thing is that the new clock test in Validation should just behave the same no matter what time it is.

The clock in the validation will show the device time which is easy to validate if it works properly.
For the roborazzi test, in order to make it generate the same screenshot every time it runs, it uses a fixed clock.

timothyfroehlich (approved):

PR #1352, opened 2024-07-11
Fixes #1296: Don't fail live update when user has no access to variables

### LINKED ISSUE #1296: Figma Live Update fails if the access token doesn't have read access for variables
@johnoneil just found this out. I don't think we'd realized this before and we'll need to update documentation to note that users will need their token to have both file read and variable read.

yiqunw700:
1. Log an error in logcat when variable fetch fails
2. Show a warning message in DesignSwitcher when variable value look up fails in VariableManager
3. Update NumOrVar and ColorOrVar to hold a fallback value


timothyfroehlich:
Looks pretty great! And you've tested it with a Figma Token that doesn't have Variable access?

yiqunw700:
> Looks pretty great! And you've tested it with a Figma Token that doesn't have Variable access?

Yes, I tested the variable mode test in the validation app with a token without variable access.

crates/figma_import/src/toolkit_font_style.rs
```diff
@@ -100,8 +100,10 @@ impl PartialEq for FontWeight {
                 NumOrVar::Num(o) => f32_eq(n, o),
                 _ => false,
             },
-            NumOrVar::Var(v) => match &other.0 {
-                NumOrVar::Var(o) => v == o,
+            NumOrVar::Var { id: id1, fallback: num } => match &other.0 {
```
iamralpht, line 103: This is the existing structure of the function, but there might be a cleaner way to do it like 
```
match (&self.0, &other.0) {
  (NumOrVar::Num(&a, _), NumOrVar::Num(&b, _)) => a == b,
...
```

iamralpht (approved):
Looks good! Only one minor thing -- can fix now or file a ticket to follow up later.

designcompose/src/main/java/com/android/designcompose/VariableManager.kt
```diff
@@ -200,23 +205,25 @@ internal object VariableManager {
...
             return v.getColor(varMap, variableState)
         }
-        return null
+        Feedback.documentVariableMissingWarning(currentDocId, varId)
+        return fallback
     }
 
     // Given a variable ID, return the number associated with it
-    internal fun getNumber(varId: String, variableState: VariableState): Float? {
+    internal fun getNumber(varId: String, fallback: Float?, variableState: VariableState): Float? {
```
timothyfroehlich, line 220: How does fallback work? Does it get the fallback value from the doc?

timothyfroehlich (changes requested):
Looks good, but I'd like to see documentation added (to `docs/_docs/variables.md`) that provides a good explanation of what's going on. It should include what is required to enable Variable support (Figma Enterprise and a key with the right permissions) and what happens when you don't have Variable support (i.e. how does the fallback work?). Please also update `docs/_docs/live-update/setup.md` to note the need for Variable access in the Figma token (it's currently incorrect).

designcompose/src/main/java/com/android/designcompose/VariableManager.kt
yiqunw700: Yes. The fallback is the value, the number in this case.

yiqunw700:
> Looks good, but I'd like to see documentation added (to `docs/_docs/variables.md`) that provides a good explanation of what's going on. It should include what is required to enable Variable support (Figma Enterprise and a key with the right permissions) and what happens when you don't have Variable support (i.e. how does the fallback work?). Please also update `docs/_docs/live-update/setup.md` to note the need for Variable access in the Figma token (it's currently incorrect).

Updated. Please take a look.
Also, I am wondering if there is a way to preview the documentation change in the browser... I would like to see how my change in the documentation looks..

crates/figma_import/src/toolkit_font_style.rs
yiqunw700: Done.

timothyfroehlich:
> > Looks good, but I'd like to see documentation added (to `docs/_docs/variables.md`) that provides a good explanation of what's going on. It should include what is required to enable Variable support (Figma Enterprise and a key with the right permissions) and what happens when you don't have Variable support (i.e. how does the fallback work?). Please also update `docs/_docs/live-update/setup.md` to note the need for Variable access in the Figma token (it's currently incorrect).
> 
> Updated. Please take a look. Also, I am wondering if there is a way to preview the documentation change in the browser... I would like to see how my change in the documentation looks..

Yep, follow the instructions in docs/README.md. They should still be correct, though I haven't done it in a while.

crates/figma_import/src/toolkit_schema.rs
```diff
@@ -43,7 +43,7 @@ impl NumOrVar {
     ) -> NumOrVar {
         let var = bound_variables.get_variable(var_name);
         if let Some(var) = var {
-            NumOrVar::Var(var)
+            NumOrVar::Var { id: var, fallback: num }
```
rylin8, line 46: Nice that using the value from Figma was relatively easy.

docs/_docs/variables.md
```diff
@@ -15,33 +16,48 @@ nav_order: 11
...
 - Corner radius
 
-Using variables instead of fixed values in your Figma design file requires no change on the developer side. In order to change a variable value in your application, DesignCompose has several functions to help with this.
+In order to change a variable value in your application, DesignCompose has several functions to help
+with this.
+
+However, developers must have a full seat in an **Enterprise** org to use the [Variables REST
+API][3]. If a design file uses variables, developers need to use a [figma auth token][4] that
+has the `file_variables:read` scope to be able to download the variables. If developers are not
+part of Enterprise plan or use an auth token that don't have access to variables, the design will
```
rylin8, line 25: I would say: "part of an Enterprise plan or use an auth token that doesn't have access to variables,"

docs/_docs/variables.md
```diff
@@ -15,33 +16,48 @@ nav_order: 11
...
+use the node values directly and the variable functions, as described below, will not change upon
+changing variable collection or variable mode.
```
rylin8, line 27: I would say something like: "use the node values directly, and the variable functions below will have no effect."

docs/_docs/live-update/setup.md
```diff
@@ -25,7 +25,10 @@ To configure access to your documents, you need to generate an access token.
 
 2. Scroll down to **Personal access tokens** and click **Generate new token**
 
-3. Automotive Design for Compose requires a token with read-only File content access. All other scopes can be left set to "No access". Set the permissions appropriately, enter a name for the token and set the expiration, then click **Generate token**
+3. Automotive Design for Compose requires a token with read-only File content access and read-only Variables access if available. All other scopes can be left set to "No access". Set the permissions appropriately, enter a name for the token and set the expiration, then click **Generate token**
+    * Note: variable access is only available to Enterprise plan. If variable access is not
```
rylin8, line 29: "only available under the Figma Enterprise plan"

docs/_docs/live-update/setup.md
```diff
@@ -25,7 +25,10 @@ To configure access to your documents, you need to generate an access token.
 
 2. Scroll down to **Personal access tokens** and click **Generate new token**
 
-3. Automotive Design for Compose requires a token with read-only File content access. All other scopes can be left set to "No access". Set the permissions appropriately, enter a name for the token and set the expiration, then click **Generate token**
+3. Automotive Design for Compose requires a token with read-only File content access and read-only Variables access if available. All other scopes can be left set to "No access". Set the permissions appropriately, enter a name for the token and set the expiration, then click **Generate token**
+    * Note: variable access is only available to Enterprise plan. If variable access is not 
+      available, developers can still get most of the functionalities work with DesignCompose
```
rylin8, line 30: "most of the DesignCompose functionality"

yiqunw700: Done.

docs/_docs/live-update/setup.md
yiqunw700: Done.

docs/_docs/variables.md
yiqunw700: Done.

docs/_docs/variables.md
yiqunw700: Done.

timothyfroehlich (approved):

PR #1377, opened 2024-07-16
Closes #1375: Migrate Variables from figma_import to dc_bundle

### LINKED ISSUE #1375: Migrate Variables from figma_import to dc_bundle
This includes `NumOrVar`, `Collection` and so on. They should all be moved to `crates/dc_bundle/src/legacy_definition/element/variable.rs`. They will need to be adjusted so that there are no imports from `figma_import/figma_schema`.

You will need to base your changes on top of #1349, which migrates some types that you'll need to work with, such as `Color`.

rylin8:
Move variable declarations into variable.rs. This includes creating some structs that are basically copies of their counterpart in figma_schema.

Add a variable_utils.rs file that includes functions to create variables from figma_schema data.

Fixes #1375

### DIFF
--- a/crates/dc_bundle/src/legacy_definition/element.rs
+++ b/crates/dc_bundle/src/legacy_definition/element.rs
@@ -19,3 +19,4 @@ pub mod font;
 pub mod geometry;
 pub mod node;
 pub mod path;
+pub mod variable;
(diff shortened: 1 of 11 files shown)


timothyfroehlich (approved):
This looks great!

Issue #485, opened 2023-10-19
Support rendering stroked/outlined text


iamralpht:
Stroked or outlined text is used fairly frequently in instrument cluster designs (e.g. for word-based telltales where background contrast is important). Currently DesignCompose supports a single shadow on text, but not stroking or outlining. It looks like Jetpack Compose added support for stroking text in 1.4.0 by adding a `drawStyle` field to the text style object, so hopefully we can just expose that.

yiqunw700:
Added basic support for text stroke: solid, centered text stroke for both non-squoosh and squoosh render.

PR #1379, opened 2024-07-17
Add simple text stroke support

yiqunw700:
Ticket: #485
- Right now only center aligned, solid(non-dash) stroke works as expected.
- Works with squoosh.
- Works with mixed styles (different fills) for single text node.


iamralpht:
Looks good! Can you also add support for this in Squoosh - it will have the same constraints since the Compose `Paragraph.paint` also just takes a `DrawStyle` to control stroke vs fill.

designcompose/src/main/java/com/android/designcompose/DesignText.kt
```diff
@@ -357,9 +363,33 @@ internal fun DesignText(
                     }
                 )
             } else {
+                var textLayoutResult: TextLayoutResult? = null
                 BasicText(
                     annotatedText,
-                    modifier = textModifier,
+                    onTextLayout = { result ->
+                        textLayoutResult = result
```
iamralpht, line 370: Oh, I didn't know we could get text layouts like this -- maybe this would also let us support text shadows better, too!

integration-tests/validation/src/testDebug/roborazzi/RenderAllExamples/Fancy-Fills.png
iamralpht, line 1: Funny that the "WordArt" line break changed. This is the way it's meant to be, but I don't know which change in this PR fixed it.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/FancyFillTest.kt
```diff
@@ -33,3 +36,10 @@ interface FancyFillTest {
 fun FancyFillTest() {
     FancyFillTestDoc.MainFrame(onTap = { Log.e("onTap", "frame clicked!") })
 }
+
+@Composable
+fun SquooshFancyFillTest() {
```
iamralpht, line 41: Is this referenced from the AllRenderTests file? Ok to do so later -- I think we should probably add a "Squoosh switch" to the validation app so we don't have to write these anymore... (Maybe doing this would be better than updating AllRenderTests...)

iamralpht (approved):

designcompose/src/main/java/com/android/designcompose/DesignText.kt
```diff
@@ -244,10 +249,11 @@ internal fun DesignText(
 
     val textBrushAndOpacity =
         style.node_style.text_color.asBrush(document, density.density, variableState)
+
     val textStyle =
         @OptIn(ExperimentalTextApi::class)
         TextStyle(
-            brush = customBrush ?: textBrushAndOpacity?.first,
+            brush = customBrush ?: textBrushAndOpacity?.first ?: SolidColor(Color.Transparent),
```
rylin8, line 256: What does this change solve?

crates/figma_import/src/toolkit_style.rs
```diff
@@ -433,7 +433,7 @@ pub struct NodeStyle {
 impl Default for NodeStyle {
     fn default() -> NodeStyle {
         NodeStyle {
-            text_color: Background::Solid(ColorOrVar::Color(Color::from_u8s(0, 0, 0, 255))),
+            text_color: Background::None,
```
rylin8, line 436: Why'd this change?

designcompose/src/main/java/com/android/designcompose/DesignText.kt
rylin8: So you have to wait for the text layout result here, and then use that in the modifier below? That's confusing to me.

integration-tests/validation/src/testDebug/roborazzi/RenderAllExamples/Fancy-Fills.png
yiqunw700: I changed the layout for the "WordArt" from **_AutoHeight_** to **_AutoWidth_** in the figma file.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/FancyFillTest.kt
yiqunw700: Filed ticket #1384 to track this.

Right now I am just changing it manually from FancyFillTest to SquooshFancyFillTest in the AllExamples file and rebuild to test.

designcompose/src/main/java/com/android/designcompose/DesignText.kt
iamralpht: I think text layout happens before the draw modifier is executed, so there should always be a valid layout. It might be worth logging if there isn't (although perhaps empty strings don't generate a layout? Then the log could be annoying).

crates/figma_import/src/toolkit_style.rs
yiqunw700: The text only with strokes doesn't have a fill. But this defaults to solid black and will draw the fill as black color. So I changed it to default to `None`.

designcompose/src/main/java/com/android/designcompose/DesignText.kt
yiqunw700: When the brush is null, it also draws the text black. So I have to set it to transparent color to draw nothing visually.

rylin8 (approved):

designcompose/src/main/java/com/android/designcompose/DesignText.kt
yiqunw700: In my understanding, this is how it works: 

When the text view is drawn on the canvas, it will measure -> layout -> draw.
- Then when it layouts, it will call the `onTextLayout` callback and we capture the `textLayoutResult`.
- Then when it draws, it calls the `Modifier.drawWithContent` and it can use the  `textLayoutResult` to draw the strokes.

I tried to set drawStyle to the TextStyle but it can only take one brush so it will not be able to draw the fill and stroke at the same time.
Other options is put two BasicText one on top of another. This solution is the best way I've gotten so far.

Issue #1384, opened 2024-07-19
Toggle for squoosh on/off in validation app


yiqunw700:
Worked on some rendering improvements which need to test both default impl and squoosh impl too. Right now I am just changing it manually in the AllExamples file and rebuild to test.

PR #1385, opened 2024-07-20
Create a squoosh toggle for validation app

yiqunw700:
Fixes: #1384 

1. Breaks all examples into 3 category: works only with squoosh, works only with default renderer, works with both.
2. Updates the main activity to include a switch to toggle squoosh renderer on/off. 
Screenshot 2024-07-22 at 16 34 52

3. Updates roborazzi to capture squoosh screenshots too for examples working with both renderers.


timothyfroehlich (approved):
Overall I think this change looks good, though I don't think we should add extra overlays in the example display area. Could you move the Squoosh toggle to be above the list of examples?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt
```diff
@@ -109,3 +76,40 @@ val EXAMPLES: ArrayList<Triple<String, @Composable () -> Unit, String?>> =
...
+            { RecursiveCustomizations() },
+            RecursiveCustomizationsDoc.javaClass.name
+        ),
+        Triple("Shared Customization", { ModuleExample() }, ModuleExampleDoc.javaClass.name),
+        Triple(
+            "Variant Interactions",
+            { VariantInteractionsTest() },
+            VariantInteractionsTestDoc.javaClass.name
+        ),
+    )
```
rylin8, line 115: Did you manually test each validation test in squoosh and non-squoosh mode to see which ones squoosh had problems with?

rylin8 (commented):
So for the set of tests that we do in both squoosh and non-squoosh we don't compare their PNGs yet, correct?

yiqunw700:
> So for the set of tests that we do in both squoosh and non-squoosh we don't compare their PNGs yet, correct?

No we don't compare yet.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt
yiqunw700: I did test each validation test but I didn't looked deep into the differences.
Squoosh only tests have the two tests for animations and default collection has those with list/grid layout which squoosh doesn't support yet.
The set we do in both squoosh and non-squoosh, there are still some differences which I believe there are bugs with squoosh there: for example squoosh doesn't cut off the texts into two lines with the Text Elide test.

rylin8: Got it. Would be nice to get a comparison of squoosh against non-squoosh sometime, and I know @iamralpht started a doc with a table describing what is different or broken in squoosh. Not for this PR though.

rylin8 (approved):

yiqunw700:
> Overall I think this change looks good, though I don't think we should add extra overlays in the example display area. Could you move the Squoosh toggle to be above the list of examples?

Done.

PR #1403, opened 2024-07-22
Replace the FontStyle and FontFeature Rust types with the Protobuf implementations

timothyfroehlich:
Why were we converting the font features to a byte array and then back to a string? Either way, it needed to change one way or another because you can't have protobuf arrays with a specified size.


rylin8:
@timothyfroehlich could explain what the problem with the italic text was? Also, do you know why the dial&gauge test is failing?

timothyfroehlich:
> @timothyfroehlich could explain what the problem with the italic text was? Also, do you know why the dial&gauge test is failing?

The problem was that the FontStyle enum shifted by one. 0 was Normal and 1 was Italics, but now 0 is Unspecified, 1 is Normal and 2 is Italics. I hadn't updated the DCF for HelloWorld so it was deserializing the DCF where FontStyle was 1, and with this update, 1 is now Normal.

One mitigation would have been to bump the number of the DCF version, so that old DCFs would be rejected, but that seemed unnecessary for a small change like this that will be followed by other changes.

timothyfroehlich:
@rylin8 Also, do you know why the dial&gauge test is failing?

Hadn't updated the screenshots.

crates/dc_bundle/src/definition/element.rs
```diff
@@ -0,0 +1,30 @@
...
+impl Display for FontStyle {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        Debug::fmt(self, f)
+    }
+}
+impl FontFeature {
+    pub fn new(tag: String) -> Self {
+        FontFeature { tag, enabled: true }
+    }
+}
```
rylin8, line 30: Is the plan to add other `impl` functions for types in `legacy_definition/element/` as we convert new types?

crates/figma_import/src/toolkit_style.rs
```diff
@@ -16,12 +16,11 @@
 //! uses.
 
 use crate::figma_schema;
+use dc_bundle::definition::element::{FontFeature, FontStyle};
 use dc_bundle::definition::layout::FlexWrap;
```
rylin8, line 20: I thought this was the first time using `dc_bundle::definition`. When was `FlexWrap` converted?

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1192,10 +1192,9 @@ fn visit_node(
             let mut font_features = Vec::new();
             for (flag, value) in flags {
                 let flag_ascii = flag.to_ascii_lowercase();
-                let flag_bytes = flag_ascii.as_bytes();
-                if flag_bytes.len() == 4 {
-                    let tag = [flag_bytes[0], flag_bytes[1], flag_bytes[2], flag_bytes[3]];
-                    font_features.push(FontFeature { tag, enabled: *value == 1 });
+                if flag_ascii.len() == 4 {
+                    // Smoke check to see if the flag is valid
```
rylin8, line 1196: What is this "Smoke check" referring to?

crates/dc_bundle/src/definition/element.rs
timothyfroehlich: You mean `definition/element/`?

Yeah. If it starts getting unwieldly then we can split them into seperate submodules of element.rs, like how they are in legacy_definition.

crates/figma_import/src/toolkit_style.rs
timothyfroehlich: I actually did that a month or so ago to make sure that it worked, but that was just an enum that had no impls.

crates/figma_import/src/transform_flexbox.rs
timothyfroehlich: A bit of crude validation. In this case validating that the flag is exactly 4 characters.

rylin8:
@yiqunw700's change just went in so you'll have to update all the .dcf files

yiqunw700:
> @yiqunw700's change just went in so you'll have to update all the .dcf files

Yeah. Sorry about the conflicts. 😿

rylin8 (approved):

PR #1405, opened 2024-07-22
Closes #1369: Support progress bars that move along a vector path

### LINKED ISSUE #1369: Support progress bars that move along a vector path
Currently our support for progress bars is only for rectangular shaped progress bars. Support any vector shape by following the path of the vector, only partially filling it in.

On the Kotlin side, this involves drawing partial lines and curves. Doing so may require splitting curves up using an algorithm such as [De Casteljau's algorithm](https://en.wikipedia.org/wiki/De_Casteljau%27s_algorithm).

On the FIgma plugin side, we need to support enabling progress bars on any vector shape. Enabling a preview of the progress bar would likely require drawing partial curves as well using the same algorithm.

rylin8:
Vectors in Figma that have a stroke path, not a fill path, can now be used as a progress bar. The Dials and Gauges plugin now has a "Progress vector" section that can be enabled on vector nodes. Currently there is no preview feature for vector progress bars, as there is not a straightforward way to render a partial vector path using Figma's APIs.

DesignCompose applications parse this new vector progress data an render a partial path by using the built in dashed path effect.


rylin8:
Added some tests for dials & gauges

crates/figma_import/src/toolkit_style.rs
```diff
@@ -179,13 +180,59 @@ pub struct ProgressMarkerMeterData {
...
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub enum MeterData {
     ArcData(ArcMeterData),
     RotationData(RotationMeterData),
     ProgressBarData(ProgressBarMeterData),
     ProgressMarkerData(ProgressMarkerMeterData),
+    ProgressVectorData(ProgressVectorMeterData),
+}
+impl From<MeterDataSchema> for MeterData {
```
iamralpht, line 224: Fwiw now that I think about it, MeterData probably belongs in the node definition (because it's functional) rather than the style definition (which is more presentation stuff), but it's OK to leave it like this for now.

designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
```diff
@@ -130,7 +131,9 @@ internal fun ViewShape.computePaths(
                 )
             }
             is ViewShape.Path -> {
-                getPaths(this.path, this.stroke)
+                // If strokePaths is not null, use it instead of this.stroke since it contains the
+                // stroked vector path of a progress vector.
+                getPaths(this.path, strokePaths ?: this.stroke)
```
iamralpht, line 136: Is there a way to move this to the transform_flexbox instead of at runtime, or is it likely to regress other functionality?

rylin8: Do you mean saving the plugin data's stroked paths into the node's stroke list instead of separately?

crates/figma_import/src/toolkit_style.rs
rylin8: I am planning to refactor this a bit during the `dc_bundle` and protobuf migration.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -229,6 +233,35 @@ private fun calculateArcData(
...
+    style: ViewStyle,
+    meterValue: Float,
+    density: Float
+) {
+    val strokeWidth = style.node_style.stroke.stroke_weight.toUniform() * density
+    val discretizedMeterValue = meterValue.coerceDiscrete(data.discrete, data.discreteValue)
+
+    // Get full length of path
+    var pathLen = 0f
+    paths.forEach {
```
iamralpht, line 251: I also wonder if this is something that can be done during doc conversion, but perhaps there are scenarios where layout can change the path length?

designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
rylin8: Done

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: I think scaled vectors can cause layout to change the size. We don't support scaled vector paths properly right now, though we should.

iamralpht (approved):

Issue #1406, opened 2024-07-22
[Squoosh Validation] Squoosh doesn't wrap text into two lines and cut off with no ellipsis in "Text Elide" test


yiqunw700:
![image](https://github.com/user-attachments/assets/f631f05c-65c7-4a71-8d5d-36c2585cde56)

PR #1407, opened 2024-07-23
Fix the squoosh for TextElide rendering

yiqunw700:
Fixes: #1406
The third line should be cut off. Max line in this case is Int.MAX and is not gating the max lines being drawn. Using canvas.clip to clip to the computed height plus extra space for shadow and stroke.


rylin8:
Looking at the Alignment test, it looks like the "28px Line Height C" text on the bottom right has been improved. So now it calculates the max lines based off of layout size, whereas before it would wrap -- is that correct? I looked at the non-squoosh version and it looks like it also wraps, but doesn't render the second line.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
```diff
@@ -248,13 +248,30 @@ private fun squooshTextRender(
...
         )
+    // Clip canvas will clip the shadow too. Not sure if there is other better ways to compute the
+    // max lines manually.
+    val bottom = computedLayout.height * density.density
+    val cutLine = paragraph.getLineForVerticalPosition(bottom)
+    val cutLineBottom = paragraph.getLineBottom(cutLine)
+    if (cutLineBottom > bottom) {
+        val maxLines = cutLine.coerceAtLeast(1)
+        if (maxLines != textInfo.maxLines) {
+            paragraph =
```
iamralpht, line 266: I think this should take a fast path through `Paragraph` because the width isn't changing, so it might be OK to construct a paragraph here during render rather than during layout -- but please do see if it's possible to do this during layout instead. We should try to get this change in as it's a big improvement in text compatibility.

yiqunw700: I looked into how the layout is done but I am not sure if I get it right.
The node in the figma file has a fixed width `291` and a fixed height `61`. The size is passed to taffy to do the measurement which will call the https://github.com/google/automotive-design-compose/blob/170257c7a91c2de2335878b337b09e5424c5654d/crates/layout/src/layout_manager.rs#L183 which then calls the kotlin https://github.com/google/automotive-design-compose/blob/170257c7a91c2de2335878b337b09e5424c5654d/designcompose/src/main/java/com/android/designcompose/DesignText.kt#L482 to do the text measurement.

However, the size passed into the measure function for the height though is 0.0 with AvailableHight to be MinContent so I got 0s for the height and available height in the kotlin measure code https://github.com/google/automotive-design-compose/blob/170257c7a91c2de2335878b337b09e5424c5654d/designcompose/src/main/java/com/android/designcompose/DesignText.kt#L482 where I think is the best place to calculate how many lines the layout could fit.

I will check with @rylin8 tomorrow when we are in the office to see if he can help me figure out more about it.

yiqunw700: I changed to do the canvas clipping approach because of max lines can not cut off part of the line:
Screenshot 2024-08-20 at 15 58 38

rylin8:
Do you know why Text-Elide_SQUOOSH_compare.png looks like it got a lot wider?

yiqunw700:
> Do you know why Text-Elide_SQUOOSH_compare.png looks like it got a lot wider?

No. Not really.

It seems the screenshot captures the UI outside of the DesignDoc. @timothyfroehlich Tim, do you know if it is expected? I added some UI elements below the figma design in the TextElideTest.

timothyfroehlich:
> > Do you know why Text-Elide_SQUOOSH_compare.png looks like it got a lot wider?
> 
> No. Not really.
> 
> It seems the screenshot captures the UI outside of the DesignDoc. @timothyfroehlich Tim, do you know if it is expected? I added some UI elements below the figma design in the TextElideTest.

I think I know what's going on, will dig in soon.

timothyfroehlich:
@yiqunw700 try cherry-picking https://github.com/google/automotive-design-compose/pull/1547. I changed the way it captures the images to only capture the DesignComposable instead of the whole test composable.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/TextElideTest.kt
```diff
@@ -25,9 +41,71 @@ import com.android.designcompose.annotation.DesignDoc
...
+            Row(modifier = Modifier.padding(4.dp), verticalAlignment = Alignment.CenterVertically) {
+                RadioButton(
+                    selected = (it == stageState.value),
+                    onClick = { stageState.value = it }
+                )
+                Text(text = it, modifier = Modifier.padding(start = 4.dp))
+            }
+        }
+
+        HorizontalDivider(modifier = Modifier.fillMaxWidth().padding(bottom = 4.dp))
```
rylin8, line 83: This `fillMaxWidth()` might be why the screenshot stretched to be so wide. I'm not sure what other consequences there are to @timothyfroehlich's screenshot changes -- if we don't want to use that yet you could try removing this.

yiqunw700: I like Tim's change to just capture the design doc. But right now there is one screenshot mismatch I haven't figured out what is happening.
I have run the `./gradlew recordRoborazziDebug` and uploaded new screenshots but the checks failed on the mask test.

integration-tests/validation/src/testDebug/roborazzi/RenderAllExamples/Masks.png
rylin8, line 1: Interesting that non-squoosh masks look broken in several places. Don't know why the bottom circular ticks don't have a white background. I also don't understand why this changed -- it got taller by 200 pixels for some reason? I guess the roborazzi change captures a larger area than before?

rylin8 (commented):
@yiqunw700 what changed that fixed the mask test?

integration-tests/validation/src/testDebug/roborazzi/RenderAllExamples/Masks.png
yiqunw700: Roborazzi now captures the design doc area, it is 3200px high which matches the height 1600dpof the Mask Test.

The RenderAllExamples is using a 1500dp emulated device. Before it captures the 1500dp height displayed on the device. Now it captures the 1600dp of the design doc with the part being transparent outside of the screen.

The squoosh mask may not work. It seems it just displays whatever in the design doc.

yiqunw700: The changes I made to fix the check:
https://github.com/google/automotive-design-compose/compare/b7d432c54e20b65925fe1ab510db01d2cd395cf8..a781a21a688b14af47b9babab4cabf0a45bad235

It was to capture root and capture design doc node which captures twice. I cleaned it up to capture the design doc node only.

rylin8 (approved):

PR #1415, opened 2024-07-23
Copy FigmaColor to dc_bundle as FloatColor

yiqunw700:


crates/figma_import/src/reflection.rs
```diff
@@ -177,6 +177,9 @@ pub fn registry() -> serde_reflection::Result<serde_reflection::Registry> {
     tracer
```
yiqunw700, line 177: Let me know if I should remove it or keep it..
For the figma_schema::FigmaColor. I am not quite sure why some need in reflection and some not.

timothyfroehlich: Yes, remove the FigmaColor one. Reflection basically defines the types that the serde code generates Java code for. Primarily that is the types used in DesignDefinition. The goal is to keep FigmaColor out of DesignDefinition so it should be removed from reflection.rs.

timothyfroehlich (changes requested):

crates/figma_import/src/reflection.rs
yiqunw700: Done.

crates/figma_import/src/variable_utils.rs
```diff
@@ -70,7 +71,9 @@ fn create_variable_value(v: &figma_schema::VariableValue) -> VariableValue {
         figma_schema::VariableValue::Boolean(b) => VariableValue::Bool(b.clone()),
         figma_schema::VariableValue::Float(f) => VariableValue::Number(f.clone()),
         figma_schema::VariableValue::String(s) => VariableValue::Text(s.clone()),
-        figma_schema::VariableValue::Color(c) => VariableValue::Color(c.into()),
+        figma_schema::VariableValue::Color(c) => VariableValue::Color(
+            (&(<&figma_schema::FigmaColor as Into<FloatColor>>::into(c))).into(),
+        ),
```
rylin8, line 76: This goes FigmaColor -> FloatColor -> Color ? Can we just keep the implementation of Into<Color> for &FigmaColor ?

yiqunw700: Done.

timothyfroehlich (approved):

rylin8 (approved):

PR #1418, opened 2024-07-23
Update crates to use package settings from workspace, set version to 0.30.0

timothyfroehlich:


Cargo.toml
```diff
@@ -2,6 +2,13 @@
 resolver = "2"
 members = ["crates/figma_import", "crates/dc_jni", "crates/layout", "crates/dc_bundle"]
 
+[workspace.package]
+version = "0.30.0"
```
rylin8, line 6: from 0.1.0 to 0.30.0? Is this supposed to match our release version?

iamralpht: Yeah - basically we've got other consumers of some of the Rust crates we're making, so it's better if the version changes when they incorporate new releases from us. Tim and I talked and didn't think it made sense to track the individual crates because we'll just end up with loads of versions, so we're doing a global version for all of the crates. (No plans to publish on crates.io, though).

iamralpht (approved):

PR #1421, opened 2024-07-24
Move reaction_schema to dc_bundle

yiqunw700:
Also copy figma_schema::Vector as Vector under dc_bundle.

### DIFF
--- a/crates/dc_bundle/src/legacy_definition/element/vector.rs
+++ b/crates/dc_bundle/src/legacy_definition/element/vector.rs
@@ -13,14 +13,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+use serde::{Deserialize, Serialize};
 
-pub mod background;
-pub mod color;
-pub mod font;
-pub mod geometry;
-pub mod node;
-pub mod path;
-pub mod reactions;
-pub mod variable;
-pub mod vector;
-pub mod view_shape;
+// XXX ColorStop, Transform
+#[derive(Deserialize, Serialize, Debug, Clone, Default, PartialEq)]
+pub struct Vector {
+    pub x: Option<f32>,
+    pub y: Option<f32>,
+}
+
+impl Vector {
+    pub fn is_valid(&self) -> bool {
+        self.x.is_some() && self.y.is_some()
+    }
+    pub fn x(&self) -> f32 {
+        self.x.unwrap_or(0.0)
+    }
+    pub fn y(&self) -> f32 {
+        self.y.unwrap_or(0.0)
+    }
+}
--- a/crates/figma_import/src/reaction_schema.rs
+++ b/crates/figma_import/src/reaction_schema.rs
@@ -14,9 +14,11 @@
 
 use serde::{Deserialize, Serialize};
 
-use crate::figma_schema;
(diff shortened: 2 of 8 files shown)


crates/figma_import/src/reaction_schema.rs
```diff
@@ -476,44 +64,11 @@ fn parse_reactions() {
...
-
-#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Copy)]
-#[serde(rename_all = "SCREAMING_SNAKE_CASE")]
-pub enum OverlayBackgroundInteraction {
-    None,
-    CloseOnClickOutside,
-}
-
 #[test]
 fn parse_frame_extras() {
```
yiqunw700, line 68: Let me know if the tests should also be moved to dc_bundle.

crates/dc_bundle/src/legacy_definition/element/vector.rs
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+use serde::{Deserialize, Serialize};
+
+// XXX ColorStop, Transform
```
timothyfroehlich, line 18: Do you happen to know what this comment is supposed to mean?

timothyfroehlich (commented):
So we need to do this a bit differently. The "Json" versions of the structs all need to stay in `figma_import`. They're the data that's being fetched from Figma, so that's why they need to stay there. So with that, the conversions from the "Json" versions also need to stay in `figma_import`, so they'll need to be "turned around", sort of, so that they're implementations on the "Json" structures, instead of implementations on the "non-Json" structures.

timothyfroehlich (changes requested):
(Also if you move the Json versions back then it's more clear that the tests should stay in `figma_import`)

crates/dc_bundle/src/legacy_definition/element/vector.rs
yiqunw700: Not really. @rylin8 Robert do you know what does this comment mean?

yiqunw700:
> So we need to do this a bit differently. The "Json" versions of the structs all need to stay in `figma_import`. They're the data that's being fetched from Figma, so that's why they need to stay there. So with that, the conversions from the "Json" versions also need to stay in `figma_import`, so they'll need to be "turned around", sort of, so that they're implementations on the "Json" structures, instead of implementations on the "non-Json" structures.

Done.

crates/figma_import/src/reaction_schema.rs
yiqunw700: Done.

crates/dc_bundle/src/legacy_definition/element/vector.rs
rylin8: Not exactly, though usually @iamralpht adds the `// XXX` comments to indicate something we should fix or add support for in the future.

rylin8 (approved):

timothyfroehlich (approved):
Looks great!

PR #1438, opened 2024-07-26
Closes #1437: Support Round and Square stroke caps for vector progress bars

### LINKED ISSUE #1437: Support Round and Square stroke caps for vector progress bars

rylin8:


crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1471,9 +1469,12 @@ fn visit_node(
         | figma_schema::NodeData::Line { vector }
         | figma_schema::NodeData::RegularPolygon { vector }
         | figma_schema::NodeData::Star { vector }
-        | figma_schema::NodeData::Vector { vector } => {
-            ViewShape::Path { path: fill_paths, stroke: stroke_paths, is_mask: vector.is_mask }
-        }
+        | figma_schema::NodeData::Vector { vector } => ViewShape::Path {
+            path: fill_paths,
+            stroke: stroke_paths,
+            stroke_cap: node.stroke_cap.clone().into(),
```
iamralpht, line 1475: Weirdly I think stroke cap is a style thing, but up to you if you want to keep it here.

iamralpht:
I can't tell what's different in the renders. I assume you'll add test cases for the butt/round cap distinction next. Can you also include some complex fills / patterns? (I'm also hopeful that these work as masks if there's something really complex to explore).

rylin8:
> I can't tell what's different in the renders. I assume you'll add test cases for the butt/round cap distinction next. Can you also include some complex fills / patterns? (I'm also hopeful that these work as masks if there's something really complex to explore).
> 
> Edit: also, combinations with other effects like shadows and blend modes. I'm not sure what an inset shadow should do here.

I added three small curves to dials & gauges: one with no end cap, one with round, and one with square. I haven't added any tests for vectors that are not progress bar vectors because Figma already returns a fill vector path that includes the end caps. Or do you just mean I should add more complex cases for progress bar vectors?

iamralpht:
> Or do you just mean I should add more complex cases for progress bar vectors?

Yeah, I think it would be good to have some more complex examples in case some combinations reveal other issues. Maybe this is a different test section for combinations, though, because the existing tests are bad for testing too many things in one screenshot.

rylin8:
Added additional tests, and fixed a bug where turning progress vector off in the plugin did not work.

iamralpht (approved):

crates/figma_import/src/transform_flexbox.rs
rylin8: I'll leave it for now, since it's also in the `Arc` shape.

Issue #1215, opened 2024-06-08
Add squoosh support for custom list content


rylin8:
The `ReplacementContent` customization in squoosh does nothing.

rylin8:
This is done

PR #1439, opened 2024-07-26
Towards #1215: Improved nested layout in Squoosh, by bridging Rust and Compose layout.

iamralpht:
This change uses a Rust Layout "measure" function (previously only used for text) when embedding other Composables (either DesignCompose generated ones, or any other Composable). The measure function implementation uses the intrinsic measurement functions of the child Composable Layout node to query the best size.

In addition, this change also cleans up Layout ID allocation in Squoosh, removing some legacy bits which were needed before Rust Layout supported multiple Layout Managers directly.

A new test case has been added which exercises nested layout using DesignCompose's ReplacementContent customization. This works well in the test case, but unfortunately the Design Switcher still shows some issues with wrapping and overlapping on occasion, so future work will include updating the test case to cover that exact scenario.


iamralpht:
Successor to https://github.com/google/automotive-design-compose/pull/1294

iamralpht:
More work is needed here, because the precise setup used by the Design Switcher still exhibits some bad behavior, but this change does fix various other cases. I think we can merge this change, and then follow up with another test case that better mirrors the Design Switcher (and a fix for that case).

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/ComponentReplaceRelayout.kt
```diff
@@ -0,0 +1,86 @@
...
+
+@Composable
+fun ComponentReplaceRelayoutTest() {
+    val (textIndex, setTextIndex) = remember { mutableStateOf(0) }
+
+    ComponentReplaceRelayoutDoc.Main(
+        modifier = Modifier.fillMaxWidth(),
+        a =
+            ReplacementContent(
+                count = 5,
```
rylin8, line 60: I think it might be worth testing the list count changing. Maybe another button can cycle through different list sizes.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshLayout.kt
```diff
@@ -106,9 +121,58 @@ internal class SquooshLayoutIdAllocator(
...
+private const val ID_USAGE_COMPONENT = 0x40000000
+private const val ID_USAGE_OVERLAY = 0x80000000.toInt()
+private const val ID_USAGE_LIST_ITEM = 0xc0000000.toInt()
+
+/// Compute a "componentLayoutId" from an incoming root layout ID, and a component instance ID.
+internal fun computeComponentLayoutId(rootLayoutId: Int, componentLayoutId: Int): Int {
+    return (rootLayoutId + componentLayoutId.shl(16)).and(ID_USAGE_MASK) + ID_USAGE_COMPONENT
+}
+
+/// Compute a final layout ID from a parent layout ID (16bits) and a unique node ID from the DCF
```
rylin8, line 156: should this comment say `parent` or `root`?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -105,41 +117,35 @@ internal class SquooshAnimationRenderingInfo(
 
 /// Apply layout constraints to a node; this is only used for the root node and gives the DC
 /// layout system the context of what it is being embedded in.
-private fun SquooshResolvedNode.applyLayoutConstraints(constraints: Constraints, density: Density) {
+private fun SquooshResolvedNode.applyLayoutConstraints(constraints: Constraints, density: Float) {
     val rootStyleBuilder = style.asBuilder()
-    val layoutStyleBuilder = style.layout_style.asBuilder()
+    val layoutStyleBuilder = view.style.layout_style.asBuilder()
```
rylin8, line 122: What is the difference between `style` and `view.style`?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -255,10 +259,11 @@ fun SquooshRoot(
             else -> ""
         }
 
-    val layoutIdAllocator = remember { SquooshLayoutIdAllocator() }
-    val rootLayoutId = remember(docId) { SquooshLayout.getNextLayoutId() * 100000000 }
+    val layoutManager = remember(docId) { SquooshLayout.newLayoutManager() }
+    val layoutIdAllocator = remember(docId) { SquooshLayoutIdAllocator() }
     val layoutCache = remember(docId) { HashMap<Int, Int>() }
     val layoutValueCache = remember(docId) { HashMap<Int, Layout>() }
+    val rootLayoutId = 0 // XXX: Remove -- we needed this when there was only one layout manager.
```
rylin8, line 266: remove?

rylin8 (commented):
Layout code is confusing as it always is to me, but I like this change better than the previous one. Can we enable some of the "Default Only" tests to "BOTH" now, so such as the "Item Spacing" one? I know that uses list content replacement.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
iamralpht: Yep, removed, and also cleaned up the rest of the references to `rootLayoutId` since it doesn't really exist now.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
iamralpht: I added a comment -- basically when this method was getting called during intrinsic width/height queries, I noticed that we were returning inconsistent widths/heights over time, because we were slowly mutating `style` (which only gets recreated when we make a new `SquooshResolvedNode` during recomposition -- but layout can happen any number of times before the next recomposition). So this just ensures that we change the view that layout has without corrupting the root node's layout style.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -255,8 +265,8 @@ fun SquooshRoot(
             else -> ""
         }
 
-    val layoutIdAllocator = remember { SquooshLayoutIdAllocator() }
-    val rootLayoutId = remember(docId) { SquooshLayout.getNextLayoutId() * 100000000 }
+    val layoutManager = remember(docId) { SquooshLayout.newLayoutManager() }
```
rylin8, line 268: This needs to move back up before line 241 so that `SquooshLayout.keepJniBits()` doesn't cause a crash because the layout manager can't be found.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/ComponentReplaceRelayout.kt
iamralpht: Added some buttons to change the count to the test page.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
iamralpht: I don't see where the crash comes from -- I think the call to `keepJniBits` is a static method that just calls a JNI method with a bad handle (so the Rust side will log an error). I don't see a crash running Validation, but I could be missing something :/.

iamralpht:
Rebased on main and resolved conflicts.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
rylin8: Yea there will be a bad handle in Rust, and then it will throw an exception. This only happens if there is no cached `.dcf` file.
```
        throw_basic_exception(
            &mut env,
            &GenericError(format!("No manager with id {}", manager_id)),
        );
```

iamralpht: Ok, I think I can probably delete `keepJniBits` entirely. I had it in because early on when working on Squoosh, the compiler would optimize away DC Classic, but since it's now a runtime thing whether we use Squoosh or not, DC Classic won't get optimized away.

rylin8:
How is item spacing handled? I tried the "Item Spacing" validation test and found that the items are spaced out more in squoosh:
Normal:
Screenshot 2024-08-20 at 11 45 13 AM
Squoosh:
Screenshot 2024-08-20 at 11 45 19 AM

iamralpht:
> How is item spacing handled? I tried the "Item Spacing" validation test and found that the items are spaced out more in squoosh.

Looks like it's triple applied or something, hmm. I don't think there's any explicit handling for item spacing, so I think somehow the children are getting expanded or asking for more space.

rylin8:
> > How is item spacing handled? I tried the "Item Spacing" validation test and found that the items are spaced out more in squoosh.
> 
> Looks like it's triple applied or something, hmm. I don't think there's any explicit handling for item spacing, so I think somehow the children are getting expanded or asking for more space.

This example is using a fixed size container. Changing it to "hug" makes the spacing correct.

rylin8:
Filed https://github.com/google/automotive-design-compose/issues/1541 to follow up after this is merged

rylin8 (approved):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
iamralpht: Ok, removed `keepJniBits`, will merge upon successful CI.

PR #1442, opened 2024-07-26
Update the localization plugin documentation since we published it

yiqunw700:


docs/_docs/localization.md
```diff
@@ -10,10 +10,17 @@ We are developing tools for developers to localize the figma design into differe
...
 used by the figma design.
 
-## 1. Build the localization plugin
+## 1. Run the localization plugin
 
-The localization plugin is still under development and is only available from building the source code.
-Follow [Get Started][1] page to download the source code.
+We have published our first version of the localization plugin. You can run it through:
+**Figma Main Menu -> Plugins -> From Google**
+   **-> Automotive Design For Compose -> Localization Tool -> Generate String Resource**
```
rylin8, line 17: I don't think it's under the "From Google" section -- at least it's not for me.

yiqunw700: There are two lists sorted alphabetically under "From Google". It is in the bottom.

rylin8: Oh, never noticed that, ok.

rylin8 (approved):

PR #1443, opened 2024-07-26
Migrate vector_schema structs to dc_bundle

rylin8:
This commit also fixes a bunch of Rust compile warnings.


crates/dc_bundle/src/legacy_definition/element/path.rs
```diff
@@ -75,3 +75,171 @@ impl Default for Stroke {
...
+// toolkit schema.
+//
+// The serialized form should be somewhat compact and fast to deserialize
+// because we're storing different types (commands, values) in big arrays,
+// reducing the number of objects that need to be constructed during
+// deserialization.
+
+#[repr(u8)]
+#[derive(Clone, Copy, PartialEq, Debug)]
+pub enum PathCommand {
```
timothyfroehlich, line 91: PathCommand and some of the others here aren't serialized, but you think it makes sense to move to `dc_bundle`?

timothyfroehlich (approved):
Looks good otherwise. Thanks for fixing the warnings!

crates/dc_bundle/src/legacy_definition/element/path.rs
rylin8: These types or so tightly bound to the other ones here that use them that it made sense to keep them here. These are not serialized but they're converted to u8s that are then serialized.

PR #1444, opened 2024-07-29
Allow explicitly excluding a node from localization

### LINKED ISSUE #1286: Make localization plugin able to exclude a node from translation
A follow up ticket of https://github.com/google/automotive-design-compose/issues/1242

Basically this means this node is not translatable.

yiqunw700:
Fixes #1286

UI-wise, use a check box to handle the action to exclude.


support-figma/extended-layout-plugin/src/ui.html
```diff
@@ -942,6 +942,20 @@
...
+        parent.postMessage({
+          pluginMessage: {
+            msg: 'localization-exclude-node',
+            node: nodeId
+          }
+      }, '*');
+      } else {
+        showToast("Action canceled...");
+      }
+    }
```
rylin8, line 958: This works, but I find it a little hard to find this feature. There's nothing in the UI to indicate that double clicking would exclude the node. I think some help or instructions are needed if we want to use a double click. Also, once you've excluded a node you can't seem to undo it without clearing out all the localization data. In my opinion, it would be better to have a column (yes I know there are a bit too many columns now) with a checkbox to exclude the node, which would also grey out the rest of the line. This is more a product design issue and I know we'd like to improve other areas as well, but this feature seems harder to use than the others. Maybe get some thoughts from @iamralpht and @timothyfroehlich

yiqunw700: I changed the double click to use a checkbox in front of the node id.

In order to support the undo action, I ended up including the excluded nodes in the output table like always.

Please take another look and let me know what you think.

rylin8: Ok I like this a lot better. There are some formatting issues when a text node is excluded, but that can be addressed when we get a designer to help...
Screenshot 2024-07-30 at 9 57 34 PM

rylin8 (approved):

support-figma/extended-layout-plugin/src/ui.html
yiqunw700: I made it wrap when it overflows. 
Screenshot 2024-07-31 at 10 49 31

Issue #1371, opened 2024-07-15
Allow different translations for same text


yiqunw700:
A follow up ticket of https://github.com/google/automotive-design-compose/issues/1242

Right now thinking about an option to just not group same texts to use the same string resource name.

PR #1445, opened 2024-07-29
Add a group option which allows to use different string resource name for nodes with same text

yiqunw700:
Fixes: #1371
Also made a change when user rename a string resource name to an existing one, it allows to merge the two together.

This user action will not be saved to the plugin data so it gets lost the next time running the plugin. This is also because user can use a different group option which may change the string resource name assigned to a text node.

### DIFF
--- a/support-figma/extended-layout-plugin/src/localization-module.ts
+++ b/support-figma/extended-layout-plugin/src/localization-module.ts
@@ -23,6 +23,7 @@ const EXPLICIT_EXCLUSION_PLUGIN_DATA_KEY = "vsw-string-explicit-exclusion";
 const CONSOLE_TAG = `${Utils.CONSOLE_TAG}-LOCALIZATION`;
 const EXCLUDE_HASHTAG_NAME_OPTION = "excludeHashTagName";
 const EXCLUDE_CUSTOMIZATION_OPTION = "excludeCustomization";
+const GROUP_SAME_TEXT_OPTION = "groupSameText";
 
 interface StringResource {
   name: string;
@@ -142,7 +143,7 @@ async function localizeNodeAsync(
     } else if (clippyTextNodes.includes(node)) {
       Utils.log(CONSOLE_TAG, "Ignore client side customization:", node.name);
     } else {
-      await localizeTextNodeAsync(node, stringResourceMap);
+      await localizeTextNodeAsync(node, stringResourceMap, options);
     }
   } else {
     // Recurse into any children.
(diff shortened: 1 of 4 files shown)


support-figma/extended-layout-plugin/src/ui.html
```diff
@@ -1114,8 +1118,39 @@
...
+                  }
+                }, '*');
+                // Refresh the table, which recreate the whole table...
+                createOutputStringTable('outputStrings');
+                hasChanged = true;
+              }
+            }
+            if (!hasChanged) {
+              input.value = stringRes["name"];
+              showToast(`${newValue} already exists. Revert since duplicates are not allowed.`);
```
rylin8, line 1152:
```suggestion
              showToast(`${newValue} already exists. Reverting resource name since duplicates are not allowed.`);
```

rylin8: Would be nice to have an option to un-merge two nodes into different string resources, but that can be done later after we get some feedback.

support-figma/extended-layout-plugin/src/localization-module.ts
```diff
@@ -394,6 +398,21 @@ function normalizeString(characters: string): string {
...
+      // - non-breaking hyphen
+      .replace(/\u2011/g, "\\u2011")
+      // - figure dash
+      .replace(/\u2012/g, "\\u2012")
+      // - n dash
+      .replace(/\u2013/g, "\\u2013")
+      // - m dash
+      .replace(/\u2014/g, "\\u2014")
+      // - minus sign
+      .replace(/\u2212/g, "\\u2212")
```
rylin8, line 415: Where did you get this list of characters to replace?

rylin8 (commented):
Nice, this is a lot simpler than I expected. It would be nice to manually break out two identical strings into two translations in the UI, but that can be done later.

support-figma/extended-layout-plugin/src/localization-module.ts
yiqunw700: I found it on wiki. I used one of the dash occasionally in the styled text run test and Android Studio suggests to use unicode for those dashes when I checked the gen_strings.xml.

rylin8:
I think it might be worth it to add a test with two or more identical strings in Figma translating to different values. Maybe in a separate PR if you prefer.

rylin8 (approved):

Issue #1448, opened 2024-07-29
Update the () -> Brush function to use a State<Brush>


yiqunw700:
#1301 leftover

yiqunw700:
See https://github.com/google/automotive-design-compose/pull/1451#discussion_r1713298815

PR #1451, opened 2024-07-29
Replace BrushFunction customization with State<Brush>

yiqunw700:
Fixes: #1448


integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushTest.kt
```diff
@@ -89,28 +91,33 @@ fun CustomBrushTest() {
...
+        val shader = RuntimeShader(CUSTOM_SHADER)
+        val shaderBrush = SizingShaderBrush(shader)
+        // The kotlin compiler seems to get confused without semicolons on these statements.
+        shader.setFloatUniform("iTime", 0.0f)
+        brushState =
+            object : State<Brush> {
+                override val value: Brush
+                    get() =
+                        shaderBrush.also {
+                            // Only sample the state in the generator function; this means that we
```
rylin8, line 109: I don't recall what the "generator" function is, but I assume that basically when you access the value it calls this and updates the uniform.

iamralpht:
This is a bit different from the other changes, because the others were going from `@Composable () -> T` to `State<T>`, where this is going from `() -> T` to `State<T>`.

If the `Brush` impl itself depends on some `State` then when that state changes, a redraw (but not a recompose or relayout) is triggered. That works with both approaches and is about as efficient as you can be in Compose.

What's new in this change is you can also create a `mutableStateOf` a `Brush` and then change that (so you could change from a `SolidColor` brush to a `RadialGradient` brush, and (I think) trigger only a redraw because the state's value is fetched only in the rendering logic. In the old implementation, because it wasn't a `@Composable` function that returned the `Brush`, there was no way of changing just the brush type -- you had to make your own `Brush` subclass and make it do different things based on some `State` it always consults.

So I think this is a useful capability (sorry for being longwinded...), but I don't really like how the validation code came out. Would it be possible to use `remember mutableStateOf` there, rather than a custom `State` implementation?

support-figma/extended-layout-plugin/src/design-spec-module.ts
```diff
@@ -19,7 +19,7 @@ import * as Utils from "./utils";
 /** See: com.android.designcompose.codegen.CustomizationType */
 export enum DesignCustomizationKind {
   Text = "Text",
-  TextFunction = "TextFunction",
+  TextFunction = "TextState",
```
timothyfroehlich, line 22: Is this change ment to be here?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushTest.kt
```diff
@@ -89,28 +91,32 @@ fun CustomBrushTest() {
...
-
-    CustomBrushTestDoc.Main(fill = brush)
+    } else {
+        brushState =
+            object : State<Brush> {
+                override val value: Brush
+                    get() = SolidColor(Color.Blue)
+            }
+    }
+    CustomBrushTestDoc.Main(fill = brushState)
```
timothyfroehlich, line 121: I guess this is existing code, but it seems rather complicated for an example. Is this a common pattern for State objects?

timothyfroehlich (commented):
Mostly LGTM but Ralph or Rob should approve.

support-figma/extended-layout-plugin/src/design-spec-module.ts
yiqunw700: Done. Reverted.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushTest.kt
yiqunw700: Oh I didn't know.
It can create non mutable state just using `remember { SolidColor(Color.Blue) }`

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushTest.kt
yiqunw700: Yes.
It used to be: https://github.com/google/automotive-design-compose/blob/9ee3d5f24905ab093f6506106c6527e40da2837d/integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushTest.kt#L108

yiqunw700:
> This is a bit different from the other changes, because the others were going from `@Composable () -> T` to `State<T>`, where this is going from `() -> T` to `State<T>`.
> 
> If the `Brush` impl itself depends on some `State` then when that state changes, a redraw (but not a recompose or relayout) is triggered. That works with both approaches and is about as efficient as you can be in Compose.
> 
> What's new in this change is you can also create a `mutableStateOf` a `Brush` and then change that (so you could change from a `SolidColor` brush to a `RadialGradient` brush, and (I think) trigger only a redraw because the state's value is fetched only in the rendering logic. In the old implementation, because it wasn't a `@Composable` function that returned the `Brush`, there was no way of changing just the brush type -- you had to make your own `Brush` subclass and make it do different things based on some `State` it always consults.
> 
> So I think this is a useful capability (sorry for being longwinded...), but I don't really like how the validation code came out. Would it be possible to use `remember mutableStateOf` there, rather than a custom `State` implementation?

I added a few buttons that can switch to solid blue, a linear gradient, a radial gradient or a  shader.
But there is something I don't quite understand about using the `mutableStateOf` that why I have to create a new `SizingShaderBrush` to have it animate. While using the function or the State with overriding the getter, it can use the same `SizingShaderBrush` object.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushTest.kt
```diff
@@ -77,40 +93,82 @@ class SizingShaderBrush(private val shader: RuntimeShader) : ShaderBrush() {
...
+            shader.setFloatUniform("iTime", 0.0f)
+            LaunchedEffect(shaderEnabled.value) {
+                if (shaderEnabled.value) {
+                    snapshotFlow { movingValue.value }
+                        .collect {
+                            shader.setFloatUniform("iTime", it)
+                            // Check if shader has been disabled. Race condition when click on
+                            // other buttons.
+                            if (shaderEnabled.value) {
+                                // Need to create new object to trigger redraw.
```
rylin8, line 154: Is this triggering just a redraw, or is it also triggering recomposition since the `MutableState` `brushState` is changing?

yiqunw700: It only triggers redraw because the state value is only read during redraw.

When I click the buttons, https://github.com/google/automotive-design-compose/blob/9ee3d5f24905ab093f6506106c6527e40da2837d/designcompose/src/main/java/com/android/designcompose/DesignFrame.kt#L699 and https://github.com/google/automotive-design-compose/blob/9ee3d5f24905ab093f6506106c6527e40da2837d/designcompose/src/main/java/com/android/designcompose/FrameRender.kt#L273 will re-run, but **NOT**  https://github.com/google/automotive-design-compose/blob/9ee3d5f24905ab093f6506106c6527e40da2837d/designcompose/src/main/java/com/android/designcompose/DesignFrame.kt#L62

For squoosh same that only the `squooshRender` and `squooshTextRender` reruns but not `squooshRoot`

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushTest.kt
```diff
@@ -77,40 +93,82 @@ class SizingShaderBrush(private val shader: RuntimeShader) : ShaderBrush() {
...
-            shader.setFloatUniform("iTime", 0.0f);
-
-            {
-                // Only sample the state in the generator function; this means that we avoid
-                // recomposition and only do the redraw phase.
-                shader.setFloatUniform("iTime", movingValue.value)
-                shaderBrush
+            shader.setFloatUniform("iTime", 0.0f)
+            LaunchedEffect(shaderEnabled.value) {
+                if (shaderEnabled.value) {
```
rylin8, line 147: I like this better than what you have before, but it is still a little confusing because I am not familiar with `snapshotFlow`. I looked it up and it makes more sense. Can you add a comment here explaining what is going on? Other than that I am good with this, though @iamralpht should take another look.

yiqunw700: Done. Added a comment.
@iamralpht Please take another look and let me know what you think, thanks!

iamralpht: I'm still a bit confused about what is getting re-run when `movingValue.value` changes. In the previous version, it looks like it mutated `shaderBrush` (by giving it a new value for the `iTime` uniform) and returned it, and I think just that generator function would get re-run when `movingValue.value` changed (and then just Compose's "redraw" phase would run, skipping Composition and Layout).

Here, I think that there might be a bit more stuff that's getting run... but I'm not sure. I am guessing that you have to make a new instance of `SizingShaderBrush` because otherwise just mutating the `iTime` uniform doesn't change the identity of the instance already in `brushState.value` (so probably the `State<>` object just ignores assigning the value to the current value and doesn't propagate an invalidation).

So the usage of this seems more involved and complicated than before, and I think it might be less efficient than before, too. Right now, I would say we shouldn't make this change because the previous version was good in terms of efficiency (where the previous versions of the other things that didn't take `State<>` was quite inefficient and the move to `State<>` was a performance win).

yiqunw700: Right now when `moving.value changes` it has to create a new `shaderBrush` to trigger recomposition. So I agree it might be less efficient than before. I will close this pr. Thanks very much for the feedback!

PR #1462, opened 2024-07-31
Support animation when variables change values

rylin8:


rylin8:
Here's a start to variable animations.

crates/figma_import/src/variable_utils.rs
```diff
@@ -114,10 +114,10 @@ pub(crate) fn create_variable(v: &figma_schema::Variable) -> Variable {
             create_variable_helper(VariableType::Number, common, values_by_mode)
         }
         figma_schema::Variable::String { common, values_by_mode } => {
```
iamralpht, line 116: Do we need to merge these changes anyway? Looks like a bug...

rylin8: Yup that one is a bug. It's not breaking anything right now but I'll merge it in separately if this change doesn't end up going in.

PR #1466, opened 2024-08-01
Closes #1463: Support opacity animations

### LINKED ISSUE #1463: Support opacity transitions
When a node changes opacity, animate between the opacity values.

rylin8:
During layout animations, if the nodes change opacity, transition between them smoothly. Add a SmartAnimate test for this and a midpoint test as well.


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -462,6 +473,10 @@ private fun needsStyleTween(a: ViewStyle, b: ViewStyle): Boolean {
     return false
 }
 
+private fun needsOpacityTween(a: ViewStyle, b: ViewStyle): Boolean {
+    return a.node_style.opacity != b.node_style.opacity
```
yiqunw700, line 477: super nit: should we compare opacity.getElse(1f)?

rylin8: Good idea, that would avoid animations between opacity 1 and opacity Optional.NULL

kovmarci86: > super nit: should we compare opacity.getElse(1f)?

+1, because `opacity.getOrElse { 1F }` is used for when `needsOpacityTween==true`.

rylin8: done

yiqunw700 (approved):

PR #1471, opened 2024-08-02
Closes #1470: Fix bug where renamed component instances don't animate

### LINKED ISSUE #1470: Renamed component instances don't animate
When a component instance is renamed from its original name derived from the component set, animations don't work when the variant changes.

rylin8:
Get the variant parent name from component_info, and restore other fixes missed when fixing #1056.


rylin8:
This basically takes fixes from https://github.com/google/automotive-design-compose/pull/1076 and copies them in areas of squoosh that I missed originally.

yiqunw700 (approved):

PR #1480, opened 2024-08-05
Remove Rust/Cargo from Dependabot

timothyfroehlich:
We need to begin to better control our Cargo dependencies and step one is to stop automatically updating them


yiqunw700 (approved):

kovmarci86 (approved):
LGTM

Issue #1449, opened 2024-07-29
Hyperlink support in text for non-squoosh renderer


yiqunw700:
#1217 leftover

PR #1494, opened 2024-08-05
Add support for url in text in DesignText

yiqunw700:
Ticket: #1449
No squoosh support in this pr. I find it a little bit hard to do the url annotation there so I might find some other time to look into that.

Also sort the examples in the validation so it will be easier to find a test.


rylin8:
I didn't know this was something we wanted to do. Can you make a ticket for it?

designcompose/src/main/java/com/android/designcompose/DesignText.kt
```diff
@@ -364,6 +384,29 @@ internal fun DesignText(
                 )
             } else {
                 var textLayoutResult: TextLayoutResult? = null
+                val uriHandler = LocalUriHandler.current
+                if (annotatedText.getUrlAnnotations(0, annotatedText.text.length).isNotEmpty()) {
+                    val onClick: (Int) -> Unit = { offset ->
+                        annotatedText
+                            .getUrlAnnotations(start = offset, end = offset)
```
rylin8, line 391: So if the hyperlink is at (10, 20) and you click at position 15, `annotatedText.getUrlAnnotation(15, 15)` will return the annotation with the hyperlink?

yiqunw700:
> I didn't know this was something we wanted to do. Can you make a ticket for it?

Yes. Ticket https://github.com/google/automotive-design-compose/issues/1449

The existing example is the tutorial app's welcome page where it has a url link.

designcompose/src/main/java/com/android/designcompose/DesignText.kt
yiqunw700: Yes. That is right.

rylin8 (approved):

PR #1496, opened 2024-08-06
Closes #1493: Support progress bar and marker constraints

### LINKED ISSUE #1493: Progress bars don't respond to layout changes
I made a design with some progress bars that get wider based on the layout (the parent fills its parent, the ultimate parent is resizable). The progress bar uses the width from the design doc (rather than the width from the parent layout) to determine its size -- so as the progress bar gets resized, the filled proportion doesn't change.

I observed this using squoosh, but I think the code path is the same for non-squoosh, too.

rylin8:
Support layout changes with progress bars and markers in squoosh. When constraints are set on a progress node, use the parent layout to calculate the extents of the progress bar/marker. This is only supported on squoosh since we have access to the parent layout info there. Additional tests have been added for this.


rylin8:
This implementation is far from perfect but it does allow progress bars to stretch with layout.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -154,6 +156,7 @@ private fun calculateProgressBarData(
     progressBarData: ProgressBarMeterData,
     meterValue: Float,
     style: ViewStyle,
+    parent: SquooshResolvedNode?,
```
iamralpht, line 159: I was thinking it's a bit cleaner to just pass `parentLayout: Layout?` in both of these, but you're consulting style to see if the layout is fixed, so you wouldn't be able to do that then.

Do you need to check if it's fixed, or could you always use `parentLayout` if it's not null (understanding that it is null on DC Classic)?

iamralpht (approved):
Thanks for picking this up! Rolling it into my demo build now...

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: That's exactly what I did originally, but it breaks in squoosh on progress bars that are skewed or rotated with a fixed size (no constraints to stretch). In that scenario, we want to use the non-rotated width/height, which is saved in the plugin data. Using the width/height from `computedLayout` yields a smaller value and renders incorrectly. However if the progress bar is modified in Figma to stretch (constraints top/bottom or left/right), then `computedLayout` has the correct size that we want to use. I'll need to dig a bit to figure out why exactly layout is computed this way, I don't recall exactly and am not sure if it is different in squoosh.

rylin8: This is because of the complicated nature of how we determine the size of a node when rendering, with most of the logic in `getNodeRenderSize()`. To make this more robust I'll try to re-use this function instead of defining additional logic here.

rylin8: Done

PR #1502, opened 2024-08-10
Closes #1501: Support TapCallback and key injection in squoosh

### LINKED ISSUE #1501: Support TapCallback customizations in squoosh
From Ralph: "There's logic in resolveVariantsRecursively to make an event handling composable when there's an interaction. It probably needs to test if there's a TapCallback and emit a Composable in that case too."

rylin8:


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -259,6 +262,11 @@ fun SquooshRoot(
     val rootLayoutId = remember(docId) { SquooshLayout.getNextLayoutId() * 100000000 }
     val layoutCache = remember(docId) { HashMap<Int, Int>() }
     val layoutValueCache = remember(docId) { HashMap<Int, Layout>() }
+    val keyEventTracker = remember(docId) { KeyEventTracker() }
```
iamralpht, line 265: Do we get in trouble if we have multiple `DesignRoot`s showing components from the same doc id?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshTreeBuilder.kt
```diff
@@ -289,8 +295,23 @@ internal fun resolveVariantsRecursively(
...
+                val keyEvent = DesignKeyEvent.fromJsKeyCodes(keyTrigger.key_codes)
+                val keyAction =
+                    KeyAction(
+                        interactionState,
+                        r.action,
+                        findTargetInstanceId(document, parentComps, r.action),
+                        customizations.getKey(),
+                        null
+                    )
+                keyTracker.addListener(keyEvent, keyAction)
```
iamralpht, line 310: I think because we build a new tree every iteration (or two new trees if there's an animation) this will cause the list of listeners to get huge. Another way to do it might be to have one listener in SquooshRoot that iterates the current SquooshResolvedNode tree and sees if any node has a key event handler. OR make resolveVariantsRecursively generate a list of `(keyEvent, keyAction)`, and the SquooshRoot key event handler iterates the most recent version of that list?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshInteraction.kt
```diff
@@ -124,6 +124,8 @@ internal fun Modifier.squooshInteraction(
                                 )
                             }
                         }
+                        // Invoke the tap callback customization if one exists on this node
+                        customizations.getTapCallback(node.view.name)?.invoke()
```
iamralpht, line 128: oh nice - glad this was straightforward!

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
rylin8: Probably -- I could add the `rootNodeQuery` to the `remember`.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshTreeBuilder.kt
rylin8: That's what I basically did originally. I had a global list and cleared out the list every iteration and built a new one in the `resolveVariantsRecursively`. However, since it was global the design switcher's root node cleared out everything when it shouldn't have. Then I made an instance of `KeyTracker` per document. I'll make some improvements here.

I don't think this list will get huge because it stores using a `HashSet`, but I don't think nodes that change under root and no longer have key event interactions will update. I'll fix that.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -259,6 +262,11 @@ fun SquooshRoot(
     val rootLayoutId = remember(docId) { SquooshLayout.getNextLayoutId() * 100000000 }
     val layoutCache = remember(docId) { HashMap<Int, Int>() }
     val layoutValueCache = remember(docId) { HashMap<Int, Layout>() }
+    val keyEventTracker = remember(docId, rootNodeQuery) { KeyEventTracker() }
```
iamralpht, line 265: 👍

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -337,6 +347,7 @@ fun SquooshRoot(
                 customizationContext,
                 variantTransitions,
                 transitionedInteractionState ?: interactionState,
+                keyEventTracker,
```
iamralpht, line 350: I think this means that during an animated transition, both sets of key events will be available and processed. I think maybe we should only have the target set (but don't think we need to make the change now)

iamralpht (approved):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
rylin8: Good point, added https://github.com/google/automotive-design-compose/issues/1514

PR #1504, opened 2024-08-12
Fix #1503: ensure child composables are drawn, by using Layout ID rather than SquooshResolvedNode identity.

### LINKED ISSUE #1503: Child composables don't draw during animations and transitions
Child composables are not rendered during animations and transitions, causing unfortunate flickering.

iamralpht:
The previous check (that the SquooshResolvedNodes were equal) failed, because during a transition we generate a new set of SquooshResolvedNodes from the source tree and dest tree. With this change, we consult the layout ID, which is unique within a tree (so still useful for this purpose of identifying the selected node for child rendering).

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
+++ b/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
@@ -687,7 +687,10 @@ fun SquooshRoot(
                 for (child in childComposables) {
                     var composableChildModifier =
                         Modifier.drawWithContent {
-                                if (child.node == childRenderSelector.selectedRenderChild)
+                                if (
+                                    child.node.layoutId ==
+                                        childRenderSelector.selectedRenderChild?.layoutId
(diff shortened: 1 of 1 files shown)


iamralpht:
@rylin8 PTAL - I want to add tests to this; sharing now because I needed it to make a demo work.

rylin8:
So replacement content, components, and nodes with interactions don't render with animations?

iamralpht:
> So replacement content, components, and nodes with interactions don't render with animations?

Interactions work fine during animation, because those are all Composables that never draw anything at all. Replacement content that draws content (i.e.: all of it) was getting hidden during animations.

rylin8:
LGTM. Let me know if you plan on adding tests to this PR or separately -- or if you'd like me to.

rylin8 (approved):
Let's get this change in since it's good. I filed a bug with animations separately, and will add tests in that change https://github.com/google/automotive-design-compose/issues/1539

PR #1505, opened 2024-08-12
Replace FontStretch

yiqunw700:


crates/dc_bundle/src/definition/element.rs
```diff
@@ -28,3 +30,43 @@ impl FontFeature {
         FontFeature { tag, enabled: true }
     }
 }
+
+impl PartialEq for FontStretch {
```
yiqunw700, line 34: @timothyfroehlich This one it implements the ParticalEq and Default which confilcts with the generated FontStretch struct with the `#[derive(Clone, PartialEq, ::prost::Message)]` annotation. Do you know if there is a way to remove the `PartialEq` and `::prost::Message` from the generated code?

timothyfroehlich: We can't remove ::prost::Message, we need that to serialize/deserialize messages. For PartialEq I'm not sure if we can remove that, can you see if the PartialEq impl that we have is similar to the generated one? And for Default, yeah, we can't override it. We just need to make sure that any place that we were initializing the structure from scratch that we set the default values there. For most of these I don't think we do initialize them from scratch. I think generally you can try to ignore the Default impls and just run the test suite and see if anything changes.

PR #1513, opened 2024-08-14
Fix the string resource mismatch

yiqunw700:
* Also fixes the node id overflow issue in the plugin ui.
* Also add the listOptions back in the plugin api call getStyledTextSegments

It is not quite possible to make the plugin to match the REST api behavior so in this change, it will check if styles are the same and merge when transforming.

Fixes: #1518

### DIFF
--- a/crates/figma_import/src/transform_flexbox.rs
+++ b/crates/figma_import/src/transform_flexbox.rs
@@ -1294,6 +1294,11 @@ fn visit_node(
                 } else {
                     style.node_style.font_size.clone()
                 };
+                let font_family = if sub_style.font_family.is_some() {
+                    sub_style.font_family.clone()
+                } else {
+                    style.node_style.font_family.clone()
+                };
                 let font_weight = if let Some(fw) = sub_style.font_weight {
                     dc_bundle::legacy_definition::element::font::FontWeight(NumOrVar::Num(fw))
                 } else {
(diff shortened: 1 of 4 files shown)


crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1325,29 +1330,47 @@ fn visit_node(
...
+                    ),
+                    hyperlink,
+                };
+                let mut has_handled = false;
+                if runs.len() > 0 {
+                    let last_run = runs.get(runs.len() - 1);
+                    if last_run.is_some() {
+                        let last_run_style = last_run.unwrap().style.clone();
+                        if last_run_style == style {
+                            error!("the two styles are the same.");
```
rylin8, line 1359: You mentioned warning the user that there might be something wrong with the text field. How would they see the error? Just in logcat?

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1325,29 +1330,47 @@ fn visit_node(
...
+                            let to_be_merged = runs.pop();
+                            let mut new_run = String::new();
+                            if to_be_merged.is_some() {
```
rylin8, line 1362: Normally we try to avoid `unwrap()`, so we'd do something like `if let Some(tbm) = to_be_merged { push(&tbm.text); }`

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1325,29 +1330,47 @@ fn visit_node(
...
+                            .opentype_flags
+                            .clone()
+                            .unwrap_or(text_style.opentype_flags.clone()),
+                    ),
+                    hyperlink,
+                };
+                let mut has_handled = false;
+                if runs.len() > 0 {
+                    let last_run = runs.get(runs.len() - 1);
+                    if last_run.is_some() {
```
rylin8, line 1356: Here also -- `if let Some(lr) = last_run { }`

crates/figma_import/src/transform_flexbox.rs
yiqunw700: Yeah, for the developer to debug why the translation isn't working. So I think logcat should be good.

Is there any other ways of showing errors to user? Using jni to show errors in the design switcher since it is the rust code?

crates/figma_import/src/transform_flexbox.rs
yiqunw700: Done.

crates/figma_import/src/transform_flexbox.rs
yiqunw700: Done.

crates/figma_import/src/transform_flexbox.rs
rylin8: I think this Rust code is called during `jniFetchDoc`, which is where it fetches everything and returns a giant blob of  the server response. So to have it show an error in the design switcher we'd have to modify the server response with some errors I think.

Is it possible to do this merging in the plugin, so that we can show a popup message and ask the user if they want to merge, or give them the option to fix it themselves?

yiqunw700: No, I know the plugin is the best place to do a fix but this really is an edge case where the sub-style is the same with the node style. Plugin API doesn't give me this info that it's different, then there's no way for me to do any comparison because they are the same style.

So we can either merge when we transform or just leave it unmerged with only some warnings or errors in logs.

Or we can try to file a bug for figma to make REST and plugin APIs be consistent.

rylin8: Ok, so without this fix, the developer will see an untranslated string if that string has this strange property from Figma. There's not really a good way for them to figure out why that string is untranslated, except possibly a logcat error. My inclination is to keep this fix, but also add a bit to the localization documentation explaining that we will merge strings if Figma presents them as a list of strings but with the same style. @iamralpht @babakbo please chime in if you have other thoughts.

rylin8: We discussed this in our last meeting and decided to merge this change, but also make a ticket to improve error handling. We would like to be able to propagate an error from Rust, where it parses the Figma data, into the dcf file so that DC can show the error in the DesignSwitcher or in any other way that the app wishes. @yiqunw700 can you make a ticket for this?

rylin8 (approved):

crates/figma_import/src/transform_flexbox.rs
yiqunw700: Done. #1544

PR #1515, opened 2024-08-15
Closes #1512: Fix gradient rendering on transformed nodes in squoosh

### LINKED ISSUE #1512: Gradients in squoosh are rendered differently than non-squoosh gradients.
This might only apply to nodes that are transformed. In this example, the skewed progress bar is rendered correctly when squoosh is off, but the colors are muted when squoosh is on.
No squoosh:
Screenshot 2024-08-14 at 11 39 33 AM

Squoosh:
Screenshot 2024-08-14 at 11 39 40 AM

rylin8:


timothyfroehlich (approved):

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -772,7 +772,8 @@ internal fun squooshShapeRender(
                 val b = background.asBrush(document, density, variableState)
                 if (b != null) {
                     val (brush, fillOpacity) = b
-                    brush.applyTo(size, p, fillOpacity)
+                    val brushSize = getNodeRenderSize(rectSize, size, style, node.layoutId, density)
+                    brush.applyTo(brushSize, p, fillOpacity)
```
yiqunw700, line 776: I checked the https://github.com/google/automotive-design-compose/blob/adda35bc074a32f748f3b90f5e7c5a169d4ad205/designcompose/src/main/java/com/android/designcompose/FrameRender.kt#L316
There is one more for stokeBrush below at line 792. Do we need to update that one too?

Why the customBrush at line 767 doesn't need to do this?

rylin8: You're right, it probably should be there. Sorry, I was rushing to get this in to try and make the demo look better. I don't think we ever noticed this, or the one at 767 because it involves either making a thick stroke with a gradient on a transformed node, or a custom brush on a transformed node. We don't have a lot of tests involving rotated/skewed nodes, and we still have some bugs in them. I made a ticket to address these. Thanks! https://github.com/google/automotive-design-compose/issues/1516

PR #1521, opened 2024-08-16
Migrate ViewShape/Style/Data and View to dc_bundle

timothyfroehlich:
Removes the now-empty toolkit_schema

Fixes Migrate ViewStyle/View to dc_bundle #1456
Fixes Migrate ViewShape/Data to dc_bundle #1455


crates/figma_import/src/lib.rs
```diff
@@ -42,11 +41,13 @@ pub use document::Document;
 pub use error::Error;
 pub use fetch::{fetch_doc, ConvertRequest, ConvertResponse, ProxyConfig};
 pub use image_context::ImageContextSession;
-pub use toolkit_schema::{View, ViewData}; // ugly hack
-                                          // Internal convenience
+// ugly hack
```
rylin8, line 44: Not sure what part exactly was the ugly hack but I think it can be removed.

rylin8 (approved):

PR #1522, opened 2024-08-19
Improve performance by caching text layouts and computed paths.

iamralpht:
Profiling showed that computing text layouts and computing paths was taking the bulk of runtime in a simple demo app. This change adds two generational caches -- one for computed text layout (e.g.: the Compose `ParagraphIntrinsics`), and one for computed paths (the `ComputedPath` type which contains all of the paths needed to render a node).

Additionally some unused logging has been removed.

In a profile of "FutureV2" design, `squooshComputeTextInfo` run time was reduced by a half, and `computePaths` was reduced by around 5x.

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
+++ b/designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
@@ -27,7 +27,6 @@ import androidx.compose.ui.graphics.asComposePath
 import androidx.core.graphics.minus
 import androidx.core.graphics.plus
 import com.android.designcompose.serdegen.BoxShadow
-import com.android.designcompose.serdegen.NumOrVar
 import com.android.designcompose.serdegen.StrokeAlign
 import com.android.designcompose.serdegen.StrokeCap
 import com.android.designcompose.serdegen.ViewShape
@@ -44,7 +43,7 @@ import com.android.designcompose.serdegen.ViewStyle
 /// In some special cases, like frames and rectangles, a simpler algorithm is used
 /// because the geometry is trivial to calculate. This results in simpler paths to
 /// render, and may be less compute intensive.
-class ComputedPaths(
+internal class ComputedPaths(
     /// The paths that need to be filled with the Background brush for
     /// this shape.
     val fills: List<Path>,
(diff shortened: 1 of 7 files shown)


iamralpht:
I'll mark this as Draft once CI has done it's checking. I'd like to figure out how to use the benchmark infrastructure to baseline performance on an instrument cluster sample, so we can avoid unintentional regressions.

iamralpht:
I don't know why this changed the gradient output, but it appears to be correct now, when it wasn't before. Did the reference images get out of sync?

rylin8:
The gradient images were fixed with this change last week, with the correct images being checked in. https://github.com/google/automotive-design-compose/pull/1515 @timothyfroehlich do you know why the image diff is showing up here?

timothyfroehlich:
s

> The gradient images were fixed with this change last week, with the correct images being checked in. #1515 @timothyfroehlich do you know why the image diff is showing up here?

Not sure, maybe something got missed. Let me try it on my system

designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
```diff
@@ -69,26 +69,78 @@ class ComputedPaths(
...
+    if (
+        cacheEntry != null &&
+            cacheEntry.viewShape == this &&
+            cacheEntry.style == style &&
+            cacheEntry.density == density &&
+            cacheEntry.frameSize == frameSize &&
+            cacheEntry.overrideSize == overrideSize &&
+            cacheEntry.customArcAngle == customArcAngle
+    ) {
+        // XXX: VariableState is never equal, even when the variables haven't changed.
```
rylin8, line 132: I remember having some issues with creation of VariableState, but I think it could be improved by putting it in a remember() with all the CompositionLocal objects that can cause variables to change.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -703,10 +703,9 @@ internal fun squooshShapeRender(
             // layout doesn't actually consider rotation yet.
             rectSize,
             customArcAngle,
-            vectorScaleX,
-            vectorScaleY,
-            0, // XXX: layoutId
+            node.layoutId,
```
rylin8, line 706: Does this change fix or change anything? Why was it 0 before?

designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
```diff
@@ -69,26 +69,78 @@ class ComputedPaths(
...
+
+    fun get(layoutId: Int): Entry? {
+        return cache[layoutId]
+    }
+
+    fun put(layoutId: Int, data: Entry) {
+        nextGeneration[layoutId] = data
+    }
+
+    fun collect() {
```
rylin8, line 103: Can you add a comment explaining why you need to have a current, nextGeneration, and this `collect()` function?

designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
```diff
@@ -69,26 +69,78 @@ class ComputedPaths(
...
+        cacheEntry != null &&
+            cacheEntry.viewShape == this &&
+            cacheEntry.style == style &&
+            cacheEntry.density == density &&
+            cacheEntry.frameSize == frameSize &&
+            cacheEntry.overrideSize == overrideSize &&
+            cacheEntry.customArcAngle == customArcAngle
+    ) {
+        // XXX: VariableState is never equal, even when the variables haven't changed.
+        pathCache.put(layoutId, cacheEntry)
```
rylin8, line 133: Is this needed because you call `collect()` at the start of each draw cycle? It's a bit confusing to me -- maybe if this were called something like `pathCache.refreshEntry(layoutId)`?

timothyfroehlich:
> s
> 
> > The gradient images were fixed with this change last week, with the correct images being checked in. #1515 @timothyfroehlich do you know why the image diff is showing up here?
> 
> Not sure, maybe something got missed. Let me try it on my system

#[1536](https://github.com/google/automotive-design-compose/pull/1536)

timothyfroehlich:
> > s
> > > The gradient images were fixed with this change last week, with the correct images being checked in. #1515 @timothyfroehlich do you know why the image diff is showing up here?
> > 
> > 
> > Not sure, maybe something got missed. Let me try it on my system
> 
> #[1536](https://github.com/google/automotive-design-compose/pull/1536)

Only difference is a difference in Squoosh, so Ralph I'd check to make sure you've rebased properly.

iamralpht:
> Only difference is a difference in Squoosh, so Ralph I'd check to make sure you've rebased properly.

Oh, I bet this is because I'm targeting another user branch, and not main, and did some rebasing right after pushing this branch...

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
iamralpht: It isn't used in Squoosh, and it wasn't easily accessible originally. Rendering uses the layout id to look up if there's a custom size applied to a node, which Squoosh doesn't support. So this doesn't change existing behavior (LayoutManager's custom size hashmap is still empty), but the layout ID is stable enough to use as a key in the (new) shape cache.

designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
iamralpht: Yeah, every time a cache entry is used it needs to be `put`, so that it doesn't get collected.

designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
iamralpht: Yep, added some comments, hopefully that's better (I didn't change the method names, but I could do that, too).

rylin8:
Since you are calling `collect()` every render cycle, does this cache only help when text layout is computed multiple times in one render cycle, or a path is computed multiple times in one render cycle? When does that happen?

iamralpht:
> Since you are calling `collect()` every render cycle, does this cache only help when text layout is computed multiple times in one render cycle, or a path is computed multiple times in one render cycle? When does that happen?

No, collect is to retain the stuff that was used in the current render or recompose cycle, and throw out the stuff that was used in previous ones but not the current one. The only case where we use stuff twice in one cycle is when we're calling `resolveVariantsRecursively` for both the base tree and the target tree -- then we'll use text twice, and that's why text inserts into `cache` and `nextGeneration`. I'll see about updating comments some more :)

rylin8:
> > Since you are calling `collect()` every render cycle, does this cache only help when text layout is computed multiple times in one render cycle, or a path is computed multiple times in one render cycle? When does that happen?
> 
> No, collect is to retain the stuff that was used in the current render or recompose cycle, and throw out the stuff that was used in previous ones but not the current one. The only case where we use stuff twice in one cycle is when we're calling `resolveVariantsRecursively` for both the base tree and the target tree -- then we'll use text twice, and that's why text inserts into `cache` and `nextGeneration`. I'll see about updating comments some more :)

Oh right, got it. You're just throwing away any cached items that weren't used in the last render cycle, so that cache is like an LRU that only remembers one time.

rylin8 (approved):

designcompose/src/main/java/com/android/designcompose/ComputePaths.kt
iamralpht: I've got an update coming for this bit -- I moved extraction of corner radius from variables up, and just include the computed border radii in the cache entry (then we get all of the factors for the generated path in the cache entry).

iamralpht:
Updated to include the corner radius (after resolving tokens) as a consideration in the path cache. This means token value changes for tokens bound to corner radius will appropriately result in not reusing an old path.

PR #1548, opened 2024-08-21
Closes #1516: Fix stroke brush size

### LINKED ISSUE #1516: Fix additional cases of applying a brush to the wrong size
Sorry late here but:
I checked the https://github.com/google/automotive-design-compose/blob/adda35bc074a32f748f3b90f5e7c5a169d4ad205/designcompose/src/main/java/com/android/designcompose/FrameRender.kt#L316
There is one more for stokeBrush below at line 792. Do we need to update that one too?

Why the customBrush at line 767 doesn't need to do this?

_Originally posted by @yiqunw700 in https://github.com/google/automotive-design-compose/pull/1515#discussion_r1718849028_

rylin8:
This change changes strokes and custom brushes to use the correct node size when applying the brush.


rylin8:
I noticed that strokes with no width or height, like a line, but with a stroke thickness, render incorrectly. Filed https://github.com/google/automotive-design-compose/issues/1549 for this.

yiqunw700 (approved):

PR #1551, opened 2024-08-22
Closes #1550: Support animation delays

### LINKED ISSUE #1550: Add support for animation delays
When specifying a custom animation, allow the caller to specify a delay in milliseconds that the animation should wait before beginning.

rylin8:


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -116,6 +116,7 @@ internal class SquooshAnimationRenderingInfo(
     val action: AnimatedAction?,
     val variant: VariantAnimationInfo?,
     var startTimeNanos: Long = 0L,
+    var delayTimeMs: Int = 0,
```
iamralpht, line 119: I don't like that we're mixing units and that the time units are in a basic type (rather than a sum type for each unit)... but [Compose does the same thing](https://developer.android.com/reference/kotlin/androidx/compose/animation/core/TweenSpec#TweenSpec(kotlin.Int,kotlin.Int,androidx.compose.animation.core.Easing)), and they don't even include the unit in the name of the `delay` argument. Compose does seem consistent in using `whateverMillis` when they do include the unit in the argument name, so maybe we should do that instead of `Ms` (*cries in Rust std::time::Duration*, but only for a few millis, or ms...).

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshVariantTransition.kt
```diff
@@ -59,12 +59,21 @@ import com.android.designcompose.serdegen.View
 
 interface AnimationTransition {
     fun animationSpec(): AnimationSpec<Float>
+
+    fun delayMs(): Int
 }
 
-class SmartAnimateTransition(val transition: AnimationSpec<Float>) : AnimationTransition {
+class SmartAnimateTransition(
```
iamralpht, line 66: Since this is public API it should have some documentation.

iamralpht:
Logic looks good; minor feedback on API stuff.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
rylin8: Done

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshVariantTransition.kt
rylin8: Added some comments and edited the animation documentation

iamralpht (approved):

Issue #1469, opened 2024-08-02
Manually break out two identical strings into two translations in the UI


yiqunw700:
1.               Nice, this is a lot simpler than I expected. It would be nice to manually break out two identical strings into two translations in the UI, but that can be done later.
_Originally posted by @rylin8 in https://github.com/google/automotive-design-compose/pull/1445#pullrequestreview-2214074591_

2.               I think it might be worth it to add a test with two or more identical strings in Figma translating to different values. Maybe in a separate PR if you prefer.
_Originally posted by @rylin8 in https://github.com/google/automotive-design-compose/issues/1445#issuecomment-2264114706_
            
            
---------------
A follow up ticket of https://github.com/google/automotive-design-compose/issues/1242

yiqunw700:
Added the support to break identical strings into two translations. But action is not saved. Will need a follow up ticket for saving the actions.

PR #1557, opened 2024-08-26
Allow breaking out strings into two translations

yiqunw700:
Add an icon button in the UI for this function. (Couldn't make the icon-button in the same line with the checkbox and the node id though...).

Action is not saved... Will look into it in a following pr.
Towards #1469

### DIFF
--- a/support-figma/extended-layout-plugin/src/localization-module.ts
+++ b/support-figma/extended-layout-plugin/src/localization-module.ts
@@ -107,10 +107,10 @@ export async function generateLocalizationData(
   }
 
   // Convert to an array of key-value pairs
-  const stringReourceArray = Array.from(stringResourceMap);
+  const stringResourceArray = Array.from(stringResourceMap);
   figma.ui.postMessage({
     msg: "localization-output",
-    output: stringReourceArray,
+    output: stringResourceArray,
   });
 }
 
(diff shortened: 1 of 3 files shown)


support-figma/extended-layout-plugin/src/localization-module.ts
```diff
@@ -134,6 +134,77 @@ export async function updateStringRes(strRes: StringResource) {
...
+        translatable: true,
+        text: strRes.text,
+        textNodes: [{ nodeId: node.id, isExcluded: isNodeExcluded }],
+        extras: strRes.extras,
+        textLength: strRes.textLength,
+      };
+      stringResourceMap.set(stringResName, newStrRes);
+
+      // Convert to an array of key-value pairs
+      const stringReourceArray = Array.from(stringResourceMap);
```
rylin8, line 189: speling: 'stringReourceArray'

support-figma/extended-layout-plugin/src/ui.html
```diff
@@ -937,8 +938,32 @@
...
+      editBtn.onclick = () => {
+        parent.postMessage({
+          pluginMessage: {
+            msg: 'localization-ungroup-node',
+            node: node["nodeId"],
+            stringResMap: Array.from(outputStringData)
+          }
+        }, '*');
+      };
+      block.append(editBtn);
```
rylin8, line 964: When I mouse over this button, it changes to a hand icon for 1-2 pixels on the perimeter of the button, then back to an arrow in the center. I think it's changing to the hand when it's over the 'icon-button' but not over the 'icon--break' icon. Can you fix this so it's either the arrow or the hand icon the whole time? I also think a short tooltip would help, even though there are instructions at the top.

rylin8 (commented):
What's the best way to "Undo" this action, or to combine similar text nodes to the same string resource label?

support-figma/extended-layout-plugin/src/localization-module.ts
yiqunw700: Done.

support-figma/extended-layout-plugin/src/ui.html
yiqunw700: Done.

yiqunw700:
> What's the best way to "Undo" this action, or to combine similar text nodes to the same string resource label?

Yes. To undo it, rename the resource label to the one it would like to merge with.

rylin8 (approved):

PR #1577, opened 2024-08-28
Add new .idea file to gitignore - emulatorDisplays

timothyfroehlich:


timothyfroehlich:
I've got my upload script set to mark them as WIP by default, but yeah, it was good to go.

PR #1578, opened 2024-08-28
Update versions to v0.31

timothyfroehlich:


Cargo.lock
```diff
@@ -317,7 +317,7 @@ dependencies = [
 
 [[package]]
 name = "dc_jni"
-version = "0.30.0"
+version = "0.31.0"
```
yiqunw700, line 320: Do the four packages always have the same version? Is it possible to create a global version?

yiqunw700 (approved):

Cargo.lock
timothyfroehlich: That's just how Cargo.lock does it, it's a generated file and is very verbose. I didn't manually change those values, the Cargo.lock file just updated once I'd changed it in the workspace Cargo.toml.

PR #1585, opened 2024-08-30
Fixes #1546: Change arc rotations to align with node rotations

### LINKED ISSUE #1546: Matching a "rotation / needle" dial with an "arc path" gauge is tricky
I recently made a bunch of dials and gauges that combined an arc gauge with a "rotation / needle" dial (the needle was bright and the arc trailed behind). I found that I used completely different angle values for the "rotation / needle" configuration than I used for the arc.

I was expecting that I'd just be able to use the same numbers in both things and get dials and gauges that point in the same direction. I made my needles by starting with a horizontal frame that spanned the whole width of the gauge, but perhaps I should have started with a vertical bar? I'm not sure and it's not super clear...

Internal reference doc is here: https://www.figma.com/design/4C3C3GlbQfJk1acq8YjoJe/Circular?node-id=1-2&t=OBD2mRpSPzObniNB-1

rylin8:
Rotate arcs in the same way that nodes are rotated, going counterclockwise instead of clockwise as it is typically done in a standard 2d Cartesian coordinate system.

Fixes: #1546


yiqunw700:
Is this still WIP or ready for review?

rylin8:
Ready now. This change will render arcs setup as dials/gauges differently than before if you have not updated your Figma design with the plugin. I was trying to think about keeping backward compatibility here, but it would make the code a bit messier.

yiqunw700:
I think `Fixes: #1546` in the description will close the issue. Now you see there is a `linked a pull request now that will close this issue` while in the title is just says `referenced  this issue`.
`Fix`, `Closes` should both work. But only when in description.

yiqunw700 (approved):

PR #1594, opened 2024-09-03
Modify the fetch tool to load the Figma token from command line, env var and .config file, in that order.

timothyfroehlich:


yiqunw700 (approved):

yiqunw700:
LGTM.
Just want to double check that with this change, it will read the api key from command line like `--api_key=` or env variable `FIGMA_API_KEY` or read it from `~/.config/figma_access_token` which only contains the api key. Right?

timothyfroehlich:
Yep, so you can set it using any of those methods.

PR #1598, opened 2024-09-05
Convert Dimension and DimensionRect to Proto

timothyfroehlich:
Dimension and DimensionRect were a big pain, because Dimension is a typed enum and proto doesn't handle those very well.

On the Rust side I basically preserved the original Dimension and DimensionRect as proxies for the Protobuf versions. The proxies offer much cleaner code than using the proto types directly and convert to and from the proto types easily. 

I did end up with a number of unrelated changes from running cargo fix, which we should really be running more often. (I should add it to the format script). 

I didn't do the same on the Kotlin side because there is much less manipulation of the types, but the same could be done as well.

The resulting code is uglier than I'd like, but we need to get this finished and move on.

# Based on PR #1634

## Full chain of PRs as of 2024-09-19

* PR #1598:
  `wb/froeht/1575-ProtoConv-Convert-Rectangle` ➔ `wb/froeht/fix-robo-test-recording`
* PR #1634:
  `wb/froeht/fix-robo-test-recording` ➔ `main`


designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -48,55 +48,19 @@ import androidx.compose.ui.graphics.isIdentity
...
 import kotlin.math.sqrt
 
 /** Multiply out a dimension against available space */
-internal fun Dimension.resolve(available: Int, density: Float): Int? {
-    return when (this) {
-        is Dimension.Percent -> (available * value).roundToInt()
-        is Dimension.Points -> (value * density).roundToInt()
+internal fun DimensionProto.resolve(available: Int, density: Float): Int? {
+    return when (this.get()) {
+        is Dimension.Percent -> (available * getValue()!!).roundToInt()
```
rylin8, line 62: I'd prefer a way to not need `!!`

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -127,14 +91,14 @@ internal fun absoluteLayout(style: ViewStyle, constraints: Constraints, density:
             0
         }
 
-    val left = style.layout_style.left.resolve(pw, density)
-    val top = style.layout_style.top.resolve(ph, density)
+    val left = style.layout_style.left.get().resolve(pw, density)
```
rylin8, line 94: I'm confused because I thought `get()` would return `Dimension`, but you changed `resolve()` to be a function on `DimentionProto`, not `Dimension`.

designcompose/src/main/java/com/android/designcompose/proto/DimensionProto.kt
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package com.android.designcompose.proto
+
+import com.android.designcompose.serdegen.Dimension
+import com.android.designcompose.serdegen.DimensionProto
+import java.util.*
+
+fun DimensionProto.get(): Dimension = dimension.orElseThrow { NoSuchFieldException("Malformed data: DimensionProto's dimension unset") }
```
rylin8, line 23: What is `dimension`? I looked at the generated `DimensionProto` file but didn't see that as a member variable.

designcompose/src/main/java/com/android/designcompose/proto/DimensionProto.kt
```diff
@@ -0,0 +1,36 @@
...
+
+fun Optional<DimensionProto>.getDim(): Dimension =
+    this.get().get()
```
rylin8, line 26: This will crash if `Optional` is empty.

designcompose/src/main/java/com/android/designcompose/Layout.kt
```diff
@@ -283,7 +287,10 @@ internal fun calcLayoutInfo(
                     },
                 selfModifier = modifier,
                 marginModifier = marginModifier,
-                padding = style.layout_style.padding,
+                padding =
+                    style.layout_style.padding.orElseThrow {
```
rylin8, line 291: Can these be functions so that the `orElseOrThrow` stays in the function instead of here

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -685,23 +680,23 @@ internal fun defaultLayoutStyle(): LayoutStyle.Builder {
     builder.align_self = AlignSelf.Auto()
     builder.align_content = AlignContent.FlexStart()
     builder.justify_content = JustifyContent.FlexStart()
-    builder.top = Dimension.Undefined()
-    builder.left = Dimension.Undefined()
-    builder.bottom = Dimension.Undefined()
-    builder.right = Dimension.Undefined()
-    builder.margin = zeroRect()
-    builder.padding = zeroRect()
+    builder.top = Optional.of(newDimensionProtoUndefined())
```
rylin8, line 683: Can you make `newDimensionProtoUndefined` return `Optional<>` to avoid having `Optional.of()` everywhere here

designcompose/src/main/java/com/android/designcompose/proto/DimensionRect.kt
```diff
@@ -0,0 +1,73 @@
...
+
+package com.android.designcompose.proto
+
+import com.android.designcompose.definition.element.dimensionRect
+import com.android.designcompose.serdegen.Dimension
+import com.android.designcompose.serdegen.DimensionRect
+import java.util.Optional
+
+fun Optional<DimensionRect>.getStart(): Dimension =
+    this.orElseThrow { NoSuchFieldException("Malformed data: DimensionRect unset") }.start.getDim()
```
rylin8, line 25: You might be able to do this as a property getter, so that you can just call `.start` instead of `.getStart()`.
```
var A.someProperty: Int
  get() = /* return something */
```

rylin8 (commented):
Not sure why commas were added everywhere. I thought the linter would do that automatically. It kind of makes the changes harder to read when there are a bunch of superfluous changes.

rylin8 (commented):
With the latest changes, do you still need `DimensionProxy` and `DimensionRectProxy`?

.idea/codeStyles/Project.xml
```diff
@@ -143,4 +138,4 @@
       </indentOptions>
     </codeStyleSettings>
   </code_scheme>
-</component>
+</component>
```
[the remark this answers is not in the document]
timothyfroehlich: Whoops

crates/figma_import/src/tools/fetch_layout.rs
```diff
@@ -131,65 +132,74 @@ fn test_layout(
     *id = *id + 1;
     if let ViewData::Text { content: _, res_name: _ } = &view.data {
         let mut use_measure_func = false;
-        if let Dimension::Auto = view.style.layout_style.width {
-            if let Dimension::Auto = view.style.layout_style.height {
+        if let Dimension::Auto(()) = view.style.layout_style.width.unwrap().dimension.unwrap() {
+            if let Dimension::Auto(()) = view.style.layout_style.height.unwrap().dimension.unwrap()
+            {
```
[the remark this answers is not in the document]
timothyfroehlich: It's right at the character limit so the formatter put it on the next line.

crates/figma_import/tests/layout_tests.rs
```diff
@@ -72,50 +73,57 @@ fn add_view_to_layout(
...
+            manager
+                .add_style_measure(
+                    my_id,
+                    parent_layout_id,
+                    child_index,
+                    view.style.layout_style.clone(),
+                    view.name.clone(),
+                    measure_func,
+                )
+                .unwrap();
```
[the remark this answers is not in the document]
timothyfroehlich: Cargo fmt says otherwise

dev-scripts/update_layout_tests.sh
```diff
@@ -0,0 +1,26 @@
...
+set -euo pipefail
+GIT_ROOT=$(git rev-parse --show-toplevel)
+
+source "$GIT_ROOT/dev-scripts/common-functions.sh"
+
+echo "Update the dcf file for the rust layout tests"
+
+warn "Remember to update the nodes in the script to keep them in sync with the layout tests!"
+
+fetch_layout_tests
```
rylin8, line 26: You can delete dev-scripts/test-scripts/get-layout-test.sh now.

rylin8 (approved):

PR #1600, opened 2024-09-06
Look up image res for images with the shared plugin data

yiqunw700:


crates/figma_import/src/transform_flexbox.rs
```diff
@@ -622,6 +622,7 @@ fn compute_background(
                 transform: Some(transform.to_2d()),
                 opacity: last_paint.opacity,
                 scale_mode: bg_scale_mode,
+                res_name: images.image_res(image_ref).map(|s| s.to_string()),
```
rylin8, line 625: not sure why you need a `to_string()` here. Maybe because `image_res()` returns `Option<&String>` instead of `Option<String>`? If so, you can probably change it to return `Option<String>`.

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -1340,10 +1343,28 @@ internal fun Background.asBrush(
...
         is Background.Image -> {
             val backgroundImage = this
+            val imageTransform = backgroundImage.transform.asSkiaMatrix()
+            backgroundImage.res_name.orElse(null)?.let {
+                val resId =
+                    appContext.resources.getIdentifier(it, "drawable", appContext.packageName)
+                if (resId != Resources.ID_NULL) {
+                    return Pair(
+                        RelativeImageFill(
+                            image = BitmapFactory.decodeResource(appContext.resources, resId),
```
rylin8, line 1353: I'd like to know if this decode function does any kind of caching, because it looks like this could get called over and over again and be quite slow.

rylin8 (commented):
Looks good, with a caveat that I'd like to explore whether `BitmapFactory.decodeResource()` is being called repeatedly when the image is redrawn. We could add a test with a localized background image that needs to be redrawn for some reason -- it is moving around or something. Let's add that test, and if it does reveal unncessary decoding, fix it. We can do this in a separate ticket.

crates/figma_import/src/transform_flexbox.rs
yiqunw700: Yes. That is right. HashMap get method returns Option<&String>. I moved this mapping to the image_res function.

designcompose/src/main/java/com/android/designcompose/Utils.kt
yiqunw700: I will look into it in a separate pr. Yes it will be called every time draw triggers. I will do caching in a following pr.

rylin8 (approved):

PR #1604, opened 2024-09-10
Fix crash in replacement component with AndroidView

### LINKED ISSUE #1603: Fix crash in replacement component with AndroidView

rylin8:
Fixes #1603. Use the composable `key()` function only for list content, not component replacement, to avoid unneeded recomposition when the replacement component is AndroidView. This avoids the recreating the AndroidView when the .dcf file is updated.


rylin8:
I'm not sure this is the best fix, but it fixes the crash. I believe they `key()` function was causing a recomposition of the component replacement that was the `AndroidView`. By removing the `key()` function, it no longer recomposes and so the `factory` function only gets called once. I still have a couple questions.

1. Is the `key()` function needed? Looking at documentation, it looks to be helpful for list content changing, which is why I kept it around for that purpose only. However even with it removed, I found no regressions with list replacement content.
2. I don't know enough about `AndroidView` to know if we should handle the case that it does legitimately get recreated.

rylin8:
FYI @timothyfroehlich if this goes in, it will needed to be cherry picked into 0.30.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -575,9 +576,15 @@ fun SquooshRoot(
...
                     // constraints that we calculate in our layout glue code above. This gives
                     // children a way to report their sizes to the Rust layout implementation.
-                    key(child.node.layoutId) {
-                        SquooshChildLayout(modifier = composableChildModifier, child = child)
-                    }
+                    val isListItem =
+                        child.node.parent?.view?.name?.let {
+                            customizationContext.getContent(it) != null
+                        } ?: false
+                    if (isListItem)
```
iamralpht, line 583: I used `key` to avoid list items getting recomposed when they get reordered, so this still meets that need. I am wondering if we had an `AndroidView` in a list item if we'd just see the same thing?

I suspect the crash we're seeing in the DriverUI demo app is really a race condition for a really short-lived Filament View, and it should just handle getting created and destroyed before it has finished loading.

I think this is OK with an updated comment on why key is used, and why it's not used, but also worth checking if it really is that we're just always crashing with a short lived Filament component (our code around Filament is doing some async stuff which probably is where the problem is).

rylin8: I thought it was related to Filament at first, but we tested with `AndroidView` for the backup camera and the map video, and those crashed as well.

iamralpht: Oh :(, this seems like a bug higher up then, so this workaround is OK with a comment and link to a ticket.

rylin8: Done

iamralpht (approved):
Ok, looks good. I think the CI failure is a config issue (missing NDK)

rylin8:
> Ok, looks good. I think the CI failure is a config issue (missing NDK)

@timothyfroehlich I believe you are looking at this in another PR.

PR #1611, opened 2024-09-12
Improve DCF Fetching, Update Doc version to 22

timothyfroehlich:
Fix the fetchAndUpdateFigmaFiles task to only run the tests that can fetch the files.

This adds a "update-dcfs.sh" script which will fetch each of the DCFs.

Remove the update_tests_dcf.sh script which is now redundant since fetchAndUpdate works properly.

Improve ./test-all.sh to detect your Unbundled Gradle project from the main gradle project.


dev-scripts/update-dcfs.sh
```diff
@@ -0,0 +1,137 @@
...
+        --nodes="#SettingsView" \
+        --nodes="#FigmaDoc" \
+        --nodes="#Message" \
+        --nodes="#MessageFailed" \
+        --nodes="#LoadingSpinner" \
+        --nodes="#Checkbox" \
+        --nodes="#NodeNamesCheckbox" \
+        --nodes="#MiniMessagesCheckbox" \
+        --nodes="#ShowRecompositionCheckbox" \
+        --nodes="#UseLocalStringResCheckbox" \
```
yiqunw700, line 26: I am having a change to update this to UseLocalResCheckbox WIP(figma file has been updated...)
```suggestion
        --nodes="#UseLocalResCheckbox" \
```

timothyfroehlich: Thanks!

.shellcheckrc
```diff
@@ -0,0 +1,3 @@
+external-sources=true
+disable=SC2250
+source-path=dev-scripts/doc-fetches
```
rylin8, line 3: What is this file?

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/RenderAllExamples.kt
```diff
@@ -59,6 +61,7 @@ fun performLiveFetch(dcfOutPath: String?) {
         }
 }
 
+@Category(Fetchable::class)
```
rylin8, line 64: So this is so that the fetch task just fetches things in here, but not the other tests?

.shellcheckrc
timothyfroehlich: It configured the shellcheck linter, which I'd definitely recommend installing into your VSCode.

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/RenderAllExamples.kt
timothyfroehlich: Yes, it's a filter tag.

rylin8 (approved):

PR #1612, opened 2024-09-12
Refactor Github Workflow Build Tool Setup 

timothyfroehlich:
Since we need to manually install the required NDK it's time to just properly abstract the setup tasks out into a reusable action.


yiqunw700:
Looks like we also need it for the `build-maven-repo` task

timothyfroehlich:
I think this will work, if everything passes then go ahead and submit it.

yiqunw700 (approved):

PR #1626, opened 2024-09-13
Support shadow colors using Figma variables

### LINKED ISSUE #1613: Shadow colors using Figma variables not supported
We currently just extract the raw color from the variable, so it never changes when variable modes or collections change dynamically.

rylin8:
Fixes #1613


rylin8:
Not sure why the design switcher image shows a diff here, but the png looks correct.

yiqunw700:
> Not sure why the design switcher image shows a diff here, but the png looks correct.

I am updating it. See #1628

yiqunw700:
I'm wondering why some of the screenshots have a diff now, for example the battle ship?

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -810,7 +810,7 @@ internal fun squooshShapeRender(
 
         // Make an appropriate paint.
         val shadowPaint = Paint().asFrameworkPaint()
-        shadowPaint.color = shadow.shadowStyle.color.toColor().toArgb()
+        shadowPaint.color = shadow.shadowStyle.color.getValue(variableState)?.toArgb() ?: 0
```
yiqunw700, line 813: I'm wondering if we should let it return@forEach when getValue(variableState) returns null

rylin8:
> I'm wondering why some of the screenshots have a diff now, for example the battle ship?

You mean why I checked in new pngs? The roborazzi task made new images, but I don't see any visual diffs. @timothyfroehlich any idea why?

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: The forEach is for the list of shadow fills. If we fail to get the variable value out of one, I would think we should still keep going through the list to draw the others.

yiqunw700: Yes. I agree. 
`return@forEach` will skip the current one and continue the next one in the forEach loop, like a `continue` in Java?

rylin8: Oh right, was confused and thought it was like `break`.

rylin8: Done

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -547,7 +548,7 @@ internal fun ContentDrawScope.render(
...
```
yiqunw700, line 551:
```suggestion
        shadowPaint.color = shadow.shadowStyle.color.getValue(variableState)?.toArgb() ?: return@forEach
```

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -810,7 +811,7 @@ internal fun squooshShapeRender(
...
```
yiqunw700, line 814:
```suggestion
        shadowPaint.color = shadow.shadowStyle.color.getValue(variableState)?.toArgb() ?: return@forEach
```

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -869,7 +870,7 @@ internal fun squooshShapeRender(
...
```
yiqunw700, line 873:
```suggestion
        shadowPaint.color = shadow.shadowStyle.color.getValue(variableState)?.toArgb() ?: return@forEach
```

yiqunw700:
> > I'm wondering why some of the screenshots have a diff now, for example the battle ship?
> 
> You mean why I checked in new pngs? The roborazzi task made new images, but I don't see any visual diffs. @timothyfroehlich any idea why?

I  compared the the old battleship screenshots using the image diff tool, but can not tell anything different. It might just be that rgb data now is argb with extra alpha data but no visual difference.

yiqunw700 (approved):

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: Done

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: Done

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: Done

Issue #1556, opened 2024-08-26
Export images from figma design


yiqunw700:
Plugin support to set drawable resource name, export drawables.
DC support to use drawable resource at runtime.

PR #1628, opened 2024-09-13
Update the useLocalStringRes to useLocalRes

yiqunw700:
Use this flag to gate loading drawable resource for images.

Towards: #1556


designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -1344,22 +1345,24 @@ internal fun Background.asBrush(
...
-                            imageDensity = density,
-                            displayDensity = density,
-                            imageTransform = imageTransform,
-                            scaleMode = backgroundImage.scale_mode
-                        ),
-                        backgroundImage.opacity
-                    )
-                } else {
-                    Log.w(TAG, "No drawable resource $it found")
+            if (useLocalDrawableRes != false) {
```
rylin8, line 1348: Do you know if getting the `useLocalDrawableRes` value from `DebugNodeManager` directly here would work? I'm just trying to avoid adding more parameters to our functions that already have a lot of parameters. I think doing it here wouldn't trigger a recomposition, but maybe that's ok (and good?). This isn't a big deal since this is just a debugging feature, but we do seem to keep adding parameters to functions.

designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt
```diff
@@ -77,7 +77,7 @@ private interface DesignSwitcher {
             "#NodeNamesCheckbox",
             "#MiniMessagesCheckbox",
             "#ShowRecompositionCheckbox",
-            "#UseLocalStringResCheckbox",
+            "#UseLocalResCheckbox",
```
rylin8, line 80: Please update the shell script also.

rylin8: Oh, I see Tim's PR updates it.

designcompose/src/main/java/com/android/designcompose/Utils.kt
yiqunw700: Squoosh works. It might because when we toggle the design switcher toggle, it draws the entire tree.
But the non-squoosh does not work upon toggling the checkbox, but it will render correctly the next time it draws.
Would you like to do this way? I can revert the parameter changes if so.

rylin8: Interestingly, I did a test and found that for text localization, the opposite is true: squoosh will not update immediately, but non-squosh does.

To fix this, I think you can use `MutableState<Boolean>` instead of `LiveData`. Then the declaration would be:
```
    private val useLocalStringRes: MutableState<Boolean> = mutableStateOf(true)
```
and the getter would return a `State<Boolean>`.

yiqunw700: I'm changing to use the State<Boolean>. For the squoosh text not updating upon toggling the checkbox in design switcher, it might be related to the text cache squoosh is using. Let me check...

yiqunw700: Done.

designcompose/src/main/java/com/android/designcompose/Utils.kt
```diff
@@ -1343,23 +1343,26 @@ internal fun Background.asBrush(
...
-                            imageDensity = density,
-                            displayDensity = density,
-                            imageTransform = imageTransform,
-                            scaleMode = backgroundImage.scale_mode
-                        ),
-                        backgroundImage.opacity
-                    )
-                } else {
-                    Log.w(TAG, "No drawable resource $it found")
+            if (DebugNodeManager.getUseLocalRes().value != false) {
```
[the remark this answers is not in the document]
yiqunw700: Yes. Done.

rylin8 (approved):

PR #1634, opened 2024-09-17
Fix regression which caused the test tasks to only run Fetchable tests.

timothyfroehlich:


dev-scripts/common-functions.sh
```diff
@@ -103,3 +103,13 @@ function fetch_layout_tests {
         --nodes='ConstraintsLayoutWidget' \
         --nodes='VectorScale'
 }
+
+# This doc is difficult to fetch because the test that would fetch it can only run on older APIs
```
yiqunw700, line 107: I don't really know what happened here. But I think it might not because of the @TargetApi annotation, because it is mainly for lint purpose and we are running on api 28+. The only reason I can think of, is because, this one passes a parameter. https://github.com/google/automotive-design-compose/blob/0efb3b07dd8ff44603c1f9c7dc5dac24e38e1de3/integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt#L60

rylin8: @timothyfroehlich do you know if removing the @TargetApi fixes the issue of the state customization not fetching? Or if it's the parameter?

rylin8 (approved):

Issue #1676, opened 2024-09-23
Fix the bitmap cache key used for drawable lookup


yiqunw700:
When change a language the bitmap loaded from drawable doesn't update(need to use a key containing the system locale info)

PR #1678, opened 2024-09-23
Add locale info to the image cache key

yiqunw700:
Fixes: #1676

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/BitmapFactoryWithCache.kt
+++ b/designcompose/src/main/java/com/android/designcompose/BitmapFactoryWithCache.kt
@@ -39,7 +39,9 @@ object BitmapFactoryWithCache {
     }
 
     fun loadResource(resources: Resources, resId: Int): Bitmap {
-        val imageKey: String = resId.toString()
+        val locales = resources.configuration.locales
+        val localeId = if (locales.isEmpty) "" else locales[0].toString()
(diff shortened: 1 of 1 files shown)


designcompose/src/main/java/com/android/designcompose/BitmapFactoryWithCache.kt
```diff
@@ -39,7 +39,9 @@ object BitmapFactoryWithCache {
     }
 
     fun loadResource(resources: Resources, resId: Int): Bitmap {
-        val imageKey: String = resId.toString()
+        val locales = resources.configuration.locales
+        val localId = if (locales.isEmpty) "" else locales[0].toString()
```
rylin8, line 43:
```suggestion
        val localeId = if (locales.isEmpty) "" else locales[0].toString()
```

designcompose/src/main/java/com/android/designcompose/BitmapFactoryWithCache.kt
```diff
@@ -39,7 +39,9 @@ object BitmapFactoryWithCache {
...
+        val imageKey: String = resId.toString() + localId
```
rylin8, line 44:
```suggestion
        val imageKey: String = resId.toString() + localeId
```

rylin8 (approved):
localId -> localeId

Issue #1545, opened 2024-08-21
Min/max width/height auto layout constraints don't seem to work


iamralpht:
I noticed that setting min/max constraints on an auto layout node seems to be ignored. It would be good to support this fully; I think that Taffy supports min/max, so we should be able to hook this up.

iamralpht:
Lower priority than #1556 for sure

yiqunw700:
Having issue with measuring with node https://www.figma.com/design/ArHVZAQIsKf2B9mATnVvLX/AutoLayout?node-id=1-8&node-type=frame&t=xrMeVa2Wr6yzULtn-0 where the outer frame has max size set and the inner frame has min size set.
Also notice that the flex direction has an impact on the result:
Vertical layout ->height is measured only with paddings added up
Horizontal layout -> width is measured only with paddings added up

This only happens when the outer container is placed inside a non-auto layout container or an auto-layout container with a different flex direction. Works if it is placed inside an auto-layout container with the same flex direction.

This is similar to what have been reported to taffy: https://github.com/DioxusLabs/taffy/issues/591

PR #1679, opened 2024-09-23
Read min/max from figma

yiqunw700:
It used to be ignored.

Fixes: #1545


integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/AllExamples.kt
```diff
@@ -29,6 +29,7 @@ val EXAMPLES: ArrayList<Triple<String, @Composable () -> Unit, String?>> =
         Triple("HelloVersion", { HelloVersion() }, HelloVersionDoc.javaClass.name),
         // Alphabetically ordered and trying to put similar tests together...
         Triple("Alignment", { AlignmentTest() }, AlignmentTestDoc.javaClass.name),
+        Triple("AutoLayout", { AutoLayoutTest() }, AutoLayoutTestDoc.javaClass.name),
```
rylin8, line 32: We have a few tests that test various aspects of autolayout. Is there a reason this one is called AutoLayout? If not can we rename it to MinMax or something?

rylin8 (approved):

PR #1680, opened 2024-09-24
Add support for animating text and vector sizes by scaling

### LINKED ISSUE #1641: Text and vectors do not not resize during animations
Perform a scale transformation for text that changes size and vector paths that change size.

rylin8:
Fixes #1641. Set the scale value in the transformation matrix for text nodes that animate to a different font size, and for vector paths that change size. Do this with a crossfade so that it will work with other changes such as font weight as well.

Modify the fade in/out animations to account for the target node opacity.

Refactor the AnimatedLayoutItem into more functions so that we can more easily support additional style animations in the future.


iamralpht:
Ok - code change looks good. I think you should add a test for the fixed partial opacity transition case. I love that the existing animation tests no longer show duplicated phone icons, this is a huge improvement!

rylin8:
Added some more tests

iamralpht (approved):

PR #1683, opened 2024-09-26
[WIP]: Yiqunw/auto layout with taffy5

yiqunw700:


iamralpht:
Oh! Battleship isn't completely broken in your image comparisons... maybe upstream did get to this?

yiqunw700:
> Oh! Battleship isn't completely broken in your image comparisons... maybe upstream did get to this?

No. Because there is a `apply_hack` in the layout_manager in the original pr(that is also why the fill container and grid layout test fail to grow to fill). It took me a while to figure out we had a hack there...

iamralpht:
> No. Because there is a `apply_hack` in the layout_manager in the original pr(that is also why the fill container and grid layout test fail to grow to fill). It took me a while to figure out we had a hack there...

Oops! I'm sorry about that, I think I thought I was close to being done with this and then realized the shrinking case was broken and I couldn't fix it. FWIW all of the other screenshots look like they're more accurate than with the old layout release, so that much at least is better.

PR #1686, opened 2024-09-27
Fix item spacing of list replacement content in squoosh

### LINKED ISSUE #1541: List replacement content in squoosh may be spaced incorrectly
Spacing seems correct if the container is set to hug contents, but incorrect if set to fixed.

rylin8:
Fixes #1541. Set the default flex_grow value to 0 for list replacement items. This prevents the items in the list from stretching to fit the container they are in.

This change puts several of the validation tests previously in the DEFAULT_RENDERER_ONLY category into the BOTH category. Squoosh is now able to render these examples correctly, with some (CrossAxis Fill) being more correct than the non-squoosh version.


rylin8:
@timothyfroehlich the ticket for this is marked as 0.32 because I didn't think I'd get to it but I did. I'm not sure on the status of 0.31 but if this happens to get in before the 0.31 RC feel free to move the ticket to 0.31.

iamralpht (approved):

iamralpht:
@timothyfroehlich I don't believe we've cut the RC, so I'm inclined to get this in now, as I suspect the PR will bitrot otherwise.

rylin8:
Ok, changed the ticket to 0.31.

Issue #1677, opened 2024-09-23
Plugin save user actions to break strings into two translations


yiqunw700:
Follow up of #1469 to save the actions when users have broken strings into two translations.

PR #1694, opened 2024-10-03
Remember user's action to break strings into different translations

yiqunw700:
This is achieved by trying to use the saved resource name whenever possible. No overriding by group options.
    
Fixes: #1677

### DIFF
--- a/support-figma/extended-layout-plugin/src/localization-module.ts
+++ b/support-figma/extended-layout-plugin/src/localization-module.ts
@@ -20,6 +20,8 @@ import * as DesignSpecs from "./design-spec-module";
 const STRING_RES_PLUGIN_DATA_KEY = "vsw-string-res";
 const STRING_RES_EXTRAS_PLUGIN_DATA_KEY = "vsw-string-res-extras";
 const EXPLICIT_EXCLUSION_PLUGIN_DATA_KEY = "vsw-string-explicit-exclusion";
+// This saves the characters of the text nodes. If it changes, the res name can become invalid.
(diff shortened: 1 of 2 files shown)


support-figma/extended-layout-plugin/src/localization-module.ts
```diff
@@ -263,65 +336,138 @@ async function localizeNodeAsync(
...
+  var preferredName = getResName(node);
+  if (preferredName) {
+    let normalizedText = normalizeTextNode(node);
+    const isNodeExcluded = isExplicitlyExcluded(node);
+    var cachedCharacters = node.getPluginData(
+      STRING_RES_CHARACTERS_PLUGIN_DATA_KEY
+    );
+    // This node has been exported as string resource before and it hasn't changed,
+    // use its res name and put in the string res map.
+    if (cachedCharacters === JSON.stringify(normalizedText)) {
```
iamralpht, line 353: I'm not sure if `JSON.stringify` is stable (I think the serialization order is based on when fields were added to an object chronologically, not the alphabetical order of fields -- there might be a flag to control this?).

yiqunw700: This `normalizedText` is a string or a string array. So I think it should be stable here.

rylin8:
I noticed that when I click on the break button, the toast that is shown on mouseover stays visible and never goes away. Can you reproduce that?
Screenshot 2024-11-12 at 11 07 31 AM

rylin8 (approved):

PR #1695, opened 2024-10-03
Migrate Actions and Reactions to Proto

timothyfroehlich:


crates/figma_import/src/component_context.rs
```diff
@@ -53,19 +53,16 @@ impl ComponentContext {
...
-            // anything with those.
-            let destination_node_id = match &reaction.action {
-                Action::Node { destination_id: Some(id), .. } => id,
-                _ => continue,
-            };
-
-            // If we've already converted this node then don't add it again.
-            if self.converted_nodes.contains(destination_node_id) {
-                continue;
+            // // anything with those.
```
yiqunw700, line 56:
```suggestion
            // anything with those.
```

yiqunw700:
I tried the interaction test. It seems the overlays are broken. I can not see any overlays. Could you try on your end, the `Interaction` -> `Overlay tests`?

timothyfroehlich:
Oi, yeah, they're not working. Let me try to add some tests that'll catch that, and then I'll fix them.

designcompose/src/main/java/com/android/designcompose/DesignView.kt
```diff
@@ -309,19 +315,25 @@ internal fun DesignView(
...
+            reaction.trigger.type?.run {
+                when (this) {
+                    is TriggerType.Click -> onClickReactions.add(reaction)
+                    is TriggerType.Press -> onPressReactions.add(reaction)
+                    is TriggerType.Drag -> onDragReactions.add(reaction)
+                    is TriggerType.AfterTimeout -> {
+                        if (value.timeout < currentTimeout) {
+                            onTimeout = reaction
+                            currentTimeout = value.timeout
+                        } else {
```
yiqunw700, line 330: We don't really need this `else {{}}` right? Remove?

timothyfroehlich: The extra {} isn't needed, but the else is needed when used in a `when`, apparently.

PR #1710, opened 2024-10-04
Update to API 35

timothyfroehlich:
Needed to add a seperate variable for MediaCompose because Unbundled hasn't released stubs for api-35 yet.


yiqunw700 (approved):

iamralpht:
Did you try this manually in the emulator? API35 opts everything into edge-to-edge mode, so if we aren't consuming insets correctly then we might have system bars or window controls drawing over the app content.

Issue #1643, opened 2024-09-19
Support "ignore auto layout" property


rylin8:
Nodes inside an autolayout frame are laid out according to the autolayout properties. Figma supports a feature to set a node within the autolayout frame to "ignore auto layout" and instead be absolutely positioned. Taffy also supports this, so we should be able to support this.

rylin8:
![Screenshot 2024-09-19 at 1 41 40 PM](https://github.com/user-attachments/assets/6975120a-1bb8-4005-aa9a-ba2a24a50284)

PR #1713, opened 2024-10-07
Support "ignore auto layout"

yiqunw700:
When ignore auto layout, figma returns layoutPositioning as absolute.

Fixes: #1643


timothyfroehlich (approved):
LGTM!

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -121,7 +122,10 @@ fn compute_layout(
 
     // Frames can implement Auto Layout on their children.
     if let Some(frame) = node.frame() {
-        style.layout_style.position_type = PositionType::Relative;
+        style.layout_style.position_type = match frame.layout_positioning {
+            LayoutPositioning::Absolute => PositionType::Absolute,
+            LayoutPositioning::Auto => PositionType::Relative,
+        };
```
rylin8, line 128: Wow nice, so this doesn't require any changes to how it is passed into taffy? I guess setting this to absolute converts to the same type of attribute in taffy::style, which it supports?

yiqunw700: Yeah that's right. We already have absolute positioning supported in non auto layout cases, so just need to update the tranform_flexbox.rs. Turned out straight forward.

Issue #1712, opened 2024-10-07
Overlay doesn't dismiss on tap for non-squoosh renderer


yiqunw700:
Run validation app->Interaction->Overlay Position test.

yiqunw700:
https://github.com/google/automotive-design-compose/blob/96cfe7c24b74614784c841929121d903a2c00fb1/designcompose/src/main/java/com/android/designcompose/DesignView.kt#L972

This looks like the root cause.

PR #1715, opened 2024-10-08
Fix the non-squoosh renderer not able to close the overlay

yiqunw700:
Also fixes crashes where removeLast() can not be found on device api level 33.

Fixes: #1712

### DIFF
--- a/common/src/main/java/com/android/designcompose/common/Utils.kt
+++ b/common/src/main/java/com/android/designcompose/common/Utils.kt
@@ -178,7 +178,7 @@ class VariantPropertyMap {
                     )
                 if (view != null) return view
             }
-            resolvedNameProperties.removeLast()
+            resolvedNameProperties.removeLastOrNull()
         }
         if (possibleValues.contains("*")) {
             resolvedNameProperties.add("$propertyName=*")
@@ -197,7 +197,7 @@ class VariantPropertyMap {
                     )
                 if (view != null) return view
             }
-            resolvedNameProperties.removeLast()
+            resolvedNameProperties.removeLastOrNull()
         }
         return null
     }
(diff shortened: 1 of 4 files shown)


yiqunw700:
I couldn't figure out how to fix the sizing of the overlay. Filed #1717 ticket to look into that afterwards.

timothyfroehlich:
It works for me! But I think Ralph should approve.

iamralpht:
I thought of one other thing to test - in the navigation tests, there's an overlay that appears from the bottom which has buttons behind it. It's worth testing that when the overlay is visible, you can't click through it to the buttons behind.

(I think this regression must have come from a Compose update, so it's also worth adding a regression test for this case)

timothyfroehlich:
> I thought of one other thing to test - in the navigation tests, there's an overlay that appears from the bottom which has buttons behind it.

There is? I'm not seeing it on my emulator, with or without squoosh.

timothyfroehlich:
> I thought of one other thing to test - in the navigation tests, there's an overlay that appears from the bottom which has buttons behind it. It's worth testing that when the overlay is visible, you can't click through it to the buttons behind.
> 
> (I think this regression must have come from a Compose update, so it's also worth adding a regression test for this case)

@iamralpht I added navigation tests to #1711, do those look like they catch what you're looking for?

yiqunw700:
> I thought of one other thing to test - in the navigation tests, there's an overlay that appears from the bottom which has buttons behind it. It's worth testing that when the overlay is visible, you can't click through it to the buttons behind.

In the overlay positioning test, when overlay shows up, I am not able to click on any buttons(nothing happens) outside of the overlay content.
In the overlay navigation test, when overlay shows up, clicking outside of the overlay content(anywhere including the buttons behind) dismisses the overlay because figma sets the CLOSE_ON_CLICK_OUTSIDE set to be true.

iamralpht:
Yeah, I think the updated tests cover the scenario. Thanks for making the improvements!

iamralpht (approved):

PR #1740, opened 2024-10-28
ProtoConv: Wrap up most of definition/elements

timothyfroehlich:
# Based on PR #1738

## Chain of upstream PRs as of 2024-10-28

* PR #1738:
  `main` ← `wb/froeht/convert-font`

  * **PR #1740 (THIS ONE)**:
    `wb/froeht/convert-font` ← `wb/froeht/almost-finish-elements`


crates/dc_bundle/src/definition/element.rs
```diff
@@ -490,3 +490,9 @@ impl FontWeight {
     /// Black weight (900), the thickest value.
     pub const BLACK: FontWeight = FontWeight::from_num(900.0);
 }
+
+impl Vector {
+    pub fn is_valid(&self) -> bool {
```
iamralpht, line 495: `is_non_zero`? `has_magnitude`? Something like that, bc a zero vector is "valid" in many contexts.

iamralpht (approved):

iamralpht:
LGTM - can do the fixup on `is_valid` in a follow-up if you've got more stacked on top of this.

crates/dc_bundle/src/definition/element.rs
timothyfroehlich: Actually I don't think it was being used anywhere so I just deleted it.

Issue #1714, opened 2024-10-07
Shader plugin support


yiqunw700:
Create a simple shader plugin that can preview and insert background image

PR #1775, opened 2024-11-05
Create a simple shader plugin with CanvasKit

yiqunw700:
Supports:
- Text area to modify/input shader code
- Preview shader in the plugin UI
- Insert shader code into the shared plugin data of the selected node that supports background image fill
- Default shaders(from skia and shader toy)
- Generate an image and insert into the figma selected node

What don't work yet:
- Not allow other uniforms other than iTime and iResolution
- Don't support strokes

Towards: #1714 
Screenshot 2024-12-12 at 14 56 50


designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -451,18 +459,30 @@ internal fun ContentDrawScope.render(
...
-    val customFillBrush =
+    var customFillBrush =
         if (customFillBrushFunction != null) {
             customFillBrushFunction()
         } else {
             customizations.getBrush(name)
         }
+    if (customFillBrush == null) {
+        view.shader.getOrNull()?.let {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
```
timothyfroehlich, line 470: So this means shaders will only work on T or higher, and if not on T it'll just not render? Is this something new or is this a restriction that we've had before? Is this documented anywhere?

I don't like the idea of the doc rendering inaccurately if you're on too old of a build. IMO, the appropriate way to handle this would be to increase our minSDK version (after discussion with Ralph of course) or to seperate the shader parts into a seperate library with a higher minSDK, so that trying to build for a lower SDK causes a build failure, rather than an inaccurate rendering that looks like a bug.

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/RenderAllExamples.kt
```diff
@@ -44,7 +45,7 @@ import org.robolectric.annotation.GraphicsMode
 // Enable Robolectric Native Graphics (RNG)
 @GraphicsMode(GraphicsMode.Mode.NATIVE)
 @RunWith(ParameterizedRobolectricTestRunner::class)
-@Config(qualifiers = "w1920dp-h1500dp-xlarge-long-notround-any-xhdpi-keyshidden-nonav")
+@Config(qualifiers = "w1920dp-h1500dp-xlarge-long-notround-any-xhdpi-keyshidden-nonav", sdk = [33])
```
timothyfroehlich, line 48: We want this for sdk 33 or higher, right? So you should uses [minsdk](https://robolectric.org/configuring/#configure-sdk-level)

timothyfroehlich: And also, why do we need to set the minSDK if the test is disabled? I realized this is RenderAllExamples, we want this to cover all of our supported SDK versions, so we shouldn't be setting sdk restrictions on it.

support-figma/extended-layout-plugin/manifest.json
```diff
@@ -21,13 +33,20 @@
...
   ],
   "networkAccess": {
     "allowedDomains": [
       "https://cdn.jsdelivr.net/gh/thomas-lowry/figma-plugin-ds/dist/figma-plugin-ds.css",
       "https://rsms.me",
-      "https://cdnjs.cloudflare.com/ajax/libs/jszip/3.10.1/jszip.min.js"
+      "https://cdnjs.cloudflare.com/ajax/libs/jszip/3.10.1/jszip.min.js",
+      "https://unpkg.com"
     ],
     "reasoning": "The plugin uses the stylesheet from https://github.com/thomas-lowry/figma-plugin-ds to make the UI consistent with the rest of the Figma UI (https://www.figma.com/plugin-docs/figma-components/). It uses https://rsms.me for the Inter fonts. jszip is the tool to download images in zip."
```
timothyfroehlich, line 51: Please add a bit explaining the need for unpkg.com

support-figma/extended-layout-plugin/src/data/shader/gradient_flow.txt
```diff
@@ -0,0 +1,75 @@
...
+mat2 Rot(float a)
+{
+    float s = sin(a);
+    float c = cos(a);
+    return mat2(c, -s, s, c);
+}
+
+
+// Created by inigo quilez - iq/2014
+// License Creative Commons Attribution-NonCommercial-ShareAlike 3.0 Unported License.
```
timothyfroehlich, line 15: We can't use code with this license, according to [go/licensetable](https://goto.google.com/licensetable) it's forbidden.

timothyfroehlich (changes requested):
We have some pretty serious licensing concerns to deal with before I can unblock this PR.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
iamralpht: Yeah, the Android framework support for shaders landed in TM. I don't think there's a practical way to support them on SCv2 or earlier.

support-figma/extended-layout-plugin/src/data/shader/gradient_flow.txt
iamralpht: Yeah - I am guessing (we should check) that the shaders included in the skia shaders playground are OK for us to include here. It's hard to find any license info for stuff from shadertoy... since these are just samples of what can be done, it's OK to have fewer of them.

support-figma/extended-layout-plugin/manifest.json
yiqunw700: Done.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: https://developer.android.com/partners/google-apps/android-13/changes?hl=en#shaders

When shader isn't supported on old sdks, our render process will continue to render the background image if there is any. Currently the shader plugin will insert an image generated from the running shader when it writes the shader code to the plugin data.

timothyfroehlich: > Currently the shader plugin will insert an image generated from the running shader when it writes the shader code to the plugin data.

I don't think that's a good solution, I think any user would think that this were a bug. So @iamralpht it looks like we'll need to increase our minSDK to support this.

support-figma/extended-layout-plugin/src/shader.html
```diff
@@ -0,0 +1,226 @@
...
+
+            const effect = CanvasKit.RuntimeEffect.Make(prog);
+
+            function drawFrame(canvas) {
+                canvas.clear(CanvasKit.BLACK);
+                const iTime = (Date.now() - startTimeMs) / 1000;
+                const uniforms = [
+                    shaderWidth,
+                    shaderHeight,
+                    1, // iResolution
```
rylin8, line 138: Can this be changed in the plugin? Would it be useful?

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: I don't think we need to restrict SDK versions just for this feature. I would imagine there can be many users of DesignCompose that don't need shader support, and I don't think we need to restrict them to newer SDK versions. We can add documentation for this in our docs as well as the shader plugin UI.

support-figma/extended-layout-plugin/src/shader.html
```diff
@@ -0,0 +1,226 @@
...
+            <div>uniform float3 iResolution;</div>
+            <div>uniform float iTime;</div>
+        </div>
+    </div>
+
+    <div class="container" style="margin-top: 16px;">
+        <button id="previewShaderButton" class="styled-button" style="width: 160px;"
+            onclick="runShader()">Preview</button>
+        <button id="insertImageButton" class="styled-button" style="width: 160px;" onclick="runShader()">Insert Running
+            Shader</button>
```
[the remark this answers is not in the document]
yiqunw700: Removed.

support-figma/extended-layout-plugin/src/data/shader/gradient_flow.txt
yiqunw700: Removed most of them.

There is one `mesh2d.txt` I copied from the DisplaySafety demo which is `inspired from` a shader toy shader. Also a `road.txt` which is similarly modified from a shader toy shader. Is it ok to keep them?

support-figma/extended-layout-plugin/src/shader.html
yiqunw700: iResolution has x,y,z. You are asking the z being 1? I tried to change it to something else, but it doesn't make any differences for those shader examples I put in this pr which only use the x,y coordinates.

I don't really know how z-depth works. The Skia CanvasKit may not even support z depth(this is the answer from [gemini](https://gemini.google.com/corp/share/9b38fa505ca2) if I am asking it the right question).

The shader width and height can change.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: I vote not to restrict SDK too.

support-figma/extended-layout-plugin/src/data/shader/gradient_flow.txt
yiqunw700: I removed those two modified shader toy shaders.

support-figma/extended-layout-plugin/src/data/shader/fibonacci_sphere.txt
```diff
@@ -0,0 +1,15 @@
...
+  vec4 o = vec4(0);
+  vec2 p = vec2(0), c=p, u=FC.xy*2.-iResolution.xy;
+  float a;
+  for (float i=0; i<4e2; i++) {
+    a = i/2e2-1.;
+    p = cos(i*2.4+iTime+vec2(0,11))*sqrt(1.-a*a);
+    c = u/iResolution.y+vec2(p.x,a)/(p.y+2.);
+    o += (cos(i+vec4(0,2,4,0))+1.)/dot(c,c)*(1.-p.y)/3e4;
+  }
+  return o + vec4(0., 0., 0., 1.);
```
rylin8, line 14: why'd you need to add this vector?

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: @iamralpht any thoughts on restricting the SDK version?

timothyfroehlich: Ok, I have an alternate suggestion then, have the shader plugin default to placing an image that states something like "Shaders require Android 13(?) or greater", rather than just the static image. Probably include a checkbox, off by default, that allows the plugin user to instead use the static image of the shader.

support-figma/extended-layout-plugin/src/data/shader/fibonacci_sphere.txt
yiqunw700: It add a black background. Before this change, it doesn't look good on white background.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: Ok. I added a line to the top of the shader plugin.
Screenshot 2024-12-06 at 10 41 01

yiqunw700:
> We have some pretty serious licensing concerns to deal with before I can unblock this PR.

Added a LICENSE file for MIT license(used by star nest) and Unlicense license(examples from skia/twitter). Please take a look.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
timothyfroehlich: I specifically meant in my comment for it to default to putting a static image with that text into the file. My concern is that a designer will see that and not know what it means, or it won't be known that some systems are on the older SDK.

timothyfroehlich:
> > We have some pretty serious licensing concerns to deal with before I can unblock this PR.
> 
> Added a LICENSE file for MIT license(used by star nest) and Unlicense license(examples from skia/twitter). Please take a look.

Ok, I think that's all fine. I've just found go/releasing/preparing#third-party-components, however, which states that those files need to be moved to a /third_party directory in the root of the repo. So we'll need to make that change as well.

timothyfroehlich (dismissed):

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: I got that. The generated image can be used for lower sdk versions(although still need to improve the resolution). But if we put that text, it will not be good to use.

I personally think add this text to the plugin UI is informative.

designcompose/src/main/java/com/android/designcompose/CustomizationContext.kt
```diff
@@ -481,6 +494,12 @@ fun CustomizationContext.getMeterState(nodeName: String): MeterState? {
     return null
 }
 
+fun CustomizationContext.getShaderUniformTimeState(nodeName: String): ShaderUniformTimeState? {
+    val c = cs[nodeName] ?: return null
+    if (c.shaderUniformTimeState.isPresent) return c.shaderUniformTimeState.get()
+    return null
+}
```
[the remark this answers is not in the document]
yiqunw700: Done.

yiqunw700:
> > > We have some pretty serious licensing concerns to deal with before I can unblock this PR.
> > 
> > 
> > Added a LICENSE file for MIT license(used by star nest) and Unlicense license(examples from skia/twitter). Please take a look.
> 
> Ok, I think that's all fine. I've just found go/releasing/preparing#third-party-components, however, which states that those files need to be moved to a /third_party directory in the root of the repo. So we'll need to make that change as well.

Done.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
timothyfroehlich: How will a tester who has no access to Figma and cannot see the plugin UI know that part of the design will not render accurately on certain test systems?

timothyfroehlich:
> > > > We have some pretty serious licensing concerns to deal with before I can unblock this PR.
> > > 
> > > 
> > > Added a LICENSE file for MIT license(used by star nest) and Unlicense license(examples from skia/twitter). Please take a look.
> > 
> > 
> > Ok, I think that's all fine. I've just found go/releasing/preparing#third-party-components, however, which states that those files need to be moved to a /third_party directory in the root of the repo. So we'll need to make that change as well.
> 
> Done.

Looks good.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: The image fill of the figma node can be replaced, there is no guarantee that the generated image will be used. Even we add the text to the image, tester may still not be able to see it.

I think users of the figma plugin and DesignCompose especially users of this feature should be aware of the limitations. And of course we should document it properly in our documents.

babakbo: I agree with folks that increasing min sdk is too much for this case since shaders might be used by only a small fraction of our users. One way of solving this is to break design compose into modules where each module has its own min sdk. However, even that wouldn't solve the main problem for the developers who want shader for devices that support it while they can live with not having it on older devices (which is probably the majority of our users who want shaders).

That said, there is a risk in assuming what the designers want in this case. By falling back to a screenshot of the shader we lose the users who want to use shaders but don't want a static image of the shader on older android versions (imagine a shader that has a weird shape on the first frame). On the other hand, simply dropping the shader can also be confusing and makes it hard for users to use shaders in any meaningful way when they also need to deploy their apps on S-. 

My 2cents here is to give users the option of using static image while making it obvious what happens when shader is not supported. 
For example, is it possible to add an image field as the default image if the shader is not supported?
If I remember correctly, shaders can also be disabled at the hardware level, so just checking for the Android version may not be enough.

In terms of testing, I agree that it complicates the testing process since now systems with shaders have to be tested twice once with support and once without. However, it is the reality of Android. Despite all the compat libraries, app developers should still test their apps on different versions of Android.

wdyt?

yiqunw700: As we discussed last meeting, I added a fallback color using a color picker when shader isn't supported on lower sdk versions.

Screenshot 2024-12-11 at 20 03 09

In the shader plugin UI, it has a checkbox to use that fallback color or the figma node image. Let me know if we want the checkbox or not(i.e. always using the fallback color as the fill).

-----
Regarding if shader can be disabled at hardware level. I didn't find any document stating that. Since AGSL is using skia engine which supports both cpu render and gpu render, the hardware acceleration may affect how shader renders but there seems no way to turn it off.

-----
Please take a look. Thanks!

rylin8: Couple comments on the UI:
- It isn't clear to me that the "Set Shader" button will save the shader into the node. Can we maybe have another little info hover popup that explains what it's going to do?
- Typo on "Andorid"
- I would separate the sentence "Only works on Android 13 and above" with the checkbox. I think the checkbox should be associated with the "Use a solid color..." sentence. Or you could just have the checkbox with a sentence like "Use a solid color when shaders are not supported", and then. have the Android 13 info inside the info hover.

timothyfroehlich: In addition to Rob's comment:
* I think "Set Shader" should be changed to "Apply". It should be in the lower right hand corner, where confirmation buttons are normally found.
* I would phrase the Android 13 bit to something like "Note: Shaders require Android 13+"


Honestly, I think it'd be best to not offer a choice about the fallback. I just realized that if we do offer the ability to fall back to a static image of the shader then we'd have the situation where designers may want a specific frame of the shader to be used, instead of the first frame. So we'd have feature requests to add controls for that. Offering a choice of color is also problematic, it makes it seem like it's a design choice, and it'll defeat the purpose of the fallback if someone chooses a color that doesn't clearly indicate that the design isn't rendering accurately. 

I'm still in favor of using the [standard purple/black checkerboard](https://www.reddit.com/r/explainlikeimfive/comments/7o3nuq/eli5_why_do_games_use_a_purple_checkerboard_for/) but I think just rendering a solid, bright color is also fine.

timothyfroehlich: BTW, I'm sorry you're having to deal with our design feedback. @iamralpht this is a situation where having a designer to help would have been really helpful.

crates/dc_bundle/src/definition/view.rs
```diff
@@ -317,6 +317,8 @@ impl View {
         reactions: Option<Vec<Reaction>>,
         scroll_info: ScrollInfo,
         frame_extras: Option<FrameExtras>,
+        shader: Option<String>,
+        shader_fallback_color: Option<String>,
```
rylin8, line 321: We usually store colors as a `Color` struct, `Int`, or four `u8`s. Any reason this is a string?

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -453,18 +463,36 @@ internal fun ContentDrawScope.render(
...
+                customFillBrush = SizingShaderBrush(shader)
+                val shaderUniformTime = customizations.getShaderUniformTimeState(name)
+                if (shaderUniformTime != null) {
+                    shader.setFloatUniform("iTime", shaderUniformTime.floatValue)
+                }
+                return@let
+            } else {
+                view.shader_fallback_color.getOrNull()?.let { colorString ->
+                    customFillBrush =
+                        SolidColor(Color(android.graphics.Color.parseColor(colorString)))
```
rylin8, line 485: What happens if `parseColor()` fails? Exception?

support-figma/extended-layout-plugin/src/shader.ts
```diff
@@ -0,0 +1,130 @@
...
+  let nodeWithFills: MinimalFillsMixin = selection[0] as MinimalFillsMixin;
+
+  if (nodeWithFills) {
+    if (shader) {
+      selection[0].setSharedPluginData(
+        Utils.SHARED_PLUGIN_NAMESPACE,
+        SHADER_PLUGIN_DATA_KEY,
+        shader
+      );
+    } else {
```
rylin8, line 103: Can this ever get called? I don't see how shader can be undefined, which brings up the question of how to clear the shader from the node?

yiqunw700: Not at the moment. I am planning to add the clear shader menu in a following pr.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: Yeah. It will. I changed it to use the `definition.element.Color` instead of String so it won't need to parse the color every time it draws.

crates/dc_bundle/src/definition/view.rs
yiqunw700: Not really. Using string is pretty straightforward but I realized that every time it draws it calls the parse color which is not very great so I changed it to use the Color struct.

rylin8 (approved):

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
timothyfroehlich: So what happens if the checkbox isn't checked? The text doesn't say what it does.

iamralpht (approved):

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: There is info button on right side of the checkbox explaining that if unchecked it will use the image from the figma node.

timothyfroehlich (approved):

PR #1795, opened 2024-11-15
Convert TextStyle to Proto

### LINKED ISSUE #1734: ProtoConv: Convert view/text_style

timothyfroehlich:
## Tree of downstream PRs as of 2024-11-19

* **PR #1795 (THIS ONE)**:
  `main` ← `wb/froeht/1734-textstyle`

    * PR #1796:
      `wb/froeht/1734-textstyle` ← `wb/froeht/fix-up-meterdata`

      * PR #1806:
        `wb/froeht/fix-up-meterdata` ← `wb/froeht/1736-nodestyle`

Fixes #1734


timothyfroehlich:
The review has sat for a few days so the dcfs are out of date but please review anyways, I'll refetch once I get an approval.

crates/dc_bundle/src/definition/view.rs
```diff
@@ -13,5 +13,115 @@
...
+
+impl StyledTextRun {
+    pub fn new(label: impl ToString) -> StyledTextRun {
+        StyledTextRun { text: label.to_string(), style: Some(TextStyle::new_default()) }
+    }
+    pub fn bold(self) -> Self {
+        StyledTextRun {
+            style: Some(TextStyle {
+                font_weight: Some(FontWeight::BOLD),
+                ..self.style.expect("This was just created")
```
rylin8, line 53: Is this `.expect()` because `style` is now an `Option<>`? What does "This was just created" mean?

timothyfroehlich: Yes, and it's because the structure was created within the same function, so it's very unlikely to have changed. Is there a cleaner way to handle this?

rylin8: I don't think the structure was created within the same function. You could so something like:
```
let test = StyledTextRun::new("hi");
let test_bold = test.bold();
```
Which is fine. If you didn't use the `new()` function and did something like:
```
let test = StyledTextRun { text: "hi".to_string(), style: None };
let test_bold = test.bold();
```
Then it would fail with the `expect()`. But no, I don't know a better way to do this.

iamralpht: It looks like you eliminated this, but I think the normal way would be to wrap the actual type and not an Option (so the outer type models the state), or create a default style when it's None and define the `Default` trait impl for the inner type.

timothyfroehlich: Yeah, it looks like nothing's actually using the functions, so I'm just deleting them.

rylin8 (approved):

PR #1800, opened 2024-11-18
Use Roborazzi to test shader

yiqunw700:
Fixes: #1801


rylin8:
Are the image differences that are not related to your shader change caused by the robolectric version change?

yiqunw700:
> Are the image differences that are not related to your shader change caused by the robolectric version change?

Yes. It looks like `robolectric.pixelCopyRenderMode` is set to `hardware` in 4.14 by default. I didn't make any changes to the `RenderAllExamples` test file but `CustomBrushTest` can run without crashing while it used to throw exception that software render doesn't support runtime shader.

rylin8 (approved):

PR #1803, opened 2024-11-19
Support custom image with context in Squoosh

### LINKED ISSUE #1785: Squoosh ImageWithContext customization support
Refers to the `@Composable (ImageReplacementContext) -> Bitmap?` customization.

rylin8:
Fixes #1785. Change the image with context customization to be a normal function instead of a composable function, and update its usages everywhere.


codegen/src/main/kotlin/com/android/designcompose/codegen/Common.kt
```diff
@@ -184,7 +184,7 @@ internal fun stringTypeToCustomizationType(strType: String): CustomizationType {
         "@Composable (ComponentReplacementContext) -> Unit" ->
             CustomizationType.ComponentReplacement
         "com.android.designcompose.ListContent" -> CustomizationType.ListContent
-        "@Composable (ImageReplacementContext) -> Bitmap?" -> CustomizationType.ImageWithContext
+        "(ImageReplacementContext) -> Bitmap?" -> CustomizationType.ImageWithContext
```
iamralpht, line 187: This must need a docs update, too, right?

reference-apps/aaos-unbundled/mediacompose/src/main/java/com/android/designcompose/reference/mediacompose/MediaAdapter.kt
```diff
@@ -415,6 +420,10 @@ class MediaAdapter(
 
         val currentlyPlaying =
             if (isMetadataSame(item, currentMetadata)) CurrentlyPlaying.On else CurrentlyPlaying.Off
+
+        // Unsubscribe this setter function whenever this leaves the composition
+        DisposableEffect(setIcon) { onDispose { artRequestManager.unsubscribe(setIcon) } }
```
iamralpht, line 425: I think ideally the effect would also perform the subscription (or the subscribe-unsubscribe thing would be encapsulated in a composable function) just so they can't get out of sync. I never remember if `DisposableEffect` executes eagerly, or if it runs after `MediaBrowseItem` finished...).

codegen/src/main/kotlin/com/android/designcompose/codegen/Common.kt
rylin8: Interestingly we never had a section for this customization, but I'll add one.

reference-apps/aaos-unbundled/mediacompose/src/main/java/com/android/designcompose/reference/mediacompose/MediaAdapter.kt
rylin8: This was a bit messy due to the removal of `@Composable`. I subscribe in the image replacement context function because that's when I know the size and color to make the image that I want. I don't have that info here. However to ensure we don't leak memory and have extra subscriptions when this composable leaves composition, I added this `DisposableEffect`.

codegen/src/main/kotlin/com/android/designcompose/codegen/Common.kt
rylin8: Done

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
```diff
@@ -162,13 +167,40 @@ internal fun Modifier.squooshRender(
                         val nodeSize =
                             Size(computedLayout.width * density, computedLayout.height * density)
 
+                        // Check for an image customization with context. If it exists, call the
+                        // custom image function
+                        // and provide it with the frame's background and size.
+                        val customImageWithContext =
```
yiqunw700, line 173: nit: move them inside the `squooshShapeRender` where image customization happens?

rylin8: good point, I think I thought that function was shared with non-squoosh. Fixed now.

designcompose/src/main/java/com/android/designcompose/InteractionState.kt
```diff
@@ -547,7 +547,7 @@ private fun searchNodes(
     }
 
     if (q is NodeQuery.NodeName) {
-        // If the node query is by name, but there are variant properties set, check for a variant
+        // If the node query is by name, but wthere are variant properties set, check for a variant
```
yiqunw700, line 550:
```suggestion
        // If the node query is by name, but there are variant properties set, check for a variant
```
Revert the typo

rylin8: Oops thanks

yiqunw700 (approved):
I don't have better suggestions on the DisposableEffect. The rest of the change looks good to me.

reference-apps/aaos-unbundled/mediacompose/src/main/java/com/android/designcompose/reference/mediacompose/MediaAdapter.kt
rylin8: Resolving so I can get this change in. If you think this needs to be changed, we can file another ticket to make the changed, but I'm not sure how to move the subscription into the @Composable function.

Issue #1801, opened 2024-11-18
Test run time shader using roborazzi


yiqunw700:
- Update robolectric to 4.14(https://github.com/robolectric/robolectric/releases)
- robolectric.pixelCopyRenderMode set to hardware(seems 4.14 set this to hardware by default and this is recommended option in roborazzi: https://github.com/takahirom/roborazzi/blob/cff3c92705807724dfc5a9f0efdb5e34d952aea1/docs/topics/gradle_properties_options.md?plain=1#L58)
- Use sdk 35(Android V)

Issue #1028, opened 2024-04-20
Gradle task fetchAndUpdateFigmaFiles fails on integration tests


rylin8:
When running `./gradlew fetchAndUpdateFigmaFiles` after making a breaking change to the serialized document format, the task fails because the integration tests on `Interaction`, `VariantProperties` and `VariantAnimation` fail. This may be because the tests are run on the old version of the .dcf file instead of run after fetching new ones.

In addition, it would be great if this task could also update:
- Design Switcher .dcf file
- Hello World .dcf file
- Tutorial .dcf file (fetch the welcome one, then rename).

This would make it much easier to update all .dcf files, as I currently have to manually update these 6 files before running the task.

Error example:
```
com.android.designcompose.testapp.validation.AnimationMidpoints > variantAnimation FAILED
    java.lang.AssertionError: Failed to assert the following: (DocRenderStatus = 'Rendered')
    Can't retrieve node at index '0' of 'DocClass = 'com.android.designcompose.testapp.validation.examples.VariantAnimationTestGen''
    There are no existing nodes for that selector.
        at androidx.compose.ui.test.SemanticsNodeInteraction.fetchOneOrDie(SemanticsNodeInteraction.kt:175)
        at androidx.compose.ui.test.SemanticsNodeInteraction.fetchOneOrDie$default(SemanticsNodeInteraction.kt:150)
        at androidx.compose.ui.test.SemanticsNodeInteraction.fetchSemanticsNode(SemanticsNodeInteraction.kt:84)
        at androidx.compose.ui.test.AssertionsKt.assert(Assertions.kt:263)
        at androidx.compose.ui.test.AssertionsKt.assert$default(Assertions.kt:255)
        at com.android.designcompose.test.InstrumentedTestUtilsKt.assertRenderStatus(InstrumentedTestUtils.kt:31)
        at com.android.designcompose.testapp.validation.AnimationMidpoints.variantAnimation(AnimationMidpoints.kt:74)

```

timothyfroehlich:
Assigning to Ralph for priority

yiqunw700:
Closing this ticket with #1611 and #1804 merged. Hello world and tutorial apps can be fetched manually using the scripts.

PR #1804, opened 2024-11-19
Create fetch rule for live update

yiqunw700:
Towards: #1028
Towards: #1801

When working on pr #1800 to create a separate shader test for `CustomBrushTest`, I would like to disable it from the `RenderAllExamples` test but then realized that it would not fetch the dcf file with the gradle task. So I created the fetch rule and made the `ShaderTest` and `StateCustomizationsUnitTest` fetchable in this pr.

Also explored a little bit further to use this fetch rule to make `DesignSwitcherBasicTests` able to fetch dcf file.

Tested locally by running `./gradlew fetchAndUpdateFigmaFiles` and `./gradlew recordRoborazziDebug`


designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt
```diff
@@ -298,15 +299,26 @@ internal fun designSwitcherDocName() = "DesignSwitcherDoc"
...
-private fun getLiveMode(): LiveUpdateMode {
-    return if (DISABLE_LIVE_MODE) {
-        LiveUpdateMode.OFFLINE
+private var enableLiveModeForTesting = false
+
+internal fun enableLiveModeForTesting(enabled: Boolean) {
+    if (Build.FINGERPRINT == "robolectric") {
+        enableLiveModeForTesting = enabled
     } else {
+        // TODO: no exception for integration tests...
```
timothyfroehlich, line 312: Clean up the Todo? I think it's fine as is, it's certainly not an error that should be handled.

timothyfroehlich (approved):
Looks great! Big fan of this change.

rylin8:
Is the removal of the custom brush test because of the separate shader test added previously?

designcompose/src/main/java/com/android/designcompose/DesignSwitcher.kt
yiqunw700: Done.

yiqunw700:
> Is the removal of the custom brush test because of the separate shader test added previously?

Yes.

rylin8 (approved):

Issue #1787, opened 2024-11-14
Support scrolling in squoosh


rylin8:
When scrolling is enabled in Figma's "Scroll behavior" section of the Prototype tab, scrolling should be enabled in the squoosh branch of DesignCompose.

PR #1809, opened 2024-11-25
Support scrolling in squoosh

rylin8:
Fixes https://github.com/google/automotive-design-compose/issues/1787. When a view has scrolling enabled, add it to the list of child composables so that SquooshRoot() can recompose it as its own SquooshRoot(). During this composition, the scrollable view gets a Modifier.scrollable() that handles scroll input. The scroll events are used to calculate a scroll position, and the scroll position is saved into a State<Offset> object that is passed to the renderer and used to translate the children of the scrollable view.

In order to calculate the size of a scrollable view's child contents, the "content_size" feature flag has been turned on in taffy. This allows taffy to calculate the size of a node's children, which we save into new content_width and content_height fields in layout.

A scrolling test, validation example and unit test have been added.


integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/Scrolling.kt
```diff
@@ -0,0 +1,60 @@
...
+@Config(qualifiers = RobolectricDeviceQualifiers.MediumTablet)
+class Scrolling {
+    @get:Rule val clearStateTestRule = TestUtils.ClearStateTestRule()
+    @get:Rule val composeTestRule = createComposeRule()
+    @get:Rule val roborazziRule = designComposeRoborazziRule(javaClass.simpleName)
+    @get:Rule val interFontRule = InterFontTestRule()
+
+    @Test
+    fun scrollTests() {
+        with(composeTestRule) {
```
timothyfroehlich, line 47: Is squoosh enabled for this test?

rylin8: I turned on squoosh, and fixed the test by using replacement content and putting test tags on the content.

timothyfroehlich (approved):

designcompose/src/main/java/com/android/designcompose/InteractionState.kt
```diff
@@ -588,6 +589,15 @@ private fun searchNodes(
             }
         }
     }
+
+    if (q is NodeQuery.NodeId) {
+        // Look for the view in our map of views indexed by node ID.
+        val view = nodeIdMap[q.value]
```
iamralpht, line 595: nit: I think since this function can return null, it would be OK to directly return `nodeIdMap[q.value]`.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
```diff
@@ -111,6 +113,16 @@ internal fun Modifier.squooshRender(
...
+                    // Translate by the scroll amount if the parent node is the scroll view
+                    val scroll = node.parent?.view == parentNode
+                    if (scroll) {
+                        drawContext.canvas.save()
+                        drawContext.canvas.translate(scrollOffset.value.x, scrollOffset.value.y)
+                    }
+                    // Call scrollRestore() whenever returning from renderNode() in order to restore
+                    // the canvas matrix/clip state. This is a good candidate for RAII but this is
+                    // not supported by Kotlin.
+                    val scrollRestore = { if (scroll) drawContext.canvas.restore() }
```
iamralpht, line 125: Do you know if there's a cost to making a closure in Kotlin? We'll make this for every node, not just those that scroll, so it might be better to just duplicate the if statement below (I agree there must be a nicer way to structure this than this kind of manual counting... but I don't know what it is either).

iamralpht (approved):
If I'm understanding this correctly, we break into a new Composable child for any scrollable nodes, and then we get to directly use Compose's `Modifier.scroll`.

Compose implements some different overscroll effects, and I've seen various requests around customizing them. Do you know how to get the Compose stretch and glow effects in these scrollables?

designcompose/src/main/java/com/android/designcompose/InteractionState.kt
rylin8: Done.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
rylin8: Good point, it looks like there is a small cost to capturing all the local variables. I couldn't find a way to do local inline functions, so I'll copy the code and add a comment about it.

rylin8:
> If I'm understanding this correctly, we break into a new Composable child for any scrollable nodes, and then we get to directly use Compose's `Modifier.scroll`.
> 
> Compose implements some different overscroll effects, and I've seen various requests around customizing them. Do you know how to get the Compose stretch and glow effects in these scrollables?

Yes we use `Modifier.scrollable`. I spent quite a bit of time trying to get the overscroll effects and tapping into customizing the fling behavior... but wasn't able to get it working. I didn't want to hold up this change for that, but would really like to get it working.

rylin8:
I messed up rebasing, fixing...

timothyfroehlich:
> I messed up rebasing, fixing...

Yeah, I just merged a few of mine

PR #1826, opened 2024-11-28
Support customizations to get and set scroll position

### LINKED ISSUE #1812: Support customizations to get and set scroll position
Add support for new customizations that allow a developer to set and get the scroll position of a node that has scrolling enabled.

rylin8:
Fixes #1812: Create a DesignScrollCallbacks customization where a developer can set two scroll callbacks. First, a setScrollableState callback that is called to get a ScrollableState object. This object can be used to set the scroll position. Second, a scrollStateChanged callback that is called whenever the scroll state changes, providing the scroll position, max scroll position, frame size and content size.


rylin8:
Going to put this on hold until https://github.com/google/automotive-design-compose/pull/1852 is in. Since I am manually implementing scrolling, I'm able to get the exact scroll position and expose it. However for LazyVerticalGrid, I only have access to `LazyGridState`, which derives from `ScrollableState`. This has access to functions that set the scroll position and get the scroll offset of the first visible item, but nothing to get the actual pixel scroll position. Calculating that would be fairly complicated since items in a grid view can be different sizes and have different column/row spans. I'm going to wait to see if that data is really necessary to expose.

rylin8:
This change will require some documentation, but I would like to get some feedback before adding it in case the API changes.

timothyfroehlich (approved):
LGTM

designcompose/src/main/java/com/android/designcompose/CustomizationContext.kt
```diff
@@ -78,6 +79,18 @@ typealias ShaderUniformTime = Float
 
 typealias ShaderUniformTimeState = FloatState
 
+data class DesignScrollState(
```
yiqunw700, line 82: nit: add some comments

yiqunw700 (approved):

designcompose/src/main/java/com/android/designcompose/CustomizationContext.kt
rylin8: done

rylin8:
I added comments, and will add more in the documentation.

PR #1837, opened 2024-12-04
Convert NodeQuery to Proto

### LINKED ISSUE #1282: Determine how to represent the map of views in the Design Definition in proto
The original rust DesignDefinition includes a map of NodeQueries to Views. Protobuf, however, doesn't let a message to be a key for a map, so we'll need to work out how best to work around that.

_Originally posted by @timothyfroehlich in https://github.com/google/automotive-design-compose/pull/1259#discussion_r1649278420_

### LINKED ISSUE #1670: ProtoConv: Convert NodeQuery
crates/dc_bundle/src/legacy_definition/element/node.rs

timothyfroehlich:
This one is a little more complicated because we use NodeQuery as they key to a map in the DesignCompose Definition and protobuf messages can't use a message as a key.

Fortunately NodeQuery only contains variations of a String, so it's fairly easy to encode it to a string for use in the DCD. We'll still keep the NodeQuery enum and just convert to/from it from the Protobuf message.

Fixes #1670
Fixes #1282


crates/dc_bundle/src/definition.rs
```diff
@@ -22,3 +25,132 @@ pub mod plugin;
...
+    NodeComponentSet(String),
+}
+
+impl NodeQuery {
+    /// Construct a NodeQuery::NodeId from the given ID string.
+    pub fn id(id: impl ToString) -> NodeQuery {
+        NodeQuery::NodeId(id.to_string())
+    }
+
+    /// Construct a NodeQuery::NodeName from the given ID string.
```
rylin8, line 47: I know these comments are copied but some are wrong. Can you fix them? This should be "from the given node name"

crates/dc_bundle/src/definition.rs
```diff
@@ -22,3 +25,132 @@ pub mod plugin;
...
+    pub fn name(name: impl ToString) -> NodeQuery {
+        NodeQuery::NodeName(name.to_string())
+    }
+
+    /// Construct a NodeQuery::NodeVariant from the given node name
```
rylin8, line 52: "from the given variant name and parent name"

crates/dc_bundle/src/definition.rs
```diff
@@ -22,3 +25,132 @@ pub mod plugin;
...
+    pub fn variant(name: impl ToString, parent: impl ToString) -> NodeQuery {
+        NodeQuery::NodeVariant(name.to_string(), parent.to_string())
+    }
+
```
rylin8, line 56: /// Construct a NodeQuery::NodeComponentSet from the given node component set name

crates/figma_import/tests/test_fetches.rs
```diff
@@ -83,3 +83,9 @@ fn fetch_dials_gauges() {
         ],
     );
 }
+
+#[test]
+#[cfg_attr(not(feature = "test_fetches"), ignore)]
+fn fetch_styled_text_runs() {
+    run_test("mIYV4YsYYaMTsBMCVskA4N", &["#MainFrame"])
```
rylin8, line 90: what does this test that isn't tested already?

timothyfroehlich: I had what turned out to be ephemeral fetching issues when I was trying to update the docs, so I had added the test and I think it's worth keeping.

Really I think we should have Rust fetch tests for most to all of our validation docs, but syncing the node names will be an issue so I haven't tried to set up anything like that.

rylin8:
Looks good. We had discussed using JSON serialization to encode/decode the different types of node queries. I like that what you have is simple enough to be able to easily read in a debugger, though there could be issues if someone chooses to name their nodes `name:` or something like that.

rylin8 (approved):

timothyfroehlich:
> Looks good. We had discussed using JSON serialization to encode/decode the different types of node queries. I like that what you have is simple enough to be able to easily read in a debugger, though there could be issues if someone chooses to name their nodes `name:` or something like that.

Oi, I forgot that we decided on JSON. Would have been quicker to do that, though what I ended up with will be more efficient (and readable, like you said). Good catch on the `name:`. I added unit tests for that on both sides and found that I'd forgotten to handle it in the Rust code. Should be fixed now.

PR #1833, opened 2024-12-03
Update localization document with image tool

yiqunw700:
Fixes: #1659

Issue #1843, opened 2024-12-06
Crash when I drag in the Grid Layout Document example


yiqunw700:
```
FATAL EXCEPTION: main
                                                                                                    Process: com.android.designcompose.testapp.validation, PID: 31569
                                                                                                    java.lang.IllegalArgumentException: Cannot coerce value to an empty range: maximum 0.0 is less than minimum 7.0.
                                                                                                    	at kotlin.ranges.RangesKt___RangesKt.coerceIn(_Ranges.kt:1441)
                                                                                                    	at com.android.designcompose.squoosh.SquooshRootKt$SquooshRoot$scrollModifier$1.invoke(SquooshRoot.kt:615)
                                                                                                    	at com.android.designcompose.squoosh.SquooshRootKt$SquooshRoot$scrollModifier$1.invoke(SquooshRoot.kt:604)
                                                                                                    	at androidx.compose.foundation.gestures.ScrollableStateKt$rememberScrollableState$1$1.invoke(ScrollableState.kt:164)
                                                                                                    	at androidx.compose.foundation.gestures.ScrollableStateKt$rememberScrollableState$1$1.invoke(ScrollableState.kt:164)
                                                                                                    	at androidx.compose.foundation.gestures.DefaultScrollableState$scrollScope$1.scrollBy(ScrollableState.kt:184)
                                                                                                    	at androidx.compose.foundation.gestures.ScrollingLogic.performScroll-3eAAhYA(Scrollable.kt:717)
                                                                                                    	at androidx.compose.foundation.gestures.ScrollingLogic.access$performScroll-3eAAhYA(Scrollable.kt:642)
                                                                                                    	at androidx.compose.foundation.gestures.ScrollingLogic$nestedScrollScope$1.scrollByWithOverscroll-OzD1aCk(Scrollable.kt:694)
                                                                                                    	at androidx.compose.foundation.gestures.ScrollableNode$drag$2$1$1.invoke(Scrollable.kt:345)
                                                                                                    	at androidx.compose.foundation.gestures.ScrollableNode$drag$2$1$1.invoke(Scrollable.kt:344)
                                                                                                    	at androidx.compose.foundation.gestures.DragGestureNode$startListeningForEvents$1$1.invokeSuspend(Draggable.kt:437)
                                                                                                    	at androidx.compose.foundation.gestures.DragGestureNode$startListeningForEvents$1$1.invoke(Unknown Source:8)
[...]

Issue #1834, opened 2024-12-03
Make integration tests work in squoosh


rylin8:
Our tests often rely on text tags or test tags on composables, which don't work in squoosh. Find a way around this so that our tests still work when we remove the non-squoosh branch.

Issue #1844, opened 2024-12-06
ANR when try to open design switcher


yiqunw700:
Looks like it get stuck in infinite call loop.
```
at androidx.compose.runtime.CompositionLocalKt.CompositionLocalProvider(CompositionLocal.kt:384)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt.SquooshRoot(SquooshRoot.kt:637)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt$SquooshRoot$8$2$1.invoke(SquooshRoot.kt:694)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt$SquooshRoot$8$2$1.invoke(SquooshRoot.kt:693)
                                                                                                        at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:130)
                                                                                                        at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:51)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt.SquooshChildLayout(SquooshRoot.kt:1073)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt.access$SquooshChildLayout(SquooshRoot.kt:1)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt$SquooshRoot$8.invoke(SquooshRoot.kt:735)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt$SquooshRoot$8.invoke(SquooshRoot.kt:637)
                                                                                                        at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:121)
                                                                                                        at androidx.compose.runtime.internal.ComposableLambdaImpl.invoke(ComposableLambda.kt:51)
                                                                                                        at androidx.compose.runtime.CompositionLocalKt.CompositionLocalProvider(CompositionLocal.kt:384)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt.SquooshRoot(SquooshRoot.kt:637)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt$SquooshRoot$8$2$1.invoke(SquooshRoot.kt:694)
                                                                                                        at com.android.designcompose.squoosh.SquooshRootKt$SquooshRoot$8$2$1.invoke(SquooshRoot.kt:693)
[...]

PR #1846, opened 2024-12-06
Fix crashes related to scrolling

rylin8:
Fixes #1833, #1834. Fix a crash with expanding the design switcher when the scrollable message list is added. This recursive call to SquooshRoot was using the previous IS_DESIGN_SWITCHER policy, causing an infinite recursion composing overlays. Fix another crash with scrollable views whose content size is not large enough to warrant scrolling.


yiqunw700 (approved):

rylin8:
Oops I marked the wrong bugs in this ticket. This fixes #1843  and #1844

PR #1859, opened 2024-12-13
Fix all tests to work with squoosh

rylin8:
Fixes https://github.com/google/automotive-design-compose/issues/1834. Set squoosh as the default renderer and update all tests to work with squoosh. To enable this, two main changes to testing have been made.

1. Instead of testing for text existence using onNodeWithText().assertExists(), add a new function assertHasText(). This works by adding all text rendered under a root node into a hash stored in the root node's Modifier.semantics(). This hash is then checked in the assertHasText() function.

2. All nodes with click/press interactions will get a Modifier.testTag() set with its node name. Tests that perform clicks on nodes use onNodeWithTag() instead of onNodeWithText().


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -653,7 +657,10 @@ fun SquooshRoot(
                         appContext = LocalContext.current,
                         scrollOffset,
                     )
-                    .semantics { sDocRenderStatus = DocRenderStatus.Rendered }
+                    .semantics {
+                        sDocRenderStatus = DocRenderStatus.Rendered
+                        sDocRenderText = textList
```
rylin8, line 662: This sets a list of all the text that is rendered in the tree starting from the root.

test/src/main/kotlin/com/android/designcompose/test/InstrumentedTestUtils.kt
```diff
@@ -40,3 +42,10 @@ fun ComposeContentTestRule.waitForContent(name: String): SemanticsNodeInteractio
         .onFirst()
         .assertRenderStatus(DocRenderStatus.Rendered)
 }
+
+fun SemanticsNodeInteraction.assertHasText(text: String) =
+    assert(
+        SemanticsMatcher("hasText: $text") {
+            it.config.getOrNull(docRenderTextSemanticsKey)?.contains(text) ?: false
```
rylin8, line 49: Checks that the given text is somewhere within the root.

rylin8 (commented):
I wanted to get some thoughts on this approach for testing that some string is rendered. Instead of `onNodeWithText().assertExists()`, I added a function `assertHasText()`. When `SquooshRoot()` goes through the view tree, every text node adds its text to a list, and then `squooshRender()` has a `Modifier.semantics()` that sets the list (can be a hash). This list is used by the function `assertHasText()` to see if it's somewhere in the view tree.

timothyfroehlich:
Sorry, I had a desk booked and someone came by to take the keyboard, I guess some keys got pressed and the PR got closed.

rylin8:
> It won't help everything, but it's a good start.
> 
> Would it be possible to add a semantic property for the original name of each of the nodes in the Figma Doc? This would allow for even better, localization-immune assertions that certain nodes are visible. The node names would probably need to be constructed with their parent names to keep them more unique, since it's so common to re-use node names. (So, for example, in HelloWorld the `#Name` node would be `#MainFrame/NameFrame/#Name`.
> 
> Oh, and at the same time, add a property to list the current variant of a node (if it has variants).

It would be possible, but I also don't want to add too much. There will be a small performance overhead to doing all this. I'd like to do the minimum initially to get squoosh tests working.

rylin8:
> Another idea, would it be possible for the Rust layout code to output a list of all the nodes that are being rendered, including their position, dimensions, node name and so on, such that we could set up a set of empty Compose Boxes that would have that info?
> 
> It would be like how the original renderer did it with each Figma node becoming a Composable, but we wouldn't render anything here. It'd be like an overlay on top that lets you look into how everything is laid out.

Rust doesn't know what nodes are being rendered, only Kotlin does because it knows what the root node is and what all the customizations are. Adding a bunch of empty compose boxes would help with testing, but doing so on top of an implementation that intentionally removes most of the composables seems wrong, and I think would cause other problems.

timothyfroehlich:
>Adding a bunch of empty compose boxes would help with testing, but doing so on top of an implementation that intentionally removes most of the composables seems wrong, and I think would cause other problems. 

I think it could probably help with rotary input and accessibility too, actually

iamralpht:
It would be ideal in this case to have a callback from Compose where libraries can provide the semantics and accessibility info on demand. There's a performance overhead for maintaining a shadow tree, and it also gets hard to keep in sync with scrolling and animations (without really hurting performance).

timothyfroehlich:
> It would be ideal in this case to have a callback from Compose where libraries can provide the semantics and accessibility info on demand. There's a performance overhead for maintaining a shadow tree, and it also gets hard to keep in sync with scrolling and animations (without really hurting performance).

Key thing though, this is only for testing, so you'd only want to enable it for debug builds.

designcompose/src/testDebug/roborazzi/DesignSwitcherBasicTests/CollapsedSwitcher.png
yiqunw700, line 1: The string "Decoded" is cut off

timothyfroehlich (commented):
The screenshot diffs are pretty big, have you reviewed all of them?

designcompose/src/testDebug/kotlin/com/android/designcompose/DesignSwitcherBasicTests.kt
```diff
@@ -76,7 +91,7 @@ class DesignSwitcherBasicTests {
         with(composeTestRule) {
             captureRootRoboImage("CollapsedSwitcher")
             onDCDoc(DesignSwitcherDoc).performClick()
-            onNodeWithText("Change").assertExists()
+            with(onDCDoc(DesignSwitcherDoc)) { assertHasText("Change") }
```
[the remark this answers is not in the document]
rylin8: Done

designcompose/src/testDebug/kotlin/com/android/designcompose/DesignRawResourceTest.kt
```diff
@@ -66,8 +66,10 @@ class DesignRawResourceTest {
             composeTestRule.setContent { HelloWorld() }
 
             onDCDoc(HelloWorldDoc).assertRenderStatus(DocRenderStatus.Rendered)
-            onNodeWithText("Testers!", substring = true).assertExists()
-            onNodeWithText("Hello", substring = true).assertExists()
+            with(onDCDoc(HelloWorldDoc)) {
```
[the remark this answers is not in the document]
rylin8: Done

rylin8:
> The screenshot diffs are pretty big, have you reviewed all of them?

Yes, there are a few main differences:

* Minor pixel shifting: This could be caused by floating point round differences or other minor layout/rendering differences.
* Transparency differences: Squoosh fixes an issue here and is correct.
* Text: Text rendering in squoosh is different, so differences are expected.
* Overlays: Squoosh has some differences with the size of overlays with respect to the size of the root frame. This is a known issue and there is a ticket for this.
* Masks: Squoosh fixes some masking problems
* Design Switcher: Squoosh screenshots were too small so I put the design switcher on top of a grey background, which accounts for most of the differences. There is another issue with the overlay to change the current doc ID not doing some customizations, but this has been an issue for awhile.

timothyfroehlich (approved):

designcompose/src/testDebug/roborazzi/DesignSwitcherBasicTests/CollapsedSwitcher.png
yiqunw700: Can we change the DesignSwitcherBasicTests to use MediumPhone? Any concerns if we change @timothyfroehlich?

timothyfroehlich:
Whichever works is fine with me

yiqunw700 (approved):
Just a minor comment regarding if we can update the design switcher test to use medium phone. But not blocking.

iamralpht (approved):
Looks like a pragmatic approach for now; I also like that we're able to use the interaction Composables for test.

rylin8:
Yea, we can change to medium phone, though I'll do it separately.

PR #1860, opened 2024-12-13
Add functionality to clear shader

yiqunw700:
1. Add a clear all shader menu to clear all shader data from the figma design file
2. Add a clear shader button to clear the shader from the current selection in the plugin UI

Screenshot 2024-12-18 at 11 39 50

Fixes: #1714

### DIFF
--- a/support-figma/extended-layout-plugin/src/shader.ts
+++ b/support-figma/extended-layout-plugin/src/shader.ts
@@ -18,6 +18,8 @@ import * as Utils from "./utils";
 
 const SHADER_PLUGIN_DATA_KEY = "shader";
 const SHADER_FALLBACK_COLOR_PLUGIN_DATA_KEY = "shaderFallbackColor";
+// Private plugin data, used for clearing shader functionalities
+const SHADER_IMAGE_HASH = "shaderImageHash";
 
 export const shaderMap: ReadonlyMap<string, string> = new Map([
   ["cloudy_sky", __uiFiles__.cloudy_sky],
@@ -71,6 +73,7 @@ export async function insertImage(imageBytes: Uint8Array) {
         scaleMode: "FILL",
       },
     ];
+    selection[0].setPluginData(SHADER_IMAGE_HASH, imageHash);
   } else {
     figma.notify(
       "Current selection doesn't support image fill. Please select 1 node that can have an image fill..."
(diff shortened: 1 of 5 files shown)


support-figma/extended-layout-plugin/src/shader.ts
```diff
@@ -131,3 +141,29 @@ export async function setShader(shader: string, shaderFallbackColor: string) {
...
+  Utils.dcLog("Clear shader data from plugin data.");
+
+  await figma.loadAllPagesAsync();
+
+  for (let page of figma.root.children) {
+    for (let child of page.children) {
+      await clearNodeAsync(child);
+    }
+  }
+  figma.closePlugin("Shader data has been cleared.");
```
rylin8, line 156: This clears all shaders from the entire file, so maybe something like "All shaders have been removed".

rylin8 (commented):
I was thinking that there would be a button next to "Apply Shader" to clear a single node of its shader, instead of, or in addition to, this method of clearing all shaders. A user might have a few shaders in a document and then accidentally apply a shader on the wrong node, in which case there's no way to just remove the shader from that node without clearing all shaders.

Secondly, I don't think there is a way right now to tell if a node has a shader set on it. I think that the shader UI should have some indication that the currently selected node has a shader saved in it, and a way to load that shader code into the text edit box.

support-figma/extended-layout-plugin/src/shader.ts
yiqunw700: Done.

yiqunw700:
> I was thinking that there would be a button next to "Apply Shader" to clear a single node of its shader, instead of, or in addition to, this method of clearing all shaders. A user might have a few shaders in a document and then accidentally apply a shader on the wrong node, in which case there's no way to just remove the shader from that node without clearing all shaders.
> 
> Secondly, I don't think there is a way right now to tell if a node has a shader set on it. I think that the shader UI should have some indication that the currently selected node has a shader saved in it, and a way to load that shader code into the text edit box.

Good point.
Added a clear shader button in the plugin ui which doesn't check if there is any shader code set though. There is a simple but not very reliable way is to check if the node uses a shader preview image.
filed a ticket #1877  to track the load shader code in the plugin ui.

rylin8 (approved):

support-figma/extended-layout-plugin/src/shader.ts
```diff
@@ -131,3 +141,66 @@ export async function setShader(shader: string, shaderFallbackColor: string) {
...
+function clearShaderFromNode(node: SceneNode) {
+  let shaderCode = node.getSharedPluginData(
+    Utils.SHARED_PLUGIN_NAMESPACE,
+    SHADER_PLUGIN_DATA_KEY
+  );
+  if (shaderCode && shaderCode.length != 0) {
+    let nodeWithFills: MinimalFillsMixin = node as MinimalFillsMixin;
+
+    if (nodeWithFills) {
+      nodeWithFills.fills = [
```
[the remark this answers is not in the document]
yiqunw700: Per discussion offline, remove shader image preview when clear shader. And if it is text, we will set the text color to the solid pink if there is no background afterwards.

PR #1885, opened 2024-12-18
Support shader float uniform

yiqunw700:


iamralpht:
- Do we need to specify every uniform type in customization? (There are a few vector types and some matrix types, tho those are used less frequently).
- I think we might also want to specify uniforms that aren't customized, but do get animated. Then if you use the same shader as the background in different variants, the auto animation would interpolate between the uniform values.

yiqunw700:
> * Do we need to specify every uniform type in customization? (There are a few vector types and some matrix types, tho those are used less frequently).
> * I think we might also want to specify uniforms that aren't customized, but do get animated. Then if you use the same shader as the background in different variants, the auto animation would interpolate between the uniform values.

I created a new pr #1969 with some new changes which supports color and float uniforms. I believe the new pr will have better extensibility to support other uniforms.

Also filed a new ticket for the component animation feature.

I will close this pr now.

PR #1930, opened 2024-12-20
Convert SquooshAnimate.kt

yiqunw700:
Fixes: #1905

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
+++ b/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
@@ -23,25 +23,23 @@ import androidx.compose.animation.core.VectorConverter
 import androidx.compose.ui.geometry.Size
 import com.android.designcompose.AnimatedAction
 import com.android.designcompose.VariableState
+import com.android.designcompose.android_interface.LayoutChangedResponse
+import com.android.designcompose.android_interface.LayoutChangedResponseKt.layout
 import com.android.designcompose.decompose
+import com.android.designcompose.definition.element.ViewShape
+import com.android.designcompose.definition.element.ViewShapeKt.vectorArc
+import com.android.designcompose.definition.element.arcOrNull
+import com.android.designcompose.definition.element.viewShape
 import com.android.designcompose.definition.view.NodeStyle
 import com.android.designcompose.definition.view.View
 import com.android.designcompose.definition.view.ViewStyle
-import com.android.designcompose.proto.ifContainerGetShape
-import com.android.designcompose.proto.ifTextGetText
-import com.android.designcompose.proto.nodeStyle
-import com.android.designcompose.serdegen.Container
-import com.android.designcompose.serdegen.Layout
-import com.android.designcompose.serdegen.Shape
-import com.android.designcompose.serdegen.VectorArc
-import com.android.designcompose.serdegen.ViewData
-import com.android.designcompose.serdegen.ViewDataType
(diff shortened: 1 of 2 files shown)


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -87,44 +87,48 @@ internal abstract class SquooshAnimatedItem(
...
-                0,
-                width,
-                height,
-                to.left * value + from.left * (1f - value),
-                to.top * value + from.top * (1f - value),
-                to.content_width,
-                to.content_height,
-            )
+        target.computedLayout = layout {
+            this.order = 0
```
timothyfroehlich, line 99: nit: the `this`'s are redundant

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -275,56 +279,31 @@ internal class SquooshAnimatedLayout(
...
-                    from.value.start_angle_degrees * iv + to.value.start_angle_degrees * value,
-                    from.value.sweep_angle_degrees * iv + to.value.sweep_angle_degrees * value,
-                    from.value.inner_radius * iv + to.value.inner_radius * value,
-                    from.value.corner_radius * iv + to.value.corner_radius * value,
-                    to.value.is_mask,
-                )
-            )
-        // Unfortunately, the ViewData and View objects are also immutable.
+        val arc = vectorArc {
+            paths.clear()
```
timothyfroehlich, line 289: nit: I don't think we need to specifically clear the vectors. Their default value is already empty. I'd just cleared them in my other PR because I hadn't included the field names for the other fields.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -275,56 +279,31 @@ internal class SquooshAnimatedLayout(
...
-                .apply {
-                    shape = Optional.of(ViewShape(Optional.of(arc)))
-                    children =
-                        (target.view.data.get().view_data_type.get() as ViewDataType.Container)
-                            .value
-                            .children
-                }
+            target.view.data.container
+                .toBuilder()
+                .apply { shape = viewShape { this.arc = arc } }
```
timothyfroehlich, line 302: Nit: pretty sure just `arc` is sufficient. At least it is in rust, when the variable you're assigning has the same name as the field.

timothyfroehlich (approved):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: Done.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: Done. Removed.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: I don't quite think the kotlin apply block can achieve this. 
https://pl.kotl.in/bPOySIRSB

Anyway. I will keep the assignment.

PR #1934, opened 2024-12-20
Convert SquooshRender.kt

yiqunw700:
Fixes: #1908

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
+++ b/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
@@ -41,26 +41,22 @@ import com.android.designcompose.CustomizationContext
 import com.android.designcompose.DocContent
 import com.android.designcompose.TextMeasureData
 import com.android.designcompose.VariableState
+import com.android.designcompose.android_interface.LayoutChangedResponse
+import com.android.designcompose.definition.modifier.TextAlignVertical
+import com.android.designcompose.definition.modifier.TextOverflow
 import com.android.designcompose.definition.view.ViewStyle
+import com.android.designcompose.definition.view.strokeOrNull
+import com.android.designcompose.definition.view.textColorOrNull
+import com.android.designcompose.definition.view.transformOrNull
 import com.android.designcompose.getBrush
 import com.android.designcompose.getBrushFunction
-import com.android.designcompose.proto.blendModeFromInt
-import com.android.designcompose.proto.getType
-import com.android.designcompose.proto.nodeStyle
-import com.android.designcompose.proto.textAlignVerticalFromInt
-import com.android.designcompose.proto.textOverflowFromInt
-import com.android.designcompose.proto.toUniform
-import com.android.designcompose.serdegen.Layout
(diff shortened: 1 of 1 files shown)


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
```diff
@@ -333,12 +332,13 @@ private fun squooshTextRender(
     drawContext.canvas.save()
     drawContext.canvas.translate(0.0f, verticalCenterOffset)
 
-    val strokeWidth = style.nodeStyle.stroke.get().stroke_weight.toUniform() * density.density
+    val strokeWidth =
+        (style.nodeStyle.strokeOrNull?.strokeWeightOrNull?.toUniform() ?: 0f) * density.density
```
timothyfroehlich, line 336: I think in this case it's better to use the normal proto behavior to return a default values (which is 0) if it wasn't present, like this:
```suggestion
        (style.nodeStyle.stroke.strokeWeight.toUniform()) * density.density
```

What do you think, @rylin?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
```diff
@@ -333,12 +332,13 @@ private fun squooshTextRender(
...
     // Only drop shadow is supported now.
-    val shadow = style.nodeStyle.text_shadow.getOrNull()
-    val blurRadius = (shadow?.blur_radius ?: 0f) * density.density
-    val xOffset = (shadow?.offset_x ?: 0f) * density.density
-    val yOffset = (shadow?.offset_y ?: 0f) * density.density
+    val shadow = style.nodeStyle.textShadowOrNull
+    val blurRadius = (shadow?.blurRadius ?: 0f) * density.density
+    val xOffset = (shadow?.offsetX ?: 0f) * density.density
+    val yOffset = (shadow?.offsetY ?: 0f) * density.density
```
timothyfroehlich, line 341: Should be able to handle this with proto returning defaults:

```suggestion
    val shadow = style.nodeStyle.textShadow
    val blurRadius = shadow.blurRadius * density.density
    val xOffset = shadow.offsetX * density.density
    val yOffset = shadow.offsetY * density.density
```

timothyfroehlich (approved):

timothyfroehlich (approved):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
rylin8: It's a bit cleaner to use the default values, but using the `orNull` fields make it explicitly clear that those values can be null...

rylin8: Done

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRender.kt
rylin8: Done

PR #1935, opened 2024-12-20
Convert SquooshText.kt

yiqunw700:
Fixes: #1911

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
+++ b/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
@@ -38,29 +38,30 @@ import com.android.designcompose.DesignSettings
 import com.android.designcompose.DocContent
 import com.android.designcompose.TextMeasureData
 import com.android.designcompose.VariableState
+import com.android.designcompose.definition.element.LineHeight.LineHeightTypeCase
+import com.android.designcompose.definition.element.TextDecoration
+import com.android.designcompose.definition.element.weightOrNull
 import com.android.designcompose.definition.view.View
+import com.android.designcompose.definition.view.fontSizeOrNull
+import com.android.designcompose.definition.view.fontWeightOrNull
+import com.android.designcompose.definition.view.styleOrNull
+import com.android.designcompose.definition.view.textColorOrNull
+import com.android.designcompose.definition.view.textShadowOrNull
 import com.android.designcompose.getText
 import com.android.designcompose.getTextState
 import com.android.designcompose.getTextStyle
 import com.android.designcompose.getValue
-import com.android.designcompose.proto.fontStyleFromInt
-import com.android.designcompose.proto.getType
-import com.android.designcompose.proto.nodeStyle
-import com.android.designcompose.proto.textAlignFromInt
-import com.android.designcompose.proto.textDecorationFromInt
-import com.android.designcompose.serdegen.LineHeightType
-import com.android.designcompose.serdegen.TextDecoration
-import com.android.designcompose.serdegen.ViewDataType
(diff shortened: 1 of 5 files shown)


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
```diff
@@ -140,157 +140,136 @@ internal fun squooshComputeTextInfo(
...
+                builder.append(normalizeNewlines(getTextContent(appContext, v.data.text)))
+                builder.toAnnotatedString()
+            } else if (v.data.hasStyledText()) {
+                val builder = AnnotatedString.Builder()
+                for (run in getTextContent(appContext, v.data.styledText)) {
+                    val style =
+                        run.styleOrNull
+                            ?: throw NoSuchFieldException(
+                                "Malformed data: StyledTextRun's style unset"
+                            )
```
timothyfroehlich, line 155: I think we can remove these `orNull ?: throw exception` bits. These fields are meant to be required, so we don't need to "dereference" the Java Optional type. If we did want to do some checking to make sure that required fields are set, then I think it'd be best to do so in a centralized set of functions.
```suggestion
                        run.style
```

timothyfroehlich (approved):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
```diff
@@ -140,157 +140,136 @@ internal fun squooshComputeTextInfo(
...
-                    .font_weight
-                    .get()
-                    .weight
-                    .get()
-                    .getValue(variableState)
-                    .roundToInt()
+                v.style.nodeStyle.fontWeightOrNull
+                    ?.weightOrNull
+                    ?.getValue(variableState)
+                    ?.roundToInt() ?: 400
```
rylin8, line 227: I'd prefer declaring some of these default values in constants somewhere.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
```diff
@@ -140,157 +140,136 @@ internal fun squooshComputeTextInfo(
...
+                color = textShadow.color.getValue(variableState) ?: Color.Transparent,
             )
         }
     // The brush and opacity is computed later at rendering time.
     val textStyle =
         (TextStyle(
             fontSize =
                 customTextStyle?.fontSize
-                    ?: v.style.get().nodeStyle.font_size.get().getValue(variableState).sp,
+                    ?: (v.style.nodeStyle.fontSize?.getValue(variableState) ?: 18f).sp,
```
rylin8, line 269: This one too

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
rylin8: I agree with making them centralized if possible. I'm not sure we need these either, but perhaps they'd be useful for debugging purposes. I haven't used this to really know if this might be helpful or not.

rylin8: Done

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
rylin8: Done

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
rylin8: Done

PR #1941, opened 2024-12-23
Convert the encoded DCF and ServerFigmaDocs to Proto

timothyfroehlich:
# Based on PR #1886

## Chain of upstream PRs as of 2024-12-23

* PR #1886:
  `main` ← `feature/protoconv`

  * **PR #1941 (THIS ONE)**:
    `feature/protoconv` ← `wb/froeht/protolize-dcf`

This PR should convert all of our code that serializes and deserializes DCF files and the ServerFigmaDoc to encode them as Proto messages (mostly). This includes sending the doc from figma_import to the kotlin code via JNI.

The old DCF format consisted of the bincode-encoded header, bincode-encoded doc, and JSON-encoded image session all concatenated together. I've simply converted the bincode encoding to proto encoding, using the `encode/decodeDelimited()` functions that proto provides to write the length of the encoded message first, followed by the message. This results in pretty clean code on both sides, especially the Kotlin side which can just read through the input stream.

I've added some tests in Common to decode and re-encode the DesignSwitcher and HelloWorld docs. These can be run right now, because we haven't broken the Common code. I've also enhanced the Figma_import fetch_tests to write the fetched doc to a file and read it back in.

You can test these by running `./gradlew common:test` and `cargo test --all-features`


common/build.gradle.kts
```diff
@@ -38,6 +38,14 @@ publishing {
     }
 }
 
+sourceSets{
+    test {
```
timothyfroehlich, line 42: It'd be better to move the common assets that we use in multiple places, like HelloWorld and DesignSwitcher, to a common directory, but I'd rather not risk a change like that while I can't check to make sure I didn't break anything.

crates/figma_import/src/fetch.rs
```diff
@@ -118,19 +118,22 @@ pub fn fetch_doc(
...
+        let server_doc = ServerFigmaDoc {
             figma_doc: Some(figma_doc),
             errors: error_list,
             branches: doc.branches.clone(),
             project_files: vec![],
-        })?);
+        };
 
+        let mut response = header.encode_length_delimited_to_vec();
+        response.append(&mut server_doc.encode_length_delimited_to_vec());
```
rylin8, line 136: Can you just call `encode_dcd_with_header()` instead

crates/figma_import/src/tools/fetch.rs
```diff
@@ -158,15 +165,16 @@ pub fn fetch(args: Args) -> Result<(), ConvertError> {
...
-    let header = bincode::serialize(&DesignComposeDefinitionHeader::current(
-        doc.last_modified().clone(),
-        doc.get_name().clone(),
-        doc.get_version().clone(),
-        doc.get_document_id().clone(),
-    ))?;
-    let doc = bincode::serialize(&dc_definition)?;
-    output.write_all(header.as_slice())?;
-    output.write_all(doc.as_slice())?;
+    // let mut output = std::fs::File::create(args.output)?;
```
rylin8, line 168: delete

crates/dc_bundle/src/legacy_definition.rs
```diff
@@ -37,45 +29,3 @@ impl DesignComposeDefinitionHeaderV0 {
         22
     }
 }
```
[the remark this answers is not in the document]
timothyfroehlich: I don't think so, because it's still part of the old legacy code, essentially.

crates/figma_import/src/fetch.rs
timothyfroehlich: Not in this case because it's the ServerFigmaDoc and I didn't make that function for it. This is the only place we do it so I don't think it's worth it here.

rylin8 (approved):

PR #1943, opened 2024-12-26
Delete the no-longer-necessary proto utils

timothyfroehlich:
# Based on PR #1886

## Chain of upstream PRs as of 2024-12-26

* PR #1886:
  `main` ← `feature/protoconv`

  * **PR #1943 (THIS ONE)**:
    `feature/protoconv` ← `wb/froeht/delete-proto-utils`


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
```diff
@@ -150,30 +146,27 @@ internal fun squooshComputeTextInfo(
             } else if (v.data.hasStyledText()) {
                 val builder = AnnotatedString.Builder()
                 for (run in getTextContent(appContext, v.data.styledText)) {
-                    val style = run.styleOrNull.getOrThrow("StyledTextRun")
+                    val style = run.style
```
rylin8, line 149: I'd probably keep these around for a little while, and delete them later once we are stable.

timothyfroehlich: What about replacing them with !!? Does basically the same thing and I don't need to restore those functions.

rylin8: I'm ok trying that. If it gets hit a lot I'd consider putting the function back in.

rylin8 (approved):

Issue #1945, opened 2024-12-26
Fix the VariantAnimationTimelineTest


timothyfroehlich:
It tries to access parts of the DesignDoc that aren't meant to be accessed, and fails to compile because Validation does not have the protobuf libraries as a dependency. We don't want to expose the Protobuf API, because this can lead to compilation problems when another library uses a different version of the Protobuf library. We'll need to change the test to resolve the compilation error without modifying it's Gradle dependencies.

PR #1946, opened 2024-12-26
Temporarily disable the VariantAnimationTimelineTest.

timothyfroehlich:
# Based on PR #1943

## Chain of upstream PRs as of 2024-12-26

* PR #1886:
  `main` ← `feature/protoconv`

  * PR #1941:
    `feature/protoconv` ← `wb/froeht/protolize-dcf`

    * PR #1943:
      `wb/froeht/protolize-dcf` ← `wb/froeht/delete-proto-utils`

      * **PR #1946 (THIS ONE)**:
        `wb/froeht/delete-proto-utils` ← `wb/froeht/temp-disable-variant-animation-test`

It tries to access parts of the DesignDoc that aren't meant to be accessed, and fails to compile because Validation does not have the protobuf libraries as a dependency. We don't want to expose the Protobuf API, because this can lead to compilation problems when another library uses a different version of the Protobuf library. We'll need to change the test to resolve the compilation error without modifying it's Gradle dependencies.

Task to fix it is #1945


rylin8:
What part of DesignDoc does it try to access? Please use multi-line comments /* */ so we don't see a line changed in the history of the file for every single line. It makes blame harder to use.

timothyfroehlich:
> What part of DesignDoc does it try to access? Please use multi-line comments /* */ so we don't see a line changed in the history of the file for every single line. It makes blame harder to use.

Something related to line 82 of VariantAnimationTimelineTest: 
```
                    val regexMatch = tNumRegex.find(context.from?.name ?: "")
```

Good call on using multi-line comments.

rylin8:
I'm fine fixing this test later, but I don't really get what the issue is with accessing part of the design doc. It just uses the context passed into the custom animation function to see what the source and destination node names are, and that's passed in from DesignCompose in the animation code.

rylin8 (approved):

PR #1953, opened 2024-12-30
Initialize an default ViewStyle while generating replacement list child nodes

timothyfroehlich:


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshTreeBuilder.kt
```diff
@@ -170,6 +176,8 @@ internal fun resolveVariantsRecursively(
     var overrideStyle: ViewStyle? = null
     var view = viewFromTree
 
+    assert(view.style.layoutStyle.hasPadding())
```
rylin8, line 179: This assert still needed? I think it can be removed.

timothyfroehlich: Sorry, hadn't cleaned it up yet.

rylin8 (approved):

PR #1964, opened 2025-01-09
Fix: Improve test setup for DocSerializationTest

timothyfroehlich:
# Based on PR #1886

## Chain of upstream PRs as of 2025-01-09

* PR #1886:
  `main` ← `feature/protoconv`

  * **PR #1964 (THIS ONE)**:
    `feature/protoconv` ← `wb/froeht/improve-docserializationtest`

This change modifies the test setup for `DocSerializationTest` to improve its robustness and coverage.

-   Removes `showStandardStreams = true` from the roborazzi gradle config, to quiet down the amount of console spam
-   Adds the validation assets directory to the test resources for the common module.
-   Replaces the single `loadSaveLoadHelloWorld` test with a parameterized test `LoadSaveLoadAllDocsTest` to make it easier to see which docs aren't passing
-   Removes the `VariantAnimationTimelineTestDoc_vJRf4zxY4QX4zzSSUd1nJ5.dcf` file. The test is disabled right now ( #1945 ) so the dcf wasn't being updated and was failing to decode.


common/src/test/kotlin/com/android/designcompose/common/DocSerializationTest.kt
```diff
@@ -49,50 +52,40 @@ class DocSerializationTest {
...
         val inputStream: InputStream =
-            DocSerializationTest::class
+            LoadSaveLoadAllDocsTest::class
                 .java
                 .classLoader!!
-                .getResourceAsStream("figma/HelloWorldDoc_pxVlixodJqZL95zo2RzTHl.dcf")!!
-        assertNotNull(inputStream)
-        val doc = decodeDiskBaseDoc(inputStream, DesignDocId("loadSaveLoad"), Feedback)
+                .getResourceAsStream("figma/$fileName")!!
+        val doc = decodeDiskBaseDoc(inputStream, DesignDocId(fileName), Feedback)
```
rylin8, line 83: I don't quite follow how this works -- does this iterate through all  the validation test dcf files? Is this test called multiple times or does the test load multiple files?

timothyfroehlich: It's a Junit4 parameterized test, just like the main Validation render all test. The `data()` function scans through all the .dcf files that were put in the test's resources (by the Gradle config) and instantiates an instance of the test for each one. 

Before this change the test loaded multiple files, which meant that if any file failed you'd need to go check stdout to see which file failed, and you wouldn't see if anything else failed. With this change it'll instantiate an instance of the test for each file, named after the file.

I had Gemini add some comments for future reference.

rylin8 (approved):

PR #1965, opened 2025-01-09
Convert the Convert Request/Response types to Proto, refactor Layout conversion code

timothyfroehlich:
# Based on PR #1886

## Chain of upstream PRs as of 2025-01-09

* PR #1886:
  `main` ← `feature/protoconv`

  * **PR #1965 (THIS ONE)**:
    `feature/protoconv` ← `wb/froeht/protolize-convertresponse`

With this change we're able to fully remove serde-reflection from the code base.

This involved a decent amount of refactoring and cleanup of old work to get things right. My goal with the refactoring was to keep the packages cleanly separated and to work towards having the ability to build a DesignCompose app without including anything related to Live Update. Ultimately I could see the dc_jni crate being split into the layout and fetch portions, which would compile into their own separate libraries, and merged in with figma_import and layout.

Some of the messages related to both the doc fetching and layout processing were a bit scattered and were declared in crates that didn't actually use them. Before this change the `dc_jni` crate didn't have any of its own proto messages, it was pulling the ones it needed (Layout conversion related) from other crates (both `dc_bundle` and `layout`). The new Convert Request/Response messages would only be used by `dc_jni`, so it made sense to add a build.rs file to it so that the generated messages would be native to that crate, rather than being part of an unrelated crate. 

While making that change it made sense to rearrange the proto messages related to the android interface, so I moved the layout related messages into a `layout_interface` proto package and put the new Convert messages into the `live_update` package.

In addition, the layout code had really only been half converted to proto, we were still parsing them and then converting them into the original Rust structs. That's all cleaned up now.

We had been manually generating a json string which was decoded into the ConvertRequest on the rust side, so I converted the ConvertRequest to Proto and removed the json code. It didn't make sense to use proto on one side and not the other. It'd probably be cleaner to change the `jniFetchDoc` method to just send a single proto message instead of the ConvertRequest, doc_id, version_id and proxy configuration, but that can be cleaned up another day. 

I also decided not to convert the image_session from json to proto because we don't actually do anything with it on the Kotlin side other than writing it into the .dcf. I'm simply including the json in the proto message and removing the redundant storage of the session's bytes in the docContent class.

Admittedly this PR is bigger than it strictly needed to be. I can split it into some smaller PRs if requested. I'm sure there's some additional cleanup that could be done as well, including simplifying the `jniFetchDoc` function and turning ProxyConfig into a proto struct.


crates/figma_import/src/proxy_config.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2023 Google LLC
```
rylin8, line 1: Why back a year?

crates/layout/src/layout_manager.rs
```diff
@@ -330,7 +329,10 @@ impl LayoutManager {
         if compute_layout {
             self.compute_node_layout(root_layout_id)
         } else {
-            LayoutChangedResponse::unchanged(self.layout_state)
+            LayoutChangedResponse {
+                layout_state: self.layout_state,
+                changed_layouts: HashMap::new(),
+            }
```
rylin8, line 335: `LayoutChangedResponse::unchanged()` ?

crates/figma_import/src/proxy_config.rs
timothyfroehlich: Odd. I think Git guessed wrong about which was the original file. This used to be fetch.rs.

crates/layout/src/layout_manager.rs
timothyfroehlich: Fixed

rylin8 (approved):

Issue #1830, opened 2024-12-02
Shader Plugin - support custom uniforms


yiqunw700:
float, int, color as float3 and float4, other use case for float2, float3 and float4.

PR #1969, opened 2025-01-10
Support shader float/color uniform

yiqunw700:
Towards: #1830 
Support float and color uniform.

Type set by plugin: float -> float, color -> float3, color with alpha -> float4
Shader API: 
- float->`setFloatUniform(name, value)`
- color->`setFloatUniform(name, value1, value2, value3)`
- color with alpha -> `setFloatUniform(name, value1, value2, value3, value4)`

Parse:
shader_schema.rs: `ShaderUniformJson` -> `ShaderUniform`

Customization supports:
- `List<ShaderUniform>`
- `List<State<ShaderUniform>>`

Helper functions to create:
- `float` -> `ShaderUniform`
- `FloatState`->`State<ShaderUniform>`
- `Color`->`ShaderUniform`
- `State<Color>`->`State<ShaderUniform>`


designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -503,19 +506,43 @@ internal fun ContentDrawScope.squooshShapeRender(
...
+                    it.shaderUniformsMap.forEach { (_, v) ->
+                        v.applyToShader(shader, it.shaderUniformsMap)
+                    }
+                    customizations.getShaderUniformList(name)?.forEach { customUniform ->
+                        if (customUniform.name == ShaderHelper.UNIFORM_TIME) {
+                            shader.setFloatUniform(
+                                customUniform.name,
+                                customUniform.value.floatValue,
+                            )
+                            return@forEach
```
rylin8, line 525: Hmm I don't follow this -- why not call `applyToShader` for the time uniform, and why return?

crates/figma_import/src/shader_schema.rs
```diff
@@ -0,0 +1,89 @@
...
+    pub uniform_type: String,
+    #[serde(rename = "uniformValue")]
+    pub uniform_value: serde_json::Value,
+}
+
+impl Into<(String, ShaderUniform)> for ShaderUniformJson {
+    fn into(self) -> (String, ShaderUniform) {
+        let uniform_value = match self.uniform_type.as_str() {
+            "float" => {
+                if let Some(float_val) = self.uniform_value.as_f64() {
```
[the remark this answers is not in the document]
yiqunw700: serde_json::Value only has as_f64(). It doesn't have a as_f32. So I have to do as_f64() then as f32.

support-figma/extended-layout-plugin/src/shader.ts
```diff
@@ -138,6 +146,27 @@ function setShaderToNode(
...
+        console.log(JSON.stringify(shaderUniform));
+      }
+
+      console.log(JSON.stringify(shaderUniforms));
+
+      node.setSharedPluginData(
+        Utils.SHARED_PLUGIN_NAMESPACE,
+        SHADER_UNIFORMS_PLUGIN_DATA_KEY,
+        JSON.stringify(shaderUniforms)
+      );
```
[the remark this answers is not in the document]
yiqunw700: a task saved for later along with loading the shader code.

Right now there is no obvious way to do that. I am planning to add a "load shader" button in the plugin ui by the side of the node id. And only when the shader is set, the button shows up.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushFromShaderPluginTest.kt
```diff
@@ -16,37 +16,62 @@
...
-        } else {
-            remember { mutableFloatStateOf(3.0f) }
-        }
+    val iTimeState = ShaderHelper.getShaderUniformTimeState()
+    val rootShaderUniformStates = ArrayList<State<ShaderUniform>>()
+    rootShaderUniformStates.add(iTimeState.toShaderUniformState(ShaderHelper.UNIFORM_TIME))
+    val rootShaderUniforms = ArrayList<ShaderUniform>()
+    rootShaderUniforms.add(ShaderHelper.createShaderFloatUniform("speed", 0.5f))
+    rootShaderUniforms.add(ShaderHelper.createShaderFloatUniform("clouddark", 0.5f))
+    rootShaderUniforms.add(ShaderHelper.createShaderFloatUniform("cloudcover", 0.23f))
```
rylin8, line 56: Is there a way to see what these uniforms do? I guess, is there a way to see the shader code?

support-figma/extended-layout-plugin/src/shader.html
```diff
@@ -112,6 +172,245 @@
 
     let currentImageSnapshot = null;
 
+    const customUniforms = [];
+    const uniformTypeMap = new Map();
+    const uniforms = document.getElementById("uniforms");
+
+    const createUniformDialog = document.getElementById("createUniformDialog");
+    const continueButton = document.getElementById("continueButton");
```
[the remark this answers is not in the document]
yiqunw700: Good catch. Done.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: The `return` will continue to the next uniform in the iterator.

`iResolution` and `iTime` are two preset uniforms. Right now I have them always present in the shader code. 
For `iTime` I made it customizable because the shader code may use it or not use it. But it is not created by the user in the plugin ui and it is not in the `shaderUniformsMap`

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushFromShaderPluginTest.kt
yiqunw700: This one is the cloud example in the plugin ui with changing speed, clouddark and cloudcover from `const float` to `uniform float`.

You will be able to see the shader code when I have the load shader code in plugin ui supported.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushFromShaderPluginTest.kt
```diff
@@ -16,37 +16,62 @@
...
+            Color(LocalContext.current.getColor(R.color.purple_700)),
+        )
+    val colorState = remember { derivedStateOf { colors[iTimeState.floatValue.toInt() % 2] } }
+    customColorStates.add(colorState.composeColorToShaderUniformState("iColor"))
     BrushFromShaderPluginTestDoc.MainFrame(
-        backgroundShaderUniformTimeState = movingValue.asFloatState()
+        rootShaderUniformStates = rootShaderUniformStates,
+        rootShaderUniforms = rootShaderUniforms,
+        customColors = customColors,
+        customColorStates = customColorStates,
```
rylin8, line 75: So it looks like you have the uniforms that don't change in the list of uniforms and the uniforms that change a lot in the list of uniform states. I assume this is so that when the uniform state values change, only the FrameRender function needs to re-render, and recomposition is not needed. Can you verify that? If the uniform values change, does `BrushFromShaderPluginTestDoc.MainFrame()` get recomposed?

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
rylin8: Ok I see, can you add a bit of this information as a comment when adding the `iTime` shader uniform?

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -503,19 +506,43 @@ internal fun ContentDrawScope.squooshShapeRender(
...
+                        }
+                        if (it.shaderUniformsMap.containsKey(customUniform.name)) {
+                            customUniform.applyToShader(shader, it.shaderUniformsMap)
+                        }
```
rylin8, line 542: Would it be helpful to print out a warning in an `else` clause here saying that a uniform has been set that was never set in the plugin?

designcompose/src/main/java/com/android/designcompose/ShaderHelper.kt
```diff
@@ -0,0 +1,102 @@
...
+import androidx.compose.runtime.produceState
+import androidx.compose.runtime.remember
+import com.android.designcompose.definition.element.floatColor
+import com.android.designcompose.definition.view.ShaderUniform
+import com.android.designcompose.definition.view.shaderUniform
+import com.android.designcompose.definition.view.shaderUniformValue
+
+object ShaderHelper {
+    const val UNIFORM_TIME = "iTime"
+
```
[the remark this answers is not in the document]
yiqunw700: Done.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: Done.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: Done. I added as debug log instead of warning. Because it will print out every time it renders, it will flush the logcat if the shader animates over a changing state customization, so it would be good to trim those in the production code.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushFromShaderPluginTest.kt
yiqunw700: I added some log in SquooshRoot. It doesn't keep printing. So I believe it is not re-composed.

rylin8 (approved):

PR #1971, opened 2025-01-10
Fix VariantAnimationTimelineTest

### LINKED ISSUE #1945: Fix the VariantAnimationTimelineTest

rylin8:
Fixes #1945: Add functions to VariantTransitionContext that allow access to the view names.


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshVariantTransition.kt
```diff
@@ -130,7 +130,27 @@ internal class VariantAnimationInfo(
     val transition: AnimationTransition,
 )
 
-class VariantTransitionContext(val from: View?, val to: View?) {
+class VariantTransitionContext(private val from: View?, private val to: View?) {
+    @Discouraged(
+        message =
+            "This helper function will the name of the `from` View. This function is intended to " +
```
timothyfroehlich, line 136: typo: "This helper function will the name of the `from` View"

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshVariantTransition.kt
```diff
@@ -130,7 +130,27 @@ internal class VariantAnimationInfo(
...
+                "temporarily help with creating custom variant transitions, but may be deprecated or " +
+                "removed in the future."
+    )
+    fun getFromViewName(): String {
+        return from?.name ?: ""
+    }
+
+    @Discouraged(
+        message =
+            "This helper function will the name of the `to` View. This function is intended to " +
```
timothyfroehlich, line 146: Same typo

timothyfroehlich (approved):
Thanks for making this change :)

rylin8:
Wasn't sure about moving forward with this since you also did https://github.com/google/automotive-design-compose/pull/1970

timothyfroehlich:
Your's is the better once since it means we're exposing fewer dependencies on our API surface. I closed mine

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshVariantTransition.kt
rylin8: fixed

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshVariantTransition.kt
rylin8: fixed

timothyfroehlich (approved):

PR #1975, opened 2025-01-13
Revert "Revert "Switch to generating and validating Roborazzi screens…

timothyfroehlich:
…hots to …"

This reverts commit 88945aaaaeeda8e4f6dce4ab2122e128c2966132.


yiqunw700 (approved):

rylin8:
Can this be a normal commit instead of a revert of a revert, also so that that merge commit in here is removed?

PR #1993, opened 2025-01-13
Fix the UnsupportedOperationException

yiqunw700:
Protobuf returns unmodifiable map so we can not update the image map in the document.

Split the images map into a DesignComposeDefinitionExtras and when save the serialized doc to disk, we will save the extras built from the mutable images map.


designcompose/src/main/java/com/android/designcompose/DocContent.kt
```diff
@@ -49,19 +50,29 @@ class DocContent(var c: GenericDocContent, previousDoc: DocContent?) {
...
                 val byteArray = ByteArray(bytes.size()) { i -> bytes[i] }
                 val bitmap = BitmapFactory.decodeByteArray(byteArray, 0, byteArray.size)
                 images[imageKey] = bitmap
             }
         }
+        c.document =
+            c.document.copy {
+                images.clear()
+                images.putAll(mutableImagesMap)
+            }
```
rylin8, line 75: I stared at this change for so long, but I think it makes sense now!

rylin8 (approved):

designcompose/src/main/java/com/android/designcompose/DocContent.kt
yiqunw700: I have a new approach. The document copy might not be very efficient. Please take another look.

designcompose/src/main/java/com/android/designcompose/DocContent.kt
```diff
@@ -46,15 +47,20 @@ class DocContent(var c: GenericDocContent, previousDoc: DocContent?) {
...
+            if (bytes.isEmpty) {
+                if (
+                    previousDoc != null &&
+                        previousDoc.images[imageKey] != null &&
+                        previousDoc.imagesMap[imageKey] != null
+                ) {
                     images[imageKey] = previousDoc.images[imageKey]!!
                     // Replace this record in the decoded doc.
-                    c.document.imagesMap[imageKey] = previousDoc.c.document.imagesMap[imageKey]
+                    imagesMap[imageKey] = previousDoc.imagesMap[imageKey]!!
```
rylin8, line 63: You're modifying `imagesMap` while iterating through it -- I'm pretty sure this is safe since you're just modifying the value and not deleting it, but just wanted to point it out. In certain scenarios `ConcurrentModificationException` can be thrown but I think this is fine.

rylin8 (approved):

yiqunw700:
Ok. I agree that if it is just live update, copy being slow might not be a big deal. I will stick to the easy fix in #1997. I will probably do the proto split if there is stronger reason to do the proto split.

PR #1996, opened 2025-01-15
Fix TapCallback on nodes that change from an interaction

### LINKED ISSUE #1995: Layout replacement test is broken
Nothing happens when click on the next button

rylin8:
Fixes #1995. When a node has both an interaction that changes the variant and a TapCallback, the TapCallback was not being found because it was customized using the node's original name. This has been fixed with the following changes:

1. Add an isPressed mutableState that gets set to true when the press happens, and false when released. This causes a recomposition which builds a new UI tree using the new variant that has changed.
2. Add a set of pressed node IDs in interaction set so that when the above recomposition happens, the new UI tree with the changed variant still finds a supported interaction and adds itself to the root node's composable child list.

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/InteractionState.kt
+++ b/designcompose/src/main/java/com/android/designcompose/InteractionState.kt
@@ -224,6 +224,12 @@ internal class InteractionState {
     var navOverlaySubscriptions: ArrayList<() -> Unit> = ArrayList()
     var variantSubscriptions: HashMap<String, ArrayList<() -> Unit>> = HashMap()
     var animationSubscriptions: ArrayList<() -> Unit> = ArrayList()
+
+    // A set of node IDs that have been pressed due to an interaction or TapCallback. This is
+    // necessary to keep track of so that nodes that change to another variant due to an
+    // interaction can still be found to have a press/click/tap event on them and be added to
+    // SquooshRoot's list of composable children.
+    var isPressed: HashSet<String> = HashSet()
 }
 
 /// Perform the "navigate" action, by appending the given node id to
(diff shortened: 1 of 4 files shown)


yiqunw700:
I also noticed some other bug introduced by the protoconv that it should not trigger animation.

https://github.com/google/automotive-design-compose/blob/d956f8901dff390cf6ee25cbc16467adf08a0649/designcompose/src/main/java/com/android/designcompose/InteractionState.kt#L404

transition will never be null now. We should use transitionOrNull in line https://github.com/google/automotive-design-compose/blob/d956f8901dff390cf6ee25cbc16467adf08a0649/designcompose/src/main/java/com/android/designcompose/InteractionState.kt#L382

yiqunw700:
> I also noticed some other bug introduced by the protoconv that it should not trigger animation for the button here:
> 
> https://github.com/google/automotive-design-compose/blob/d956f8901dff390cf6ee25cbc16467adf08a0649/designcompose/src/main/java/com/android/designcompose/InteractionState.kt#L404
> 
> transition will never be null now. We should use transitionOrNull in line
> 
> https://github.com/google/automotive-design-compose/blob/d956f8901dff390cf6ee25cbc16467adf08a0649/designcompose/src/main/java/com/android/designcompose/InteractionState.kt#L382

ok. Nevery mind. I don't really know what is going on. When I use transitionOrNull, the VariantsTest in the interaction test is broken...

yiqunw700 (approved):

rylin8:
Not sure what you mean by the VariantsTest in the interaction test being broken, but I think you're right in that we should use `transitionOrNull`. However, I don't think it's causing an issue not doing that at the moment because the default transition returned will translate into an instant snap animation when `asAnimationSpec()` is called. I did test out `transitionOrNull` and everything seemed to work fine, but if we want to fix that I'd do it separately.

```
internal fun Transition.asAnimationSpec(): AnimationSpec<Float> {
    val transitionTypeCase = this.transitionTypeCase
    val easing =
        when (transitionTypeCase) {
            Transition.TransitionTypeCase.SMART_ANIMATE -> smartAnimate.easing
            Transition.TransitionTypeCase.SCROLL_ANIMATE -> scrollAnimate.easing
            Transition.TransitionTypeCase.PUSH -> push.easing
            Transition.TransitionTypeCase.MOVE_IN -> moveIn.easing
            Transition.TransitionTypeCase.MOVE_OUT -> moveOut.easing
            Transition.TransitionTypeCase.DISSOLVE -> dissolve.easing
            Transition.TransitionTypeCase.SLIDE_IN -> slideIn.easing
            Transition.TransitionTypeCase.SLIDE_OUT -> slideOut.easing
            else -> return snap(0)
        }
```

PR #1997, opened 2025-01-15
Quick fix for the UnsupportedOperationException

yiqunw700:
Create a mutable image map in memory and when write to disk, create a copy of the document and replace the images map with the in memory map.

Fixes: #1998

### DIFF
--- a/common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
+++ b/common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
@@ -18,6 +18,7 @@ package com.android.designcompose.common
 
 import com.android.designcompose.definition.DesignComposeDefinition
 import com.android.designcompose.definition.DesignComposeDefinitionHeader
+import com.android.designcompose.definition.copy
 import com.android.designcompose.definition.view.View
 import com.android.designcompose.live_update.ConvertResponse
 import com.android.designcompose.live_update.figma.FigmaDocInfo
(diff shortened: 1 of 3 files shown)


designcompose/src/main/java/com/android/designcompose/DocContent.kt
```diff
@@ -50,11 +50,15 @@ class DocContent(var c: GenericDocContent, previousDoc: DocContent?) {
...
     init {
         for ((imageKey, bytes) in c.document.imagesMap) {
-            if (bytes.isEmpty()) {
-                if (previousDoc != null && previousDoc.images[imageKey] != null) {
+            if (bytes.isEmpty) {
+                if (
+                    previousDoc != null &&
+                        previousDoc.images[imageKey] != null &&
+                        previousDoc.c.imagesMap[imageKey] != null
+                ) {
```
rylin8, line 58: Is it possible for `previousDoc.images[imageKey]` to exist, but `previousDoc.c.imagesMap[imageKey]` to be null?

yiqunw700: I believe it is not possible but I added the check anyway since below has a !! assertion.

common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
```diff
@@ -57,7 +60,17 @@ class GenericDocContent(
...
             val outputStream = ByteArrayOutputStream()
             header.writeDelimitedTo(outputStream)
-            document.writeDelimitedTo(outputStream)
+            // Write cached images map to disk
+            if (imagesMap == document.imagesMap) {
+                document.writeDelimitedTo(outputStream)
+            } else {
+                document
+                    .copy {
+                        images.clear()
```
rylin8, line 69: What confuses me a lot with these protobuf generated classes is that here, you have to access `images` using the getter/setter `images`, but in other scenarios you have to use `imageMap`, which made me think for a long time that there were two different fields.

rylin8 (approved):

common/src/main/java/com/android/designcompose/common/GenericDocContent.kt
yiqunw700: ok. I renamed the mutable map to the inMemoryImagesMap. Hope that helps.

Issue #2002, opened 2025-01-16
Component instances all dispatch interactions


rylin8:
When using a `ReplacementContent` customization and populating this content with component instances, these component instances dispatch interactions when any one of the is triggered. For example if the replacement content is two buttons that are the same component, and this component has a `while pressing` interaction to change to another variant, pressing either button will cause both to change.

rylin8:
I looked into this and this is not a bug. In order to support multiple replacements of the same component that has interactions, the `key` parameter must be unique for each replacement instance. This is demonstrated in our `VariantInteractionsTest.kt` using code that looks like:
```kotlin
    VariantInteractionsTestDoc.MainFrame(
        content =
            ReplacementContent(
                count = 9,
                content = { index ->
                    {
                        when (index) {
                            0 ->
                                VariantInteractionsTestDoc.ButtonVariant1(
                                    type = ItemType.List,
                                    title = "One",
                                    onTap = { println("Tap One") },
                                    key = "One",
                                )
                            1 ->
                                VariantInteractionsTestDoc.ButtonVariant1(
                                    type = ItemType.List,
                                    title = "Two",
                                    onTap = { println("Tap Two") },
                                    key = "Two",
                                )
                            )
                        }
                    }
                }
           )
    )
```

Issue #2001, opened 2025-01-16
Replacement component of a text node does not work


rylin8:
The `@Composable (ComponentReplacementContext) -> Unit` customization does not work on text nodes anymore. Furthermore, the `ComponentReplacementContext` does not have any `TextStyle` information to be used for the text style if we are replacing the node with another text field, which is the usual use case for replacing a text node.

Issue #2000, opened 2025-01-16
Animation Bugs


rylin8:
While testing the Interaction test in Validation, under "Variant Tests":

* Tap on one of the squares and release quickly. The square ends up in the wrong variant. This may be because the press triggers an animation to another variant, and the release happens before the animation is done. The release interaction doesn't work properly and the node ends up in the pressed-down variant.

* If the `transition` in `InteractionState.dispatch()` is changed to: `val transition = action.node.transitionOrNull`, tapping on the squares sometimes results in a translation animation from a different square.

Issue #1717, opened 2024-10-08
DesignOverlay fills max size of root


yiqunw700:
Squoosh or figma design has <img width="201" alt="8FWovdWxwg5LWgo" src="https://github.com/user-attachments/assets/7b38c798-366a-474c-a7a3-a5a2024b4ca6">
while non-squoosh has 
![9GWpq8dGKBaE4jP](https://github.com/user-attachments/assets/f1a91e6c-cf22-4c8d-8675-5732e7b68255)

yiqunw700:
This bug was for the classic renderer. Close as obsolete since it was deleted.

rylin8:
I was thinking that the bug is in squoosh, and that the overlay should fill the max size of root, though it could be argued either way. @iamralpht can you think of other reasons to limit the overlay to within the root node vs the device size?

yiqunw700:
I'm wondering if there is a way to set the overlay size to within the root node vs the device size in figma..
I will reopen it until we figure it out.

PR #2003, opened 2025-01-16
Make Design Switcher functional again

### LINKED ISSUE #1881: Design Switcher is broken in several areas
* The Change button opens up an overlay that is sometimes off to the right and does not have customizations applied (the help text and sample doc ID are both in the same area)
* The overlay in the last bullet point can't be closed when clicking outside (tap to close overlay not working)
* Pressing on the Change button again crashes (probably tries to open the same overlay again which shouldn't be allowed since clicking anything outside the overlay should close it)
* The buttons to change to a different branch don't work.

rylin8:
Fixes #1881. The design switcher has been broken in squoosh due to a number of bugs: https://github.com/google/automotive-design-compose/issues/1717 https://github.com/google/automotive-design-compose/issues/2000 https://github.com/google/automotive-design-compose/issues/2001 https://github.com/google/automotive-design-compose/issues/2002

Fixing all of these will take some time. In an effort to make the design switcher functional again, this change makes some small changes in the Figma design by removing some animations and interactions that are buggy. A few small changes in the design switcher code fix the functionality to switch between other documents or branches.


integration-tests/validation/src/testDebug/roborazzi/RenderAllExamples/SA-Variant-Timelines.png
rylin8, line 1: When I ran roborazzi, this file changed but I'm not sure why. It's not any change that is noticeable.

yiqunw700: This is because the migrate to Mac screenshot change was submitted without the timeline test being enabled. I believe...

yiqunw700 (approved):

PR #2008, opened 2025-01-17
Support shader as text background

yiqunw700:
Fixes: #2007


rylin8:
Do you have an example of this?

yiqunw700:
> Do you have an example of this?

Done. Added one example with color customization.

rylin8 (approved):

PR #2009, opened 2025-01-18
Support replacement component of a text node

### LINKED ISSUE #2001: Replacement component of a text node does not work

rylin8:
Fixes #2001. Check for component replacement customizations on text nodes. When one exists, skip the text rendering path, and don't return early so that the replacement component is rendered. Save the TextStyle of the text being replaced and use it in the ComponentReplacementContext.

This change adds a new test for text replacement with a BasicTextField that inserts some text with keyboard events.


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
```diff
@@ -241,9 +244,20 @@ internal fun squooshComputeTextInfo(
...
-    // The brush and opacity is computed later at rendering time.
+    val customBrush = customizations.getBrush(v.name)
+    val textBrushAndOpacity =
+        v.style.nodeStyle.textColorOrNull?.asBrush(
+            appContext,
+            document,
+            density.density,
+            variableState,
+        )
+    // The brush and opacity is stored here for use with ComponentReplacementContext. They are
```
yiqunw700, line 255: It seems TextStyle can store alpha too. If we set alpha to the text style here, can we avoid computing it later at rendering time?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
```diff
@@ -297,10 +311,16 @@ internal fun squooshComputeTextInfo(
 
     textMeasureCache.put(
         layoutId,
-        TextMeasureCache.Entry(textMeasureData, customizedText, customTextStyle, annotatedText.text),
+        TextMeasureCache.Entry(
+            textMeasureData,
+            customizedText,
+            customTextStyle,
+            annotatedText.text,
+            textStyle,
```
yiqunw700, line 319: Do we need to cache customTextStyle and textStyle built from node at the same time? What about `customTextStyle ?: textStyle`?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/TextReplaceTest.kt
```diff
@@ -0,0 +1,57 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.text.TextStyle
+import com.android.designcompose.ComponentReplacementContext
+import com.android.designcompose.annotation.Design
+import com.android.designcompose.annotation.DesignComponent
+import com.android.designcompose.annotation.DesignDoc
+
+@DesignDoc(id = "yX3mcO0qFnvG6ruUYhHBMR")
```
yiqunw700, line 34: It seems this file `yX3mcO0qFnvG6ruUYhHBMR` is  now under your drafts folder.

yiqunw700:
LGTM overall. Just two quick questions.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
rylin8: I could attempt that, but the text rendering code doesn't touch the cache at this time -- it just gets the text color and alpha from the style, so I guess it's not really a computation. I'll update this comment. Changing the renderer to get the data from the cache would be a slightly bigger change, so I don't think it's worth it at the moment.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
rylin8: The new `textStyle` is needed so that we can return it to `SquooshRoot()` so that it can use it in the `ComponentReplacementContext`. `customTextStyle` is used in `SquooshText()` when it retrieves data from the text measure cache as a check to make sure you have the correct entry. I wouldn't combine them as that would change the caching logic.

yiqunw700 (approved):

PR #2013, opened 2025-01-23
Support float2, float3 and float4 in shader plugin

yiqunw700:
Towards: #1830


support-figma/extended-layout-plugin/src/shader.html
```diff
@@ -292,6 +303,70 @@
...
+        const uniformName = floatArrayUniformNameInput.value;
+        let selectedUniformType = document.querySelector('input[name="floatArrayUniformType"]:checked');
+        createShaderFloatArrayUniform(floatArrayUniformNameInput.value, selectedUniformType.value, null);
+
+        createFloatArrayUniformDialog.close();
+        await runShader();
+    };
+
+    function createShaderFloatArrayUniform(uniformName, uniformType, uniformValue) {
+        if (customUniforms.includes(uniformName)) {
```
rylin8, line 322: Should we also check against the built in values, `iResolution` and `iTime`?

support-figma/extended-layout-plugin/src/shader.html
```diff
@@ -292,6 +303,70 @@
...
+            alert("This uniform already exists, please use a different name.");
+            return;
+        }
+        const regex = /^[a-zA-Z_].*$/;
+        if (uniformName.length == 0 || !uniformName.match(regex)) {
+            alert("Invalid uniform name, please use a different name");
+            return;
+        }
+        customUniforms.push(uniformName);
```
rylin8, line 331: This checking for a valid uniform name is used in the other uniform creation functions, probably good to put in a common function.

support-figma/extended-layout-plugin/src/shader.html
```diff
@@ -292,6 +303,70 @@
...
+
+        const shaderLine = document.createElement("div");
+        shaderLine.classList.add("center-vertically-container");
+        const shaderCode = document.createElement("span");
+        shaderCode.textContent = `uniform ${uniformType} ${uniformName};`;
+        shaderLine.appendChild(shaderCode);
+
+        const deleteButton = createDeleteButton(uniformName, newUniformEntry);
+        shaderLine.appendChild(deleteButton);
+        newUniformEntry.appendChild(shaderLine);
```
rylin8, line 345: This code to create the html is also the same as other functions like `createShaderFloatUniform()`

support-figma/extended-layout-plugin/src/shader.html
yiqunw700: Done.

support-figma/extended-layout-plugin/src/shader.html
yiqunw700: Done.

rylin8 (approved):

PR #2014, opened 2025-01-23
Support shader int uniform

yiqunw700:
Towards: #1830


rylin8 (commented):
Looks good, I just have the same comment as https://github.com/google/automotive-design-compose/pull/2013 about using some shared code between all the functions that create uniform values.

yiqunw700:
> Looks good, I just have the same comment as #2013 about using some shared code between all the functions that create uniform values.

Done. Please take another look. Thanks!

rylin8 (approved):

PR #2018, opened 2025-01-25
Fix overlays to be positioned relative to screen size again

### LINKED ISSUE #1717: DesignOverlay fills max size of root

rylin8:
Fixes #1717: Instead of rendering overlays as a child of the root node, render them after the root node as a composable inside of a full screen box. This allows the overlay to be positioned relative to the size of the entire screen instead of the root node. Use the overlay extras to color the box and make the box clickable if that option is turned on.


designcompose/src/testDebug/kotlin/com/android/designcompose/DesignSwitcherBasicTests.kt
```diff
@@ -105,7 +105,7 @@ class DesignSwitcherBasicTests {
         with(composeTestRule) {
             onDCDoc(DesignSwitcherDoc).performClick()
             with(onDCDoc(DesignSwitcherDoc)) {
-                onNodeWithTag("#ChangeButton").performClick()
+                onNodeWithTag("#ChangeButton", useUnmergedTree = true).performClick()
                 assertHasText("Load")
             }
             captureRootRoboImage("ChangeFileScreen")
```
rylin8, line 111: Hi @timothyfroehlich, it looks like the design switcher tests are passing, so the overlays should be rendering because the overlay is what has the "Load" text in it. However, the screenshot captured doesn't have the overlay shown at all. Any idea why?

yiqunw700 (approved):

PR #2022, opened 2025-01-28
Make shader iTime state a global customization

yiqunw700:
By making the `iTime` state a global customization, newly created nodes will apply the `iTime` customization automatically. This way, it requires no client side change for new node to animate.
Also removed logging when customized uniform is not found in the shader code. Too many log outputs when it is animating.

Before: create a new node with animating shader, no animation after live update because iTime customization is not set
After: create a new node with animating shader, animates after live update

* Also update Dials and Gauges with usage of shader


rylin8:
Do you know why the progress-constraints* screenshots changed?

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/DialsGauges.kt
```diff
@@ -36,13 +37,18 @@ import org.robolectric.annotation.GraphicsMode
 
 @RunWith(AndroidJUnit4::class)
 @GraphicsMode(GraphicsMode.Mode.NATIVE)
-@Config(qualifiers = "w1920dp-h1500dp-xlarge-long-notround-any-xhdpi-keyshidden-nonav")
+@Config(qualifiers = "w1920dp-h1500dp-xlarge-long-notround-any-xhdpi-keyshidden-nonav", sdk = [35])
```
rylin8, line 40: Is 35 needed for something?

yiqunw700:
> Do you know why the progress-constraints* screenshots changed?

Yes. I enabled hardware rendering for Dials and Gauges tests now. 35 is needed for shader to render correctly with Robolectric.

rylin8 (approved):

integration-tests/validation/src/testDebug/kotlin/com/android/designcompose/testapp/validation/DialsGauges.kt
yiqunw700: Yes. When testing with sdk 34, 33, Robolectric only renders the shader as black. It only renders correctly on api level 35.

Issue #1554, opened 2024-08-22
Hyperlink support in text for Squoosh


yiqunw700:
#1449 has the default render done.

PR #2033, opened 2025-02-05
Support hyperlink in Squoosh

yiqunw700:
Fixes:  #1554

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
+++ b/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
@@ -22,6 +22,7 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.Shadow
 import androidx.compose.ui.graphics.SolidColor
 import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.LinkAnnotation
 import androidx.compose.ui.text.ParagraphIntrinsics
 import androidx.compose.ui.text.PlatformTextStyle
 import androidx.compose.ui.text.SpanStyle
@@ -136,6 +137,8 @@ internal fun squooshComputeTextInfo(
         return Pair(cachedText.data, cachedText.textStyle)
     }
 
+    val hyperlinkOffsetMap = LinkedHashMap<Int, String?>()
+
     val annotatedText =
         if (customizedText != null) {
             val builder = AnnotatedString.Builder()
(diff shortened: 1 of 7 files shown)


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
```diff
@@ -193,6 +203,17 @@ internal fun squooshComputeTextInfo(
...
                     )
+                    if (run.style.hasHyperlink()) {
+                        val link = run.style.hyperlink.value
+                        builder.addLink(
+                            LinkAnnotation.Url(link),
+                            startIndex,
+                            startIndex + run.text.length,
+                        )
+                        hyperlinkOffsetMap[startIndex] = link
+                        hyperlinkOffsetMap[startIndex + run.text.length] = null
```
rylin8, line 214: Since startIndex is always increasing, is there a reason you didn't use something like an ArrayList instead of a HashMap? Then you wouldn't need to sort the keys... although I don't think think this would matter for any practical usage of hyperlinks in text since you would need to have a lot of hyperlinks before the sorting got slow.

rylin8 (commented):
Nice, this is the last missing feature from squoosh! Except for a few bugs...

integration-tests/validation/src/androidTest/java/com/android/designcompose/testapp/validation/HyperlinkTests.kt
```diff
@@ -37,16 +48,19 @@ class HyperlinkTests {
     @get:Rule val interFontRule = InterFontTestRule()
     @get:Rule var intentsRule: IntentsRule = IntentsRule()
 
-    // These are disabled because squoosh does not support clickable hyperlink text
-    /*
     @Test
     fun testClickSingleNodeLink() {
+        val result = Instrumentation.ActivityResult(Activity.RESULT_OK, null)
         val url = "https://github.com/google/automotive-design-compose"
+        intending(hasData(url)).respondWith(result)
```
rylin8, line 55: can you add a couple comments for what `intending()` and `intended()` do in these tests?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
yiqunw700: Done. Changed to LinkedHashMap to no need to sort later.

rylin8 (approved):

PR #2037, opened 2025-02-06
Support int vec uniform

yiqunw700:
Fixes: #2038


designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -749,6 +749,29 @@ fun ShaderUniform.applyToShader(
             }
         }
 
+        ValueTypeCase.INT_VEC_VALUE -> {
+            val intVecValue = value.intVecValue.intsList
+            when (intVecValue.size) {
+                1 -> if (definedType == "int") shader.setIntUniform(name, intVecValue[0])
```
rylin8, line 755: Can this ever happen since the plugin only lets you make int2 int3 or int4?

yiqunw700: oh. It can happen in the uniform customization.

rylin8 (approved):

PR #2045, opened 2025-02-11
Support matrix uniforms

yiqunw700:
Fixes: #2017


integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/BrushFromShaderPluginTest.kt
```diff
@@ -82,11 +85,26 @@ fun BrushFromShaderPluginTest() {
     }
     val customIntStates = ShaderUniformCustomizations()
     customIntStates.customBackgroundShaderUniformStates(intState.toShaderUniformState("iCase"))
+
+    val angleState = remember {
+        derivedStateOf {
+            val angle = iTimeFloatState.floatValue % 360.toDouble()
```
rylin8, line 91: Not a huge deal, but I think cos/sin take radians, so this when this number goes over 360, the rotation will jump. Changing 360 to (2 * PI) would fix it, I think.

rylin8 (approved):

PR #2047, opened 2025-02-12
Optimize by caching shader brushes

yiqunw700:


designcompose/src/main/java/com/android/designcompose/ShaderBrushCache.kt
```diff
@@ -0,0 +1,56 @@
...
+import com.android.designcompose.definition.element.ShaderData
+
+class ShaderBrushCache {
+    data class Entry(var shaderBrush: Brush, var shaderData: ShaderData)
+
+    private val cache = LruCache<Int, Entry>(64)
+
+    fun get(layoutId: Int?, shaderData: ShaderData?): Brush? {
+        if (layoutId == null || shaderData == null) return null
+        return if (cache[layoutId]?.shaderData == shaderData) {
```
rylin8, line 36: Is there a way to avoid this comparison? I'm thinking that `shaderData` incudes all the code and uniforms of the shader program, which could potentially be pretty big.

rylin8 (approved):

Issue #2043, opened 2025-02-07
Performance of using shaders


yiqunw700:
Needs further evaluation since people are concerned about the performance of using shaders.

yiqunw700:
Low end device can override the `config_shader_frame_interval_ms` to decrease the draw frequency. Also medium precision is now supported.

PR #2048, opened 2025-02-12
Change how iTime is simulated

yiqunw700:
Configurable frame rate using a resource integer to define the delay between frames. Also check the global animator scale settings to see if animation is enabled.

Also after the change, iTime starts from 0.

Towards: #2043

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/ShaderHelper.kt
+++ b/designcompose/src/main/java/com/android/designcompose/ShaderHelper.kt
@@ -16,9 +16,14 @@
 
 package com.android.designcompose
 
+import android.content.Context
+import android.database.ContentObserver
 import android.os.Build
+import android.provider.Settings
+import androidx.compose.animation.core.InfiniteTransition
 import androidx.compose.animation.core.withInfiniteAnimationFrameMillis
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.FloatState
 import androidx.compose.runtime.IntState
 import androidx.compose.runtime.State
@@ -27,11 +32,13 @@ import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.mutableFloatStateOf
 import androidx.compose.runtime.produceState
 import androidx.compose.runtime.remember
+import androidx.compose.ui.platform.LocalContext
 import com.android.designcompose.definition.element.ShaderUniform
 import com.android.designcompose.definition.element.ShaderUniformValueKt.floatVec
 import com.android.designcompose.definition.element.floatColor
 import com.android.designcompose.definition.element.shaderUniform
 import com.android.designcompose.definition.element.shaderUniformValue
+import kotlinx.coroutines.delay
 
 data class ShaderUniformCustomizations(
     val backgroundShaderUniforms: MutableList<ShaderUniform> = mutableListOf(),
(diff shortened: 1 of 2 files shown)


designcompose/src/main/java/com/android/designcompose/ShaderHelper.kt
```diff
@@ -81,10 +85,20 @@ object ShaderHelper {
...
                     while (true) {
-                        withInfiniteAnimationFrameMillis { value = it / 1000f }
+                        val currentTime = withInfiniteAnimationFrameMillis { it }
+                        if (isMotionEnabled(context)) {
+                            value = (currentTime - startTime) / 1000f
+                        }
+                        delay(
+                            context.resources
+                                .getInteger(R.integer.config_shader_frame_interval)
+                                .toLong()
```
rylin8, line 100: Can a 3rd party who uses DesignCompose change this frame interval value? I would think they should be able to. Right now we have a few minor settings that they can set in the global `DesignSettings` class. Would it make sense to set it there instead?

designcompose/src/main/java/com/android/designcompose/ShaderHelper.kt
```diff
@@ -141,4 +155,16 @@ object ShaderHelper {
     fun IntState.toShaderUniformState(name: String): State<ShaderUniform> {
         return remember { derivedStateOf { createShaderIntUniform(name, intValue) } }
     }
+
+    private fun isMotionEnabled(context: Context): Boolean {
+        return try {
+            Settings.Global.getFloat(
+                context.contentResolver,
+                Settings.Global.ANIMATOR_DURATION_SCALE,
```
rylin8, line 163: As I had no idea what this setting was, we should make sure to document it when we add the documentation for the shader plugin.

designcompose/src/main/java/com/android/designcompose/ShaderHelper.kt
yiqunw700: Yes. They can. Android app can override android library resources. OEMs can also overlay this resource to set different interval on different devices using rro while using the same apk. So I would prefer to use resources here.

It seems a little bit hard to have DesignSettings to initialize from resources at the moment. Otherwise I will try to do that.

designcompose/src/main/java/com/android/designcompose/ShaderHelper.kt
yiqunw700: Sure. I also added some comment here.

designcompose/src/main/java/com/android/designcompose/ShaderHelper.kt
```diff
@@ -82,10 +87,18 @@ object ShaderHelper {
...
+        val context = LocalContext.current
+        val frameInterval =
+            context.resources.getInteger(R.integer.config_shader_frame_interval_ms).toLong()
         return if ("robolectric" != Build.FINGERPRINT)
             produceState(0f) {
+                    val startTime = withInfiniteAnimationFrameMillis { it }
                     while (true) {
-                        withInfiniteAnimationFrameMillis { value = it / 1000f }
+                        val currentTime = withInfiniteAnimationFrameMillis { it }
+                        if (isMotionEnabled(context)) {
```
rylin8, line 98: Should this check be pulled out to the beginning of the function, so that if it returns false, you can just return something like `floatStateOf(0f)`?

yiqunw700: To make it more efficient, I added the settings change event listener to update the animation enabled state instead of keeping reading the content resolver.

It works well when I have the check before setting the value. When I pull it out, I don't know why the shader will stop animation when the setting changes to turning off the  device animation but it will not start animating again when I re-enable it.

designcompose/src/main/java/com/android/designcompose/ShaderHelper.kt
```diff
@@ -82,14 +91,41 @@ object ShaderHelper {
...
+        return if ("robolectric" != Build.FINGERPRINT) {
             produceState(0f) {
+                    val startTime = withInfiniteAnimationFrameMillis { it }
                     while (true) {
-                        withInfiniteAnimationFrameMillis { value = it / 1000f }
+                        val currentTime = withInfiniteAnimationFrameMillis { it }
+                        if (isAnimationEnabled.value) {
+                            value = (currentTime - startTime) / 1000f
+                        }
+                        delay(frameInterval)
```
rylin8, line 124: Can you pass in `isAnimationEnabled.value` into `produceState()`:
```
            produceState(0f, isAnimationEnabled.value) {
                if (isAnimationEnabled.value) {
                    val startTime = withInfiniteAnimationFrameMillis { it }
                    while (true) {
                        val currentTime = withInfiniteAnimationFrameMillis { it }
                        value = (currentTime - startTime) / 1000f
                        delay(frameInterval)
                    }
                }
```

rylin8:
Looks good. Let's make sure to make a note of this in the documentation, because only the shaders use the animation duration setting. We should think about also supporting it for our other animations.

rylin8 (approved):

Issue #1973, opened 2025-01-13
Component animate interpolating shader uniform value


yiqunw700:
Will add more details later

PR #2071, opened 2025-02-27
Support smart shader animation when changing variants

yiqunw700:
Fixes: #1973


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -653,6 +847,39 @@ private fun needsStyleTween(a: ViewStyle, b: ViewStyle): Boolean {
...
+        return false
+    }
+
+    // Only tweenable when they are not animating shaders
+    if (a.shaderUniformsMap.containsKey(ShaderHelper.UNIFORM_TIME)) return false
+    // Only tweenable when they have the same uniforms
+    if (a.shaderUniformsMap.keys != b.shaderUniformsMap.keys) return false
+    // Only tweenable when they have the same shader. The shader code might be a long string,
+    // we do a hashcode and real string compare for improved efficiency.
+    if (a.shader.hashCode() != b.shader.hashCode() && a.shader != b.shader) return false
```
rylin8, line 865: Is the shader comparison needed or does `hashCode()` take care of that?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -653,6 +847,39 @@ private fun needsStyleTween(a: ViewStyle, b: ViewStyle): Boolean {
...
+    a.shaderUniformsMap
+        .filterValues {
+            it.value.valueTypeCase in
+                arrayOf(
+                    ValueTypeCase.FLOAT_VALUE,
+                    ValueTypeCase.FLOAT_VEC_VALUE,
+                    ValueTypeCase.FLOAT_COLOR_VALUE,
```
rylin8, line 872: What about integers?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -653,6 +847,39 @@ private fun needsStyleTween(a: ViewStyle, b: ViewStyle): Boolean {
...
+                )
+        }
+        .forEach { (key, value) ->
+            if (b.shaderUniformsMap[key]?.value != value.value) {
```
rylin8, line 876: What if `a` and `b` both have the same uniform name, but with different types?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -576,11 +744,25 @@ private fun mergeRecursive(
         // XXX: Refactor this so we don't inspect every type right here.
 
         val fromArc = from.view.data.containerOrNull?.shapeOrNull?.arcOrNull
-        val toArc = from.view.data.containerOrNull?.shapeOrNull?.arcOrNull
+        val toArc = to.view.data.containerOrNull?.shapeOrNull?.arcOrNull
```
rylin8, line 747: Good catch. I don't know if we have any arc animations in our tests...

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -301,6 +314,161 @@ internal class SquooshAnimatedArc(
...
+                                        )
+                                }
+                            },
+                        )
+                    ValueTypeCase.FLOAT_VEC_VALUE ->
+                        if (
+                            fromUniformValue?.floatVecValue?.floatsCount !=
+                                toUniformValue?.floatVecValue?.floatsCount
+                        ) {
+                            Log.e(TAG, "Mismatch from $fromUniformValue to $toUniformValue")
```
rylin8, line 370: can we catch this case in the `isTweenable()` function?

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -920,59 +962,62 @@ internal fun getShaderBrush(
...
+                customShaderUniforms
+                    ?.get(k)
+                    ?.applyToShader(
+                        shaderBrush as SizingShaderBrush,
+                        shaderData.shaderUniformsMap,
+                        document,
+                        appContext,
+                        density,
+                    )
+            if (customized == true) {
```
rylin8, line 982: Did you change `applyToShader()` to return a boolean in order to fix a problem, or is this to improve efficiency?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: I should use `||` instead of `&&`...

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: The shader tweenable state checks if the shader code are the same which includes the uniform declaration. It means uniforms are the same types.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
yiqunw700: To improve efficiency. I made this change to only apply the shaderUniform once each redraw.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
```diff
@@ -653,6 +847,42 @@ private fun needsStyleTween(a: ViewStyle, b: ViewStyle): Boolean {
...
+    }
+
+    // Only tweenable when they are not animating shaders
+    if (a.shaderUniformsMap.containsKey(ShaderHelper.UNIFORM_TIME)) return false
+    // Only tweenable when they have the same uniforms
+    if (a.shaderUniformsMap.keys != b.shaderUniformsMap.keys) return false
+    // Only tweenable when they have the same shader. The shader code might be a long string,
+    // we do a hashcode and real string compare for improved efficiency.
+    if (a.shader.hashCode() != b.shader.hashCode()) return false
+    // When hashes are the same, the shader string might still be different.
```
rylin8, line 866: When would the hashcode be the same but the strings different? Isn't the point of hashcode to be a unique integer based on the contents of the string? Are you just referring to the rare scenario where two strings can map to the same hashcode?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: Not going to support integers at this point:
1. It is hard to produce the intermediate integers due to the value generated is floats in [0, 1]. 
2. The produced intermediate integers might not be something the code expected

For example, we have the integer as a case number. It probably won't produce any valid animation from case `1` to case `2`.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: Done.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: Yeah. Do you think it worth it to add the a != b check for that rare case? I am ok to remove this check due to the rareness....

rylin8: I don't think it's needed, the chances are really small. If in the future we do want to be more careful, then I would consider doing md5 or sha256 on the string and using that.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: There seems to be one https://www.figma.com/design/RW3lFurXCoVDeqY2Y7bf4v/Animated-Transition-Test?node-id=30-33&t=bVNAgUQde0uDHUw6-0 but it is not in the main frame or the AnimationMidpoints tests. Filed #2081. I can add that in a separate pr

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshAnimate.kt
yiqunw700: Sounds good. Removed it. :)

rylin8 (approved):

PR #2086, opened 2025-03-05
Support scrolling when dragging nodes with press/tap interactions

rylin8:
Use the awaitEachGesture function of PointerInputScope to manually handle touch events. This allows the touch gestures to propagate to the parent, but a valid click will consume the event so as not to also click the parent.

This change also fixes some scrolling bugs.
- Fixes an issue where the amount scrolled did not take into account density. T
- Fixes an issue where the children of a scrollable frame were being composed as children of the root node instead of the scrollable frame.
- Fixes an issue where nodes with press/tap interactions were rendered scrolled, but their tappable areas were not, so tapping would dispatch interactions on the wrong node.


rylin8:
I fixed the screenshot test issue with VariantOnPress-anim4.png by moving that part of the test into a new test to start from a fresh state. Since everything works when testing manually, I believe there is an edge case scenario bug with roborazzi that I was running into.

yiqunw700:
Looks good overall. But I also noticed two issues:
1. Layout replacement test is now somehow broken
2. Scrolling can now over-scroll. It is supposed to stop scrolling when the last item is at the right/bottom but it now is still able to continue to scroll. Is it possible somewhere the content height/width is not computed correctly?

yiqunw700:
> Looks good overall. But I also noticed two issues:
> 
> 1. Layout replacement test is now somehow broken
> 2. Scrolling can now over-scroll. It is supposed to stop scrolling when the last item is at the right/bottom but it now is still able to continue to scroll. Is it possible somewhere the content height/width is not computed correctly?

I noticed that when I scroll, the list is scrolling faster than the drag...

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -1194,8 +1200,12 @@ private fun MeasureScope.squooshLayout(
                 val offsetFromRoot = node.offsetFromAncestor()
 
                 placeable.placeRelative(
-                    x = (offsetFromRoot.x * density).roundToInt(),
-                    y = (offsetFromRoot.y * density).roundToInt(),
+                    x =
```
yiqunw700, line 1203: Reverting changes here will fix the issue I noticed. I am wondering if the scrollable modifier already handles the offset...

rylin8: Yea I tested reverting this also. The problems is then that the scroll test with tappable elements does not work properly. The child nodes with click interactions don't scroll, so clicking on them when they've scrolled will trigger an interaction on a different node. So I need to figure out how to scroll the interactive child components in a scrollview without breaking other scrolling.

rylin8: Ok, problem was that `SquooshLayout()` was being called twice for replacement content, since each node of the replacement content is a child composable and goes through the `SquooshLayout()` code. That caused those items to get scrolled double the amount that they were supposed to be. I've fixed the issue, though it's not very clean.

yiqunw700: Cool. Looks good to me. Feel free to file a ticket if you would like to look into a cleaner solution.

yiqunw700 (approved):

PR #2088, opened 2025-03-05
Migrate Rust protobuf code generation from prost to protobuf

### LINKED ISSUE #2089: Migrate Rust proto code from prost to libprotobuf

rylin8:
Fixes #2089


crates/figma_import/src/document.rs
```diff
@@ -80,7 +81,13 @@ fn get_branches(document_root: &figma_schema::FileResponse) -> Vec<FigmaDocInfo>
     if let Some(doc_branches) = &document_root.branches {
         for hash in doc_branches {
             if let (Some(Some(id)), Some(Some(name))) = (hash.get("key"), hash.get("name")) {
-                branches.push(FigmaDocInfo::new(name.clone(), id.clone(), String::new()));
+                let figma_doc = FigmaDocInfo {
+                    name: name.clone(),
+                    id: id.clone(),
+                    version_id: String::new(),
+                    special_fields: Default::default(),
```
rylin8, line 88: I've been doing ..Default::default() as the last parameter since that's what the other code normally does.

yiqunw700: Looks like it works.

crates/figma_import/src/tools/fetch.rs
```diff
@@ -18,9 +18,10 @@ use std::io::{Error, ErrorKind};
 use crate::{proxy_config::ProxyConfig, Document};
 /// Utility program to fetch a doc and serialize it to file
 use clap::Parser;
-use dc_bundle::definition::DesignComposeDefinitionHeader;
-use dc_bundle::definition::{DesignComposeDefinition, NodeQuery};
+use dc_bundle::definition::NodeQuery;
 use dc_bundle::definition_file::save_design_def;
+use dc_bundle::design_compose_definition::DesignComposeDefinition;
```
yiqunw700, line 23: This is the only fix in this commit. Others are just formatting.

yiqunw700:
Error with running `cargo test`
---- reaction_schema::parse_reactions stdout ----

thread 'reaction_schema::parse_reactions' panicked at crates/figma_import/src/reaction_schema.rs:664:44:
called `Result::unwrap()` on an `Err` value: Error("unknown variant `NAVIGATE`, expected one of `NAVIGATION_UNSPECIFIED`, `NAVIGATION_NAVIGATE`, `NAVIGATION_SWAP`, `NAVIGATION_OVERLAY`, `NAVIGATION_SCROLL_TO`, `NAVIGATION_CHANGE_TO`", line: 1, column: 208)

---- design_definition::serialized_document_tests::load_save_load stdout ----

thread 'design_definition::serialized_document_tests::load_save_load' panicked at crates/figma_import/src/design_definition.rs:50:55:
Failed to load design bundle.: DCDLoadError("Failed to parse DCD: Unexpected wire type")

crates/dc_bundle/src/definition/layout.rs
```diff
@@ -36,8 +36,8 @@ impl LayoutStyle {
 
     pub(crate) fn new_default() -> LayoutStyle {
         LayoutStyle {
-            margin: Some(DimensionRect::new()).into(),
-            padding: Some(DimensionRect::new()).into(),
+            margin: Some(DimensionRect::new_with_default_value()).into(),
```
rylin8, line 39: What does `DimensionRect::new()` do?

yiqunw700: top, bottom, start, end are MessageField with None. I checked the impl before, it used to be unspecified. But I haven't double checked if it will work using `::new()`.

yiqunw700:
with #2095 it seems it works on kotlin side. The error left is the one in reaction_schema test

crates/dc_bundle/src/definition/layout.rs
yiqunw700: double checked. Using new() will throw `Protobuf ConversionError: Missing field DimensionProto` because try_info_taffy will throw it when it is None.

deny.toml
```diff
@@ -49,7 +49,9 @@ targets = [
 # they are connected to another crate in the graph that hasn't been pruned,
 # so it should be used with care. The identifiers are [Package ID Specifications]
 # (https://doc.rust-lang.org/cargo/reference/pkgid-spec.html)
-#exclude = []
+exclude = [
+    "protobuf",
```
yiqunw700, line 53: Wondering if we should add RUSTSEC-2024-0437 to ignore at line 85??

rylin8: Oh yea, that would be better than just allowing protobuf, let me try that.

Issue #1842, opened 2024-12-06
Component Instances Show Incorrect Text in Pressed Variant


Elangovan4ever:
### Description
When reusing components to create multiple instances with different texts, the "Pressed" variant of the component displays the default text from the main component instead of the text assigned to the specific instance.

### Steps to Reproduce
1. Create a component with two variants: Enabled and Pressed.
2. Set up a Figma prototype to transition from Enabled to Pressed on press.
3. Assign different text to each instance of the component.
4. In Design Compose app, press any instance of the component in the prototype.

### Expected Behavior
Each instance should display its assigned text when transitioning to the "Pressed" variant.

### Actual Behavior
The instance displays the default text ("Label") from the main component instead of the text assigned to it.

### Example Setup: 
**Refer attached project and Figma file from that project**
Component: content-type-list-item with two variants (Enabled and Pressed).
Instances: "Audio", "Image", "Video"
Issue: Pressing any instance shows "Label" instead of the assigned text.

### Workaround
Currently, to resolve the issue, separate components need to be created for each text variant:

**Example for workaround**:
Refer genre-type-list which has distinct components for each item.

### Note
The list is static and designed in Figma.
Text is not supplied programmatically.

### Attachments
[StaticListTest.zip](https://github.com/user-attachments/files/18033160/StaticListTest.zip)

rylin8:
When a component instance overrides some content from the original component, we currently only support style overrides. This includes some properties such as colors and strokes, but not others like corner radii and child content. This has been a long outstanding task but we haven't had much need for it.

Code comment referring to this is in `document.rs`:
```
                // Currently we are only computing the override style for the root of a component.
                // We should expand this to traverse the tree, matching up nodes from each side
                // and determining deltas.
```

PR #2090, opened 2025-03-06
Iterate tree for component overrides

yiqunw700:
The component_info now has one more field:
-  map<string, ComponentOverrides> overrides_table = 4; which stores the computed overrides.

This pr has a limited support for the component overrides:
* The override_table holds a name to override data map. If name is not identical, it will have unexpected behaviors. 
* Only check if there is any character overrides, when text is overridden, use the text or styledText from the overrides_table stored in the component_info.
* Nested component instance override does not work.

Fixes: #1842


crates/figma_import/src/document.rs
```diff
@@ -463,65 +464,180 @@ impl Document {
...
+                        if view_style == component.unwrap().style {
+                            MessageField::none()
+                        } else {
+                            let diff: Option<ViewStyle> = Some(
+                                component.unwrap().style().difference(view_style.as_ref().unwrap()),
+                            );
+                            diff.into()
+                        }
+                    } else {
+                        // let view_name = overrides_table_key.clone();
```
yiqunw700, line 599: When doing this, design switcher has some diffs with the screenshots. I would prefer to look into that in a following ticket. #2113

crates/figma_import/src/document.rs
```diff
@@ -463,65 +464,180 @@ impl Document {
...
+                        {
+                            for child in children {
+                                for_each_component_instance(
+                                    reference_components,
+                                    child,
+                                    None,
+                                    None,
+                                    action,
+                                );
+                            }
```
yiqunw700, line 559: Here is some duplicate code with above. I could not figure out how to move them into a common function...

rylin8 (commented):
Took a first pass at reviewing this, I'm going to test it out a bit.

rylin8 (commented):
Clicking the top right rect when it is blue doesn't work, and clicking the bottom right rect doesn't ever work. Is that intended?

crates/figma_import/src/document.rs
```diff
@@ -463,65 +464,180 @@ impl Document {
...
+                        if overrides.fields.contains(&text_key.to_string()) {
+                            view_data
+                        } else {
+                            MessageField::none()
+                        }
+                    } else {
+                        MessageField::none()
+                    };
+
+                    component_info.overrides_table.insert(
```
rylin8, line 631: Instead of inserting `override_view_data` into this table, why can't you just directly set `component_info.override_view_data` to it?

rylin8: Oh is it because the child nodes of a component instance don't have a `component_info`?

yiqunw700: Yes. Child nodes don't have a `component_info`.

yiqunw700:
> Clicking the top right rect when it is blue doesn't work, and clicking the bottom right rect doesn't ever work. Is that intended?

This change actually shouldn't change the style behaviors. I believe there are existing bugs that use the view's style instead of resolved node style for the node size and also the text measurement. It actually works if click on the left center.

I will take a look.

The bottom right rect is because of the background color override.

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -943,9 +951,11 @@ fn visit_node(
                     .unwrap_or(&figma_schema::ComponentSet::default())
                     .name
                     .clone(),
-                overrides: None.into(),
+                overridden_fields_by_id: overridden_fields_by_id.unwrap_or(HashMap::new()),
+                overrides_table: HashMap::new(),
                 ..Default::default()
             });
+            error!("component_info for instance: {:?}", component_info);
```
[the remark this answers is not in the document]
yiqunw700: Done. Removed.

crates/figma_import/src/document.rs
```diff
@@ -463,65 +464,180 @@ impl Document {
...
+                    let queries = [
+                        NodeQuery::NodeId(info.id.clone()),
+                        NodeQuery::NodeName(info.name.clone()),
+                        NodeQuery::NodeVariant(info.name.clone(), info.component_set_name.clone()),
+                    ];
+
+                    let reference_component_option =
+                        queries.iter().find_map(|query| reference_components.get(query));
+
+                    if let Some(_) = reference_component_option {
```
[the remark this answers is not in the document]
yiqunw700: Done.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshTreeBuilder.kt
```diff
@@ -210,23 +211,33 @@ internal fun resolveVariantsRecursively(
...
     var view = viewFromTree
 
     // If we have a component then we might need to get an override style, and we definitely
     // need to get a different layout id.
     if (viewFromTree.hasComponentInfo()) {
         parentComps =
             ParentComponentData(parentComponents, viewFromTree.id, viewFromTree.componentInfo)
+        System.err.println(viewFromTree.componentInfo.overridesTableMap)
+        System.err.println(viewFromTree.name)
+        System.err.println(viewFromTree.componentInfo.componentSetName)
```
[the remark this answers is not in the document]
yiqunw700: Done. Removed.

crates/figma_import/src/document.rs
```diff
@@ -463,65 +464,180 @@ impl Document {
...
+                        // We should expand this to traverse the tree, matching up nodes from each side
+                        // and determining deltas.
+                        //
+                        // This is the same algorithm we'll have to make to do Smart Animate (including
+                        // determining delta styles).
+                        if view_style == component.unwrap().style {
+                            MessageField::none()
+                        } else {
+                            let diff: Option<ViewStyle> = Some(
+                                component.unwrap().style().difference(view_style.as_ref().unwrap()),
```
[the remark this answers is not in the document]
yiqunw700: Done. Please take another look.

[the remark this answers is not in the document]
rylin8: oh, sorry I meant do something like, `if let Some(a) = b` instead of `b.unwrap()` or `b.expect()`. `expect()` is the same as `unwrap()` in that it will crash the application.

yiqunw700:
> > Clicking the top right rect when it is blue doesn't work, and clicking the bottom right rect doesn't ever work. Is that intended?
> 
> This change actually shouldn't change the style behaviors. I believe there are existing bugs maybe somewhere measurement is not quite right or maybe somewhere squoosh interaction detection is not quite right. It actually works if click on the left center.
> 
> I will take a look.
> 
> The bottom right rect is because of the background color override.

Figured out. The difference function in view.rs for ViewStyle doesn't count the bounding box.

crates/figma_import/src/document.rs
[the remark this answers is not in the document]
yiqunw700: Done.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshTreeBuilder.kt
```diff
@@ -210,23 +211,33 @@ internal fun resolveVariantsRecursively(
...
+
+        overrideStyle =
+            viewFromTree.componentInfo.overridesTableMap[
+                    viewFromTree.componentInfo.componentSetName]
+                ?.styleOrNull
+        // overrideStyle = viewFromTree.componentOverridesOrNull?.styleOrNull
+        overrideViewData =
+            viewFromTree.componentInfo.overridesTableMap[
+                    viewFromTree.componentInfo.componentSetName]
+                ?.viewDataOrNull
```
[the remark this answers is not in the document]
yiqunw700: I will keep this for later when non-text nodes are supported. Cause it is possible to hide some child nodes in the instance.
(Also wondering if it is impossible to have a text node as the root node of the component instance)???

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshTreeBuilder.kt
```diff
@@ -171,6 +172,10 @@ internal class ParentComponentData(
 
         return true
     }
+
+    fun getRootComponentInfo(): ComponentInfo {
+        return parent?.getRootComponentInfo() ?: componentInfo
+    }
```
rylin8, line 178: Is this used?

crates/dc_bundle/src/definition/view.rs
```diff
@@ -460,4 +479,30 @@ impl View {
...
         self.style.as_mut().expect("ViewStyle is required.")
     }
+
+    /** This function is now only called by a view that is a COMPONENT. */
+    pub fn find_view_by_id(&self, view_id: String) -> Option<&View> {
+        if self.id == view_id {
+            return Some(&self);
+        } else if let Some(id) = view_id.split(";").last() {
+            // If this is a descendent node of an instance, the last section is the node id
+            // of the view in the component. Example: I70:17;29:15
```
rylin8, line 489: This seems to rely on the way node IDs are setup in Figma. Is there some way to do this without relying on Figma's mechanism for assigning layout IDs?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/ComponentOverridesTest.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.android.designcompose.annotation.DesignDoc
+
+@DesignDoc(id = "0yGWgsnmpZk1yVVcGnoEKN")
+interface ComponentOverridesTest {
+    @DesignComponent(node = "#MainFrame") fun Main()
+}
+
+@Composable
+fun ComponentOverridesTest() {
+    ComponentOverridesTestDoc.Main()
```
rylin8, line 30: I'm not sure what the behavior is supposed to be for this test. I noticed when I click on some of them, the text changes, but not all of them. Shouldn't the "OK" button at the bottom stay as "OK" when I press on it? Can you add some roborazzi tests that click on the buttons and verify that the text changes or stays the same appropriately, and takes some screenshots too?

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshTreeBuilder.kt
yiqunw700: Removed. Was testing some of the nested component thing...

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/ComponentOverridesTest.kt
yiqunw700: Added some text info. There are two tests not working now. Added screenshot tests.

crates/dc_bundle/src/definition/view.rs
yiqunw700: I actually think it is a reliable way to find the matched view from the component. Do you have concerns that figma will change this in the future? 
If so, another option I can think of is to iterate the tree. But I will prefer to look into that in the next ticket together with the nested component instances.

crates/figma_import/src/document.rs
```diff
@@ -463,65 +462,162 @@ impl Document {
...
-                    reference_components.get(&NodeQuery::NodeName(info.name.clone()))
-                {
-                    action(view, reference_component);
-                } else if let Some(reference_component) = reference_components.get(
-                    &NodeQuery::NodeVariant(info.name.clone(), info.component_set_name.clone()),
-                ) {
-                    action(view, reference_component);
-                }
-            }
+            match (view.component_info.as_mut(), parent_component_info) {
```
rylin8, line 477: Can you add a comment to each branch of this match to explain the scenario (this is a component, this is the child of a component, this is some child of the root that hasn't encountered a component)

crates/dc_bundle/src/definition/view.rs
rylin8: I meant that the data in `View` shouldn't be tied to Figma. The data we get in Figma is parsed in `figma_schema`, but then saved into our own format as `View` and `ViewStyle`. The idea is that this format can be transformed from any design format, not just Figma. However, it does seem that the data in `View` is a bit tied to Figma due to the way Figma uses components, component sets, component instances, etc. I guess if we end up supporting another design tool there will be a bit of work to make `View` and `ViewStyle` more generic.

crates/figma_import/src/document.rs
yiqunw700: Done.

crates/dc_bundle/src/definition/view.rs
yiqunw700: Yeah. That is true that some of the data is tied to Figma like how the variants are named. I will leave it as it is in this pr. Will try to see if I can make it generic next.

rylin8:
Ok changes look good. Since we have now officially started using protobufs in 0.34.0, is your change to the component override protos going to break backward compatibility? `ComponentInfo` is the samek, except that it has moved, so I'm guessing that will be fine, but `ComponentOverrides` has changed. @timothyfroehlich any thoughts on this?

yiqunw700:
> Ok changes look good. Since we have now officially started using protobufs in 0.34.0, is your change to the component override protos going to break backward compatibility? `ComponentInfo` is the samek, except that it has moved, so I'm guessing that will be fine, but `ComponentOverrides` has changed. @timothyfroehlich any thoughts on this?

Good point. For compatibility, assigned new numbers to reserve the old ones instead of replacing them. So the old file can be still loaded but the override style won't work.

rylin8:
I've add https://github.com/google/automotive-design-compose/issues/2126 to test loading of old dcf files. For this PR, is it worth keeping the old component override data to maintain backward compatibility? Let's discuss this a bit before merging.

yiqunw700:
> I've add #2126 to test loading of old dcf files. For this PR, is it worth keeping the old component override data to maintain backward compatibility? Let's discuss this a bit before merging.

For ComponentOverrides, the removed fields weren't used anywhere so I believe it is ok to remove them.
For the ComponentInfo, I added the removed `overrides = 4` back but with [deprecated = true] assuming that DisplaySafety or DriverUi might be still using it and it is probably better to do a deprecation and migration there.

Let's discuss this at tomorrow's team meeting.

rylin8:
I'm ok with the code duplication for now, we can look into cleaning it up but it's not critical. To get this change in, I think we need to either bump up the serialized doc version to force an update to all dcf files, or support backward compatibility for old dcf files that have the old component override data. I think we decided to go with the version bump since it's easier and this may be the last time we can do this, if @kovmarci86 and @urpatil are ok with that.

rylin8 (approved):

PR #2110, opened 2025-03-11
Add README file to crates/dc_jni

yiqunw700:
Towards: #2103


rylin8 (approved):
I don't think we will publish `dc_jni` initially, but this README is fine.

PR #2123, opened 2025-03-14
Fix shader brush not initialized exception and add two improvements

yiqunw700:
1. Add image uniform customization support from drawable resource. Also can do image customization from bytes array but no examples and no helper functions created. Recommended to use drawable resource to do customization.
2. Fix shader brush not initialized exception on api level 32 where fallback color is not set. Also updated the figma design doc to test those cases.
3. BuildProcessor now set the shader time uniform state in the generated code. It also adds a designFeatures to the DesignDoc annotation to decide if the time uniform state will be set to the customization context or not.

Fixes: #2124


annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
```diff
@@ -29,6 +29,7 @@ annotation class DesignDoc(
     val id: String,
     val designVersion: String = "",
     val customizationInterfaceVersion: String = "0",
+    val designFeatures: Array<String> = [], // Allowed values: "shader"
```
yiqunw700, line 32: This is added so I only set the shader time uniform state to customization context when it has the shader feature.

But if you're ok with setting the time uniform state to customization context even no shaders are using, I can remove this field. It just creates a unused float state and shader uniform state in that case.

yiqunw700:
> I assume this needs to be cherry-picked into the 0.34 RC?

Not quite sure what cherrypicking will end up with? 0.34.0 rc3? It would be great to cherry-pick but that crash only happens on 32 and below. Not critical. I feel it ok not to cherrypick.

designcompose/src/main/java/com/android/designcompose/FrameRender.kt
```diff
@@ -941,27 +957,29 @@ internal fun getShaderBrush(
     appContext: Context,
     density: Float,
     asBackground: Boolean = true,
-): Brush {
+): Brush? {
```
rylin8, line 960: So previously, even though this function returned `Brush`, it could be `null` for API 32 if the fallback color was null? Shouldn't that have been a compile error?

annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
rylin8: I'm ok with it, let's add it to the documentation.

rylin8 (approved):

annotation/src/main/java/com/android/designcompose/annotation/Builder.kt
yiqunw700: Will do.

Issue #2148, opened 2025-04-02
Progress bar starts from non-zero offset


yiqunw700:

yiqunw700:
<img width="376" alt="Image" src="https://github.com/user-attachments/assets/90cf8c7d-aa0b-418d-b6b2-3c535b2e7392" />

iamralpht:
The sample here looks good (I'm resurrecting my checkout to test locally) -- what was the behavior before that's corrected by recording the startX/Y from the plugin?

yiqunw700:
Before the progress bar starts from width 0 to full width of parent.

https://github.com/user-attachments/assets/05e95d80-e0e5-4944-8ce7-40ae37befe53

The indicator's frame has a padding horizontally which is half of the indicator's width.

<img width="305" alt="Image" src="https://github.com/user-attachments/assets/3fb128a7-fdba-4233-b99d-b6fd63b1a6a5" />

iamralpht:
Oh, I see -- yeah, that looks totally wrong :p.

PR #2150, opened 2025-04-02
Add a RB hvac example to validation

yiqunw700:
Towards: #2138 
Fixes: #2148


iamralpht:
This change looks OK to me. Do we still need to bump the DCF version for the proto change?

yiqunw700:
> This change looks OK to me. Do we still need to bump the DCF version for the proto change?

This release will have a version bump done in pr #2090. I recall Rob told me that 1 version bump per release should work. So I think it is ok not to bump the version in this pr.

iamralpht (approved):

support-figma/extended-layout-plugin/src/code.ts
```diff
@@ -1001,15 +1001,17 @@ if (figma.command === "sync") {
...
+        barData.startY = node.minHeight ?? 0;
         barData.endY = parent.height;
+        const height = percentToValue(value, 0.01, barData.endY - barData.startY);
+        node.resize(parent.width, height + barData.startY);
+        node.y = parent.height - height;
       } else {
-        const width = percentToValue(value, 0.01, parent.width);
-        node.resize(width, parent.height);
-        node.y = 0;
+        barData.startX = node.minWidth ?? 0;
```
rylin8, line 1010: In the hvac example you made, I don't see a min width set on the node -- am I missing something? Also, can we update the plugin to take this startX/startY into account?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/HvacComponents.kt
```diff
@@ -0,0 +1,426 @@
...
+    val coolingOnState = remember { mutableStateOf(false) }
+    val fanDirection = remember { mutableIntStateOf(0) }
+    val recycleAirState = remember { mutableStateOf(false) }
+    val defrostFrontState = remember { mutableStateOf(false) }
+    val defrostRearState = remember { mutableStateOf(false) }
+    val seatHeaterDriverState = remember { mutableIntStateOf(0) }
+    val seatHeaterPassengerState = remember { mutableIntStateOf(0) }
+    val fanSpeedState = remember { mutableFloatStateOf(100f) }
+
+    val dayNightMode = remember { mutableStateOf(HvacVariant.DayNightMode.Default) }
```
rylin8, line 171: Is there a way to toggle this?

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/HvacComponents.kt
```diff
@@ -0,0 +1,426 @@
...
+        high,
+    }
+}
+
+@DesignDoc(id = "YUU5qZIHRlSwXEbPPk3INh")
+interface HvacComponents {
+    @DesignComponent(node = "#stage")
+    fun hvacPanel(
+        @Design(node = "#hvac_on_off")
+        hvacOnOff: @Composable ((ComponentReplacementContext) -> Unit),
```
rylin8, line 86: Instead of using `ComponentReplacementContext` customizations, another way to approach this could have been to make all the buttons be instances of their own components, and then changing their variants using `@DesignVariant` parameters. Is there a reason you went with component replacement? Just curious.

yiqunw700: Oh I was trying to make it generic. I created a few buttons that are the instances of the ToggleButton and tried to do customizations on those instances with "DesignVariant" then I realize it can not be done that way.

support-figma/extended-layout-plugin/src/code.ts
yiqunw700: No. I was reverting the show what it was look like before but probably forgot to revert back. Thanks for catching it.

For the latter, I am wondering if you would like to show the `startX: 88` in the plugin. Cause the progress is still from 0 to 100. Right now when it changes the progress in the plugin, it should be able to reflect the startX in the progress in the node.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/HvacComponents.kt
```diff
@@ -0,0 +1,426 @@
...
+        @DesignVariant(property = "#icon-enabled") iconEnabled: HvacVariant.EnabledState,
+        @DesignVariant(property = "#icon") icon: HvacVariant.Icon,
+    )
+
+    @DesignComponent(node = "#FanSpeedControl")
+    fun fanSpeedControl(
+        @DesignVariant(property = "#fan-speed-control-enabled") enabled: HvacVariant.EnabledState,
+        @Design(node = "#speed") speed: String,
+        @Design(node = "#progress") progress: Meter,
+        @Design(node = "#indicator") progressIndicator: Meter,
```
rylin8, line 153: Can the indicator be a child of the #progress frame so that you only need only `Meter` customization? I saw you had that in the main Figma file, but not in the branch (which this example uses).

support-figma/extended-layout-plugin/src/code.ts
rylin8: Nope, I got confused with this code for some reason, it's fine.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/HvacComponents.kt
yiqunw700: Filed ticket #2157 to track this as discussed at our meeting.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/HvacComponents.kt
yiqunw700: Done. Now it will toggle along with the system light/dark theme change.

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/HvacComponents.kt
```diff
@@ -0,0 +1,443 @@
...
+    val fanDirection = remember { mutableIntStateOf(0) }
+    val recycleAirState = remember { mutableStateOf(false) }
+    val defrostFrontState = remember { mutableStateOf(false) }
+    val defrostRearState = remember { mutableStateOf(false) }
+    val seatHeaterDriverState = remember { mutableIntStateOf(0) }
+    val seatHeaterPassengerState = remember { mutableIntStateOf(0) }
+    val fanSpeedState = remember { mutableFloatStateOf(100f) }
+
+    val isSystemInDarkMode = isSystemInDarkMode(LocalContext.current)
+    val dayNightMode = remember {
```
rylin8, line 182: Since this only changes by changing your system setting, is there any benefit in using `remember` and `mutableStateOf()`?

yiqunw700: Done.

rylin8 (approved):

PR #2151, opened 2025-04-02
Support shape overrides and nested component overrides

yiqunw700:


crates/dc_bundle/src/definition/view.rs
```diff
@@ -334,6 +334,18 @@ impl ViewData {
                 return Some(other.clone());
             }
         }
+        if let Some(View_data_type::Container { 0: Container { shape, .. } }) =
```
iamralpht, line 337: I'm not familiar with what proto is generating here, but do you want { 0: ... } or can you do ( instead (e.g.: if it's a tuple).

yiqunw700: Looks like I can use `if let Some(View_data_type::Container(container)) = self.view_data_type` or use the proto generated `container()` function to get the container. Switched to the latter.

PR #2162, opened 2025-04-09
Speed up fetch of components in other files

### LINKED ISSUE #2161: Retrieving component data from other files is very slow
When a document contains component instances that point to components in other files, the retrieval process can be very slow. It seems that multiple instances of the same component result in multiple fetches, even though only one is needed.

rylin8:
Fixes #2161. When retrieving component data from other files, keep track of the components retrieved so we don't download them again when encountering other component instances from the same component.


rylin8:
fyi @Gurupreet this should speed up fetching in the future.

yiqunw700 (approved):

rylin8:
Yes, 0.35.

PR #2169, opened 2025-04-11
Create Scalable UI plugin

### LINKED ISSUE #2180: Create plugin for Scalable UI
Create a plugin to input data to component sets for scalable ui. Add support to parse this additional data from Figma files into a DCF file so that it can be read by scalable ui.

rylin8:
Fixes #2180.

- Create a plugin to add scalable UI data
- Proto and Figma schema changes to parse and store the data

To fetch a file to be used to load scalable ui panel states, run the fetch tool with the --scalableui option and the root node that contains all the panel component instances. For example:

cargo run --bin fetch --features fetch -- --doc-id <doc_id> --nodes "#stage" --output ScalableSystemUi_<doc_id>.dcf --scalableui

### DIFF
--- a/crates/dc_bundle/src/proto/definition/element/scalable.proto
+++ b/crates/dc_bundle/src/proto/definition/element/scalable.proto
@@ -1,0 +1,92 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package designcompose.definition.element;
+
+option java_multiple_files = true;
+option java_package = "com.android.designcompose.definition.element";
+option optimize_for = LITE_RUNTIME;
+
+// An event that causes one or more variants to change
+message Event {
+  string event_name = 1;
+  string event_tokens = 2;
+  string from_variant_id = 3;
+  string from_variant_name = 4;
+  string to_variant_id = 5;
+  string to_variant_name = 6;
+}
+
+// A key frame that shows the given variant at the given frame
+message Keyframe {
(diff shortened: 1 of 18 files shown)


rylin8:
FYI this is the code for the scalable ui Figma plugin and the code to parse and store the data.

crates/dc_bundle/src/proto/definition/element/scalable.proto
```diff
@@ -0,0 +1,92 @@
...
+  string role = 3;
+  string default_variant_id = 4;
+  string default_variant_name = 5;
+  repeated Event events = 6;
+  repeated KeyframeVariant keyframe_variants = 7;
+  repeated string variant_ids = 8;
+}
+
+// Represents a dimension in pixels or percentage
+message ScalableDimension {
```
iamralpht, line 58: I guess we're doing it this way bc we don't have dynamic layout in ScalableUI yet (but I think we should remove this and use DesignCompose layout).

rylin8: I just went with the simplest path to get started, which was to assume a fixed size device per design.

crates/dc_bundle/src/proto/definition/view/node_style.proto
```diff
@@ -103,4 +104,6 @@ message NodeStyle {
   // Runtime shader from the shader plugin as background.
   // When shader is present, background will not be drawn.
   optional element.ShaderData shader_data = 43;
+
+  optional element.ScalableUIData scalable_data = 44;
```
iamralpht, line 108: I could see this going into the parent object (View?) since stuff like id/name and (soon) role isn't really style related and I think the style stuff (layout, opacity) should eventually fall out of the existing style definitions.

This is ok to start with, though, since it puts everything in one place.

iamralpht (approved):
LGTM, one feedback on how to do the flag to control the behavior of skipping hidden nodes.

crates/figma_import/src/document.rs
```diff
@@ -296,9 +296,10 @@ impl Document {
         error_list: &mut Vec<String>,
         error_hash: &mut HashSet<String>,
         completed_hash: &mut HashSet<String>,
+        skip_hidden: bool,
```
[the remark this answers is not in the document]
rylin8: good idea, fixed.

crates/dc_bundle/src/proto/definition/view/node_style.proto
rylin8: 👍

PR #2171, opened 2025-04-14
Fix font loader to resolve fonts correctly

### LINKED ISSUE #2170: System fonts do not render
System fonts can be added using DesignSettings.addFontFamily(), but specifying that font in a Figma text node will not render with the font.

rylin8:
Fixes #2170. Use font family loader instead of deprecated font resource loader


iamralpht (approved):

iamralpht:
Does this clean up the deprecation warnings around FontLoader?

rylin8:
> Does this clean up the deprecation warnings around FontLoader?

Yes

PR #2182, opened 2025-04-29
Support interactable progress bars and markers

rylin8:
Fixes https://github.com/google/automotive-design-compose/issues/2146. This change adds an option in the Dials & Gauges plugin UI to set a progress bar or marker as draggable. Doing so will result in adding a Modifier on progress nodes that track tap/drag and then updates the value of the progress bar/marker. A new OnProgressChangedCallback() customization can be set, allowing a developer to receive updates whenever a tap or drag causes the value of the progress bar to change.


iamralpht (approved):
LGTM to merge; can you add some documentation?

PR #2184, opened 2025-05-02
Call scroll state callback on init

### LINKED ISSUE #2172: DesignScrollState doesn't emit initial values until scroll
The DesignScrollState will not emit the "containerSize", "contentSize", "maxValue", and "value" until the list is scrolled. This causes issues with creating a scrollbar as the metrics will not be set until the list is moved.
The current work around is to trigger scrollBy(0) to get these values at initialization.

rylin8:
Fixes #2172.


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -650,6 +650,9 @@ fun SquooshRoot(
             }
             else -> {}
         }
+    }
+    val scrollState = rememberScrollableState { delta ->
```
iamralpht, line 654: `scrollState` is still memoized, so the `LaunchedEffect` won't get re-run even though a new value of `setScrollState` is computed every run through.

iamralpht (approved):

PR #2186, opened 2025-05-04
Fix #2185: remove stale reference to generateSerdegenCode from idea config

### LINKED ISSUE #2185: Android Studio build fails missing generateSerdegenCode
I think this is from a stale idea config -- we have these files checked in, and they don't match the gradle tasks after migrating to protobuf and removing serde/bincode from the DCF definition.

iamralpht:


iamralpht:
This fixes an error when opening the project in Android Studio the first time. I've never really been sure how to manage the .idea files (or if we can just delete them from git), since they seem to be machine derived from the gradle config.

rylin8 (approved):

PR #2187, opened 2025-05-06
Test for backward compatibility of old dcf files

### LINKED ISSUE #2126: Test for backward compatibility of old dcf files
Moving forward, protobuf changes should not break backward compatibility of loading old dcf files. Add some tests to ensure old dcf files still load after protobuf changes. This would entail checking in one or more old dcf files.

Investigate having one consolidated place in code where special handling of old versions can be done.

rylin8:
Fixes #2126. This commit makes a change to the field containing a text node's font color, renaming it from text_color to font_color. Any updates to or new DCF files will save text color to font_color in node_style instead of text_color. This change's only purpose is to introduce a protobuf change and demonstrate our capability to support old protobufs that still set text_color. This is supported by using font_color if it is set, and if not, using text_color.

To facilitate support of old DCF files as protobufs continue to evolve, a new utility file utils/ProtoVersions.kt has some functions to manage checking of old fields. This file should grow when a protobuf change that requires code to support old versions is made.

Additionally, a new test has been added to test for old protobufs. Currently this tests that v27 protobufs (that still use text_color) have a header that says v27, and that they render correctly, including a screenshot. New tests should be added when protobufs again change.


rylin8:
Not sure what the best way to test for backward compatibility is without making a protobuf change, so here is an initial attempt where we make a proto change and add a test to ensure an old DCF file before the proto change still renders correctly.

iamralpht (approved):
I guess that from now on, when CI fails due to screenshot tests failing, we should be updating the code rather than rebasing all of the DCF files.

PR #2211, opened 2025-05-21
Fix TapCallback on replacement content lists

### LINKED ISSUE #2205: TapCallback on list widget content called on wrong node
When a list widget is used for horizontal or vertical content with scrolling enabled, and the content has a TapCallback customization, tapping items after the view has been scrolled calls the callback on the wrong node.

rylin8:
Fixes #2205


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -1264,7 +1277,9 @@ private fun SquooshChildLayout(modifier: Modifier, child: SquooshChildComposable
             // Place all children at the origin (with the same size constraints that
             // we were given).
             layout(layoutWidth, layoutHeight) {
-                placeables.forEach { child -> child.placeRelative(0, 0) }
+                val x = -scrollOffset.value.x.roundToInt()
```
iamralpht, line 1280: I think reading `scrollOffset` here means that layout will get re-run whenever that state value changes. So I suspect there's a performance impact here, but I'm not sure if there's an alternative in Compose (like rendering the offset only until there's an event, and then doing the layout).

Or maybe Compose layout is fast enough that this is a non-issue...

farivar-google (approved):

PR #2230, opened 2025-06-18
Update VCS mapping for protobuf submodule

timothyfroehlich:
This change adds a new VCS mapping for the `crates/dc_bundle/src/proto` directory. This directory is a Git submodule containing protobuf definitions.

The copyright header was also removed from `.idea/vcs.xml`. Since this is a generated file (and just a workspace configuration file) I don't think it's necessary.

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,22 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
- Copyright 2024 Google LLC
-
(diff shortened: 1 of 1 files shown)


farivar-google (commented):
this cl is just removing things, no? or am I missing something?

timothyfroehlich:
> this cl is just removing things, no? or am I missing something?

Line 5 is added. Android Studio has been trying to add it ever since I initialized the submodule.

PR #2244, opened 2025-06-30
Fix #2243: ignore SECTION node type after parsing.

### LINKED ISSUE #2243: Update parser to ignore "section" nodes
There's a new [SECTION](https://www.figma.com/developers/api#node-types) node type, which causes the parser to abort. It's used for organizing a doc, and has some impact on prototyping actions. I think it's safe to ignore this node type, as a section cannot be embedded in a frame (so won't be encountered within a component).

iamralpht:


iamralpht:
My first public PR entirely authored by [gemini-cli](https://github.com/google-gemini/gemini-cli)! I just pointed it at the issue and it worked through the change!

iamralpht:
I think I still want this to recurse, rather than skipping the children of a section.

crates/figma_import/src/transform_flexbox.rs
```diff
@@ -1865,6 +1865,22 @@ fn visit_node(
...
     // Figure out the ViewShape from the node type.
     let view_shape = match &node.data {
+        figma_schema::NodeData::Section {} => {
+            // Don't render sections at all.
+            return Ok(View::new_rect(
+                &node.id,
+                &node.name,
+                ViewShape::new_rect(view_shape::view_shape::Box::default()),
+                style,
+                None,
```
farivar-google, line 1875: nit: plz add /* variable name */  and other places.

iamralpht: It turned out I didn't need any of this part of the change, since the default handling for an unknown node data type is good enough.

PR #2247, opened 2025-07-03
Add SurfaceView with ReplacementContent Validation test

Gurupreet:


Gurupreet:
Do I need to place document anywhere, Tests are failing with.

`RenderAllExamples > testRender[TestConfig(fileName=Layout-Replacement-SurfaceView, fileComposable=androidx.compose.runtime.internal.ComposableLambdaImpl@4f74980d, fileClass=com.android.designcompose.testapp.validation.examples.LayoutReplacementSurfaceViewTestGen)] FAILED
    java.lang.AssertionError: Failed to assert the following: (DocRenderStatus = 'Rendered')
    Can't retrieve node at index '0' of 'DocClass = 'com.android.designcompose.testapp.validation.examples.LayoutReplacementSurfaceViewTestGen''
    There are no existing nodes for that selector.
`

iamralpht:
It can’t find the serialized doc, but I think there is a script in
dev-scripts which will fetch all of the docs. Alternatively, you can fetch
the serialized doc from your emulator image where you ran the validation
app, and put it with the others…

iamralpht:
There's also a script to generate reference snapshots so that we don't have this diff reported -- then when we get the fix in, we can update the reference snapshot to show the successful case.

iamralpht:
Thanks, Guru -- you should be able to add it to the merge queue now, and you, me, or Saeid can bisect to find the culprit.

PR #2254, opened 2025-07-08
Update SurfacView to use ComponentReplacementContext

Gurupreet:


docs/_docs/modifiers/modifiers.md
```diff
@@ -398,6 +398,44 @@ HelloWorldDoc.MainFrame(
...
+    )
+}
+
+// The custom Composable that wraps the SurfaceView.
+@Composable
+fun DummySurfaceView(modifier: Modifier = Modifier) {
+    AndroidView(
+        factory = { context ->
+            SurfaceView(context).apply {
+                // In a real app, you would initialize the SurfaceView here.
```
farivar-google, line 431: maybe for completeness add a comment that the draw needs to happen when the surface onCreated is called?

farivar-google (approved):

docs/_docs/modifiers/modifiers.md
Gurupreet: I think that's already part of full demo we added so felt redundant here.

PR #2257, opened 2025-07-15
Feature/nested component with variables

farivar-google:


crates/figma_import/Cargo.toml
```diff
@@ -42,6 +42,7 @@ clap = { workspace = true, optional = true }
 
 # logging
 log.workspace = true
+env_logger = { workspace = true, optional = true }
```
iamralpht, line 45: @kovmarci86 any concerns with a new dependency?

iamralpht:
Huh - are the visual diffs from paths that are connected to variables? (Presumably not from the macOS -> Linux test runner change that just went in)

iamralpht:
Oh - I think the differences are from the macOS -> Linux switch, bc they're in fills & strokes that are rendered by shaders on device (but shaders are unsupported in roborazzi).

farivar-google:
> Changes LGTM - do old DCF files still open OK with this change applied? (I think that should be the case, but just checking)

all tests are passing. roborazzi isn't unhappy :) so I'd say yes.

Gurupreet (approved):
LGTM,   Nit: We can move "VariableID" prefix to a const value in future.

PR #2264, opened 2025-07-24
feat(coverage): Add code coverage reporting for Kotlin and Rust

farivar-google:
Integrates JaCoCo for Kotlin and cargo-tarpaulin for Rust to generate code coverage reports.

- Configures a new `jacocoTestReport` Gradle task to aggregate coverage data for core library modules.
- Modifies the `main.yml` GitHub Actions workflow to:
  - Generate coverage reports on pull requests.
  - Upload reports to Codecov.io for analysis and tracking.
- Adds `cobertura.xml` to `.gitignore` to exclude Rust coverage reports from version control.

This enables automated coverage checks, helping to maintain and improve code quality by guarding against drops in test coverage.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -38,6 +38,7 @@ plugins {
     alias(libs.plugins.jetbrains.compose) apply false
     alias(libs.plugins.ksp) apply false
     alias(libs.plugins.roborazzi) apply false
+    jacoco
 }
 
 // Apply some of our convention plugins to the reference apps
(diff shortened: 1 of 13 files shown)


build.gradle.kts
```diff
@@ -48,3 +49,25 @@ for (projectName in listOf("tutorial-app", "helloworld-app", "cluster-demo", "me
...
+        val sourceSets = it.extensions.findByType(SourceSetContainer::class.java)
+        sourceSets?.findByName("main")?.allSource
+    })
+    sourceDirectories.setFrom(source)
+    classDirectories.setFrom(files(subprojects.filter { it.name in coreModules }.map {
+        val sourceSets = it.extensions.findByType(SourceSetContainer::class.java)
+        sourceSets?.findByName("main")?.output
+    }))
+    executionData.setFrom(fileTree(project.rootDir) {
+        include("**/build/jacoco/test.exec")
```
timothyfroehlich, line 67: This might be too broad, but probably not a big deal.

Double check that the Jacoco plugin doesn't already have an "aggregate results" task or something like that

annotation/build.gradle.kts
```diff
@@ -31,3 +32,7 @@ publishing {
         }
     }
 }
+
+tasks.withType<JacocoReport> {
+    enabled = tasks.withType<Test>().any { it.state.executed }
+}
```
timothyfroehlich, line 38: Why is this needed?

integration-tests/app-common/build.gradle.kts
timothyfroehlich, line 1: I wouldn't add coverage for the test apps personally, not unless you can filter the results to just the core packages (which would probably have to rely on the gradle config, not the package name). You're going to get a very low number because of all of the generated KSP code.

test/build.gradle.kts
timothyfroehlich, line 1: These are just testing libraries, I can't remember if we're even publishing this one. I don't think we need coverage here.

timothyfroehlich (changes requested):

annotation/build.gradle.kts
farivar-google: That line is a safeguard to make our build more robust and efficient.

  The core issue is that we apply the jacoco plugin to several modules, but not all of them have a test source set or any tests to run. When Gradle executes, it sees there are no tests in these modules and skips their test task.

  Without the line you mentioned, the jacocoTestReport task for those modules would still try to run. This would cause a problem for two main reasons:

   1. Potential Build Failures: The JaCoCo task would fail because it depends on test execution data (a test.exec file) that was never created.
   2. Inefficiency: It would create unnecessary work for the build process.

  The code tasks.withType<JacocoReport> { enabled = tasks.withType<Test>().any { it.state.executed } } solves this by adding a simple rule:

  "Only enable the JaCoCo report task if at least one test task in this module has actually run."

  This ensures that we only generate coverage reports in modules that have tests, which prevents build failures and makes our overall build process cleaner and more reliable.

timothyfroehlich (approved):

PR #2267, opened 2025-07-24
fix(dc_layout): Correct flex direction mapping for row-reverse

farivar-google:
The  variant was incorrectly mapped to . This commit corrects the mapping to .


iamralpht (approved):

kovmarci86:
Figma does not seem to support this feature yet. Horizontal/vertical layouts are supported, but not in reverse order.

Found an open feature request for the same:
https://forum.figma.com/suggest-a-feature-11/auto-reverse-auto-layout-order-of-display-13803

farivar-google:
> Can we get a render test for this case?

seems like Figma doesn't support this yet. opened an issue to do it later.
https://github.com/google/automotive-design-compose/issues/2270

PR #2353, opened 2026-01-27
Set version for dc_layout to 0.38.0

dipenpradhan:


kovmarci86 (approved):
I think we need to improve how we version our crates, but likely not in this change. Approved.

PR #2371, opened 2026-02-24
Fix: Prevent InteractionState overlays from persisting across composable roots

### LINKED ISSUE #2329: Figma Prototype is placed on top of Figma Frame
**Setup**

We have multiple Figma Frames which serve as an Entry e.g #Frame/A and #Frame/B 

in Code:
```
@DesignComponent(node = "#Frame/A", hideDesignSwitcher = true)
    fun FrameA(...)
```

```
@DesignComponent(node = "#Frame/B", hideDesignSwitcher = true)
    fun FrameB(...)
```

and we switch between those frames programatically 

```
when(condition) {
        "A"->FigmaDoc.FrameA(...)
        "B"-> FigmaDoc.FrameB(...)
        else-> ...
}

```

--------------

**Issue Description**

#Frame/A has Figma prototype interactions that navigate to other Figma frames (e.g., #Frame/A/Detail). When triggering this prototype interaction on hardware or an emulator, it appears that the prototype opens on top of the Android activity. As a result, when we programmatically switch to FigmaDoc.FrameB(...), the prototype layer is still visible, and it seems as though FrameB is opened behind it.

**Expected Behavior**

When switching programmatically to a different frame, any prototype frames from the previous layer should be removed rather than staying overlaid on top of the new frame.

**Version:** 0.38.0

ghost:
## Description
This PR addresses Issue #2329, where prototype features (like Overlays and Navigation) from one `@DesignComponent` incorrectly persist on top of a newly rendered frame when switching between them programmatically.

## Root Cause
[InteractionState](cci:2://file:///usr/local/google/home/priyanksingh/automotive-design-compose/designcompose/src/main/java/com/android/designcompose/InteractionState.kt:172:0-237:1) is maintained globally per Figma Document ID via `InteractionStateManager.states`. However, there was no lifecycle management for this global state. Consequently, when a [SquooshRoot](cci:1://file:///usr/local/google/home/priyanksingh/automotive-design-compose/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt:225:0-924:1) was destroyed, its active overlays remained in the global state and bled into the subsequent root for the same document.

## Changes
* **Lifecycle Tracking**: Implemented reference-counted lifecycle tracking (`activeRoots`) within `InteractionStateManager`.
* **Registration Methods**: Added `registerRoot(docId)` and `unregisterRoot(docId)` to correctly track active view roots.
* **State Cleanup**: `states.remove(docId)` is now invoked when the active root count drops to `0`.
* **Side Effects**: Integrated `DisposableEffect(docId)` directly into [SquooshRoot.kt](cci:7://file:///usr/local/google/home/priyanksingh/automotive-design-compose/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt:0:0-0:0) to monitor component lifecycles.
* **Testing**: Added [InteractionStateManagerTest.kt](cci:7://file:///usr/local/google/home/priyanksingh/automotive-design-compose/designcompose/src/test/kotlin/com/android/designcompose/InteractionStateManagerTest.kt:0:0-0:0) unit tests to explicitly verify the registry counting logic.

Fixes #2329

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/InteractionState.kt
+++ b/designcompose/src/main/java/com/android/designcompose/InteractionState.kt
@@ -829,8 +829,33 @@ internal fun InteractionState.getPressedTapCallback(nodeId: String): TapCallback
 /// of the view tree organization.
 internal object InteractionStateManager {
     val states: HashMap<DesignDocId, InteractionState> = HashMap()
+    internal val activeRoots: HashMap<DesignDocId, Int> = HashMap()
+
+    fun registerRoot(docId: DesignDocId) {
+        val current = activeRoots[docId] ?: 0
+        activeRoots[docId] = current + 1
+    }
+
+    fun unregisterRoot(docId: DesignDocId) {
+        val current = activeRoots[docId] ?: 0
+        if (current <= 1) {
+            activeRoots.remove(docId)
+            states.remove(docId)
+        } else {
+            activeRoots[docId] = current - 1
+        }
+    }
 }
 
 internal fun InteractionStateManager.stateForDoc(docId: DesignDocId): InteractionState {
     return states.getOrPut(docId) { InteractionState() }
+}
+
+@Composable
+internal fun InteractionStateManager.rememberStateForDoc(docId: DesignDocId): InteractionState {
+    DisposableEffect(docId) {
+        registerRoot(docId)
+        onDispose { unregisterRoot(docId) }
+    }
+    return stateForDoc(docId)
 }
--- a/designcompose/src/test/kotlin/com/android/designcompose/InteractionStateManagerTest.kt
+++ b/designcompose/src/test/kotlin/com/android/designcompose/InteractionStateManagerTest.kt
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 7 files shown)


designcompose/src/test/kotlin/com/android/designcompose/InteractionStateManagerTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [35])
+class InteractionStateManagerTest {
+
+    @Before
+    fun setUp() {
+        InteractionStateManager.states.clear()
+        // Reset the active roots using reflection or manually clearing it for tests if needed
+        val field = InteractionStateManager::class.java.getDeclaredField("activeRoots")
```
iamralpht: Can we expose this to the test without reflection and without making test methods part of the API?

designcompose/src/main/java/com/android/designcompose/InteractionState.kt
```diff
@@ -829,6 +829,22 @@ internal fun InteractionState.getPressedTapCallback(nodeId: String): TapCallback
 /// of the view tree organization.
 internal object InteractionStateManager {
     val states: HashMap<DesignDocId, InteractionState> = HashMap()
+    private val activeRoots: HashMap<DesignDocId, Int> = HashMap()
+
+    fun registerRoot(docId: DesignDocId) {
+        val current = activeRoots[docId] ?: 0
+        activeRoots[docId] = current + 1
```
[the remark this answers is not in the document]
priyanksingh-web: I've added a new @Composable function called rememberStateForDoc to InteractionStateManager. It handles both fetching the state and automatically managing the root lifecycle (registering and unregistering) via a DisposableEffect. I've updated SquooshRoot to just call this composable instead.

designcompose/src/test/kotlin/com/android/designcompose/InteractionStateManagerTest.kt
priyanksingh-web: Yes, good call. I've updated activeRoots to be internal within InteractionStateManager. This allows our tests to access and clear the state directly without using reflection, while keeping it hidden from the external public API.

PR #2402, opened 2026-03-31
Squoosh Designer Plugin: Figma plugin for orchestrating layout animations

kovmarci86:
Work-in-progress beta/previews state of a Figma plugin that allows creating complex animations between variant changes.


.github/workflows/roborazzi-comment.yml
```diff
@@ -0,0 +1,136 @@
+name: Roborazzi Comment
```
dipenpradhan: Is splitting roborazzi workflow required for this PR to pass GH Actions? Can we split this into a separate change

crates/dc_figma_import/src/figma_schema.rs
```diff
@@ -1044,8 +1044,10 @@ pub struct Node {
     pub absolute_bounding_box: Option<Rectangle>,
     #[serde(default)]
     pub absolute_render_bounds: Option<Rectangle>,
-    #[serde(default)]
+    #[serde(rename = "sharedPluginData", default)]
```
dipenpradhan: Can this field name stay original? This might introduce compatibility issues with different versions of figma plugin

dipenpradhan (changes requested):

support-figma/DC Squoosh designer/COLORS.md
```diff
@@ -0,0 +1,61 @@
+# Timeline Color Scheme
```
dipenpradhan: Should this go into README / docs?

support-figma/DC Squoosh designer/MANUAL.md
```diff
@@ -0,0 +1,100 @@
+# DC Squoosh Designer Manual
```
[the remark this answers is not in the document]
dipenpradhan: Since this is a user manual, it should probably go into https://github.com/google/automotive-design-compose/tree/main/docs/_docs

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
 resolver = "2"
-members = ["crates/dc_figma_import", "crates/dc_jni", "crates/dc_layout", "crates/dc_bundle"]
+members = ["crates/dc_figma_import", "crates/dc_jni", "crates/dc_layout", "crates/dc_bundle", "support-figma/DC Squoosh designer/rs"]
```
dipenpradhan: Can this rust code go into a `crates/*` location instead of `support-figma/`?

dipenpradhan (changes requested):

dipenpradhan (changes requested):

Cargo.toml
kovmarci86: done

crates/dc_figma_import/src/figma_schema.rs
```diff
@@ -1046,6 +1046,8 @@ pub struct Node {
     pub absolute_render_bounds: Option<Rectangle>,
     #[serde(default)]
     pub shared_plugin_data: HashMap<String, HashMap<String, String>>,
+    #[serde(skip)]
```
dipenpradhan: Why do we skip this here and then lazy-load / lazy-parse? dc_figma_import is used only for importing, should be ok to keep parsing enabled?

crates/dc_figma_import/Cargo.toml
```diff
@@ -39,6 +39,7 @@ clap = { workspace = true, optional = true }
 
 # logging
 log.workspace = true
+env_logger = "0.10.0"
```
dipenpradhan: Can this come from `env_logger.workspace = true`?

support-figma/dc_squoosh_designer/rs/examples/simple_interpolation.rs
```diff
@@ -0,0 +1,64 @@
...
+            },
+            "customKeyframeData": {
+                "Button-cornerRadius": "Linear|0.5,WzUwLjAsIDUwLjAsIDUwLjAsIDUwLjBd,Linear"
+            }
+        }
+    }"#;
+
+    // 2. Parse the Variant from JSON
+    let variant: Variant = from_str(json_data).expect("Failed to parse Variant JSON");
+    println!("Parsed Variant: {}", variant.name);
```
dipenpradhan: `log::`?

support-figma/dc_squoosh_designer/rs/src/lib.rs
```diff
@@ -0,0 +1,939 @@
...
+            if let Ok(mut parsed) = serde_json::from_str::<ParsedTimelineData>(encoded_data) {
+                // Ensure sorted
+                parsed.keyframes.sort_by(|a, b| {
+                    a.fraction.partial_cmp(&b.fraction).unwrap_or(std::cmp::Ordering::Equal)
+                });
+                return Ok(parsed);
+            }
+        }
+
+        // LEGACY PARSING
```
dipenpradhan: This file is new, what the legacy version?

support-figma/dc_squoosh_designer/rs/src/lib.rs
```diff
@@ -0,0 +1,939 @@
...
+
+    /// Retrieves all parsed timeline data for a specific node.
+    pub fn get_for_node(&self, node: &str) -> Option<Arc<NodeTimelines>> {
+        self.timelines.get(node).cloned()
+    }
+
+    /// Retrieves the parsed timeline data for a specific node and property.
+    pub fn get(&self, node: &str, prop: AnimatableProperty) -> Option<&ParsedTimelineData> {
+        // Find the node's NodeTimelines, then find the property within it.
+        // Needs a workaround since we return an Arc and want to return a reference to its contents. Wait, the old getter returned a reference. If we return a reference tied to `self`, that's fine.
```
dipenpradhan: Is this comment still relevant?

crates/dc_figma_import/Cargo.toml
kovmarci86: done.

support-figma/dc_squoosh_designer/src/plugin/compare.ts
```diff
@@ -0,0 +1,87 @@
...
+    const children2Map = new Map(
+      children2.map((child) => [child.name, child]),
+    );
+
+    for (const child1 of children1) {
+      const child2 = children2Map.get(child1.name);
+      if (child2) {
+        await compareNodes(child1, child2, currentPath);
+      } else {
+      }
```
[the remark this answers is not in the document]
kovmarci86: added logs.

ghost (commented):
[Review deleted per user request. Moved to inline comments.]

ghost (commented):
[Review deleted per user request. Moved to inline comments.]

ghost (commented):
[Review moved to inline comments]

ghost (commented):
Adding my architectural review points as inline comments for better context.

support-figma/dc_squoosh_designer/src/timeline/serialization.ts
```diff
@@ -0,0 +1,140 @@
+import { Keyframe, ArcData } from "./types";
+
+// Serialization format: "fraction,value,easing;fraction,value,easing;..."
+// Value will be JSON.stringify-ed if it's an object, and then base64 encoded.
+
+/**
```
priyanksingh-web: This comment (`and then base64 encoded`) seems outdated. Looking at `serializeKeyframes` below, we switched to native JSON serialization instead of the legacy pipe-delimited base64 format. Consider removing this to avoid confusion for future maintainers.

crates/dc_squoosh_animation/src/lib.rs
```diff
@@ -0,0 +1,647 @@
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::sync::Arc;
+
+// --- High Level Runtime Types ---
+
+/// Supported easing functions for interpolation.
+#[derive(Debug, Clone, PartialEq, Copy, Serialize, Deserialize)]
+pub enum Easing {
```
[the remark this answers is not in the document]
kovmarci86: ack. thanks!

crates/dc_figma_import/src/document.rs
```diff
@@ -49,6 +50,8 @@ pub enum HiddenNodePolicy {
     Keep,
 }
 
+/// Helper function to deserialize JSON strings while reporting errors mapped with the exact property path.
```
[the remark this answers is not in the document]
kovmarci86: ack

support-figma/dc_squoosh_designer/src/timeline/serialization.ts
kovmarci86: Removed.

Cargo.toml
```diff
@@ -26,7 +26,9 @@ dc_layout = { version = "0.38.2", path = "crates/dc_layout" }
 dc_figma_import = { version = "0.38.2", path = "crates/dc_figma_import" }
 android_logger = "0.13.1"
 anyhow = "1.0"
+base64 = "0.21"
```
dipenpradhan: Are we still using this

crates/dc_figma_import/src/animation_override.rs
```diff
@@ -0,0 +1,496 @@
...
+use dc_bundle::animationspec::{
+    animation_override, animations, easing, repeat_type, stop_type, AnimationSpec, Animations,
+    BezierCurve, CustomKeyframe, CustomTimeline, Easing, KeyFrame, KeyFrameAnimation, RepeatType,
+    SmoothAnimation, StopType,
+};
+
+/// Converts from the JSON `AnimationOverrideJson` to the protobuf `AnimationOverride`.
+pub struct AnimationOverride(animationspec::AnimationOverride);
+
+impl Into<animationspec::AnimationOverride> for AnimationOverride {
```
anshu10sep: In Rust, it's idiomatic to implement `From<T> for U` rather than `Into<U> for T`. Implementing `From` automatically provides the blanket `Into` implementation (via the standard library), but not vice-versa.
Consider:
```rust
impl From<AnimationOverride> for animationspec::AnimationOverride {
    fn from(val: AnimationOverride) -> Self { val.0 }
}
```

crates/dc_figma_import/src/animation_override.rs
```diff
@@ -0,0 +1,496 @@
...
+            },
+        })
+    }
+}
+
+/// Converts from the JSON `CustomKeyframeJson` to the protobuf `CustomKeyframe`.
+impl From<CustomKeyframeJson> for CustomKeyframe {
+    fn from(json: CustomKeyframeJson) -> Self {
+        let mut typed_value = animationspec::CustomKeyframeValue::default();
+        if let Some(scalar) = json.value_json.as_f64() {
```
anshu10sep: Manually branching on the `serde_json::Value` type like this is a bit brittle. If the JSON structure for a keyframe value ever changes or expands, this will be hard to maintain. Consider defining a strongly-typed enum or struct for the expected JSON representation of a keyframe value (e.g., `CustomKeyframeValueJson`), and use `serde_json::from_value` for the entire structure to take advantage of `serde`'s automatic deserialization and error reporting.

crates/dc_figma_import/src/animation_override.rs
```diff
@@ -0,0 +1,496 @@
...
+                animationspec::RgbaValue {
+                    r: (color.r * 255.0) as u32,
+                    g: (color.g * 255.0) as u32,
+                    b: (color.b * 255.0) as u32,
+                    a: (color.a * 255.0) as u32,
+                    ..Default::default()
+                },
+            ));
+        } else {
+            println!("Failed to parse custom keyframe value JSON: {:?}", json.value_json);
```
anshu10sep: Avoid using `println!` for logging in library code, as it cannot be intercepted or disabled by the consumer. Please use the `log` crate (e.g., `log::warn!` or `log::error!`). The same applies to `transform_flexbox.rs` around line 2074.

crates/dc_figma_import/src/document.rs
```diff
@@ -698,6 +731,35 @@ impl Document {
                                 error!("ViewStyle is required.");
                                 MessageField::none()
                             };
+
+                            let anim_over = view_style
```
anshu10sep: If `anim_over` is sourced from `template_view` (via the `or_else` fallback), then the instance doesn't actually override the animation – it just inherits it. Writing it back to `override_view_style` in this case explicitly duplicates the template's value into the override delta. Is this intended? Usually, an override should only be set if it *differs* from the template to keep the payload small.

crates/dc_figma_import/src/document.rs
```diff
@@ -888,6 +950,29 @@ impl Document {
                             }
                         }
                     }
+
+                    if let Some(data) = vsw_data.get("squoosh") {
```
anshu10sep: The squoosh animation data is parsed here but only for logging purposes (the parsed result isn't assigned to anything). Later in `transform_flexbox.rs` (around line 2060), the exact same parsing logic is executed again to actually populate `parsed_plugin_override`. Consider removing this redundant parsing and logging from `document.rs` to keep it clean and avoid duplicate work.

crates/dc_squoosh_animation/src/lib.rs
```diff
@@ -0,0 +1,647 @@
...
+        Easing::Inherit
+    }
+}
+
+impl Easing {
+    pub fn apply(&self, t: f32) -> f32 {
+        match self {
+            Easing::Inherit => t,
+            Easing::Linear => t,
+            Easing::CubicBezier(_, _, _, _) => t, // FIXME: Implement cubic bezier solver
```
anshu10sep: A `FIXME` here for `CubicBezier`. Since this feature is being introduced in this PR, falling back to a linear interpolation `t` might cause noticeable jitter or unexpected motion when designers specifically choose a custom bezier curve in Figma. Should we implement a basic approximation for the beta release, or at least log a warning so it's clear why the animation looks linear?

crates/dc_squoosh_animation/src/lib.rs
```diff
@@ -0,0 +1,647 @@
...
+                    a[2] + (b[2] - a[2]) * t,
+                    a[3] + (b[3] - a[3]) * t,
+                ])
+            }
+            (KeyframeValue::Arc(a), KeyframeValue::Arc(b)) => KeyframeValue::Arc(ArcData {
+                starting_angle: a.starting_angle + (b.starting_angle - a.starting_angle) * t,
+                ending_angle: a.ending_angle + (b.ending_angle - a.ending_angle) * t,
+                inner_radius: a.inner_radius + (b.inner_radius - a.inner_radius) * t,
+            }),
+            (KeyframeValue::Gradient(a), KeyframeValue::Gradient(b)) => {
```
anshu10sep, line 164: If the number of gradient stops changes between keyframes, this returns the starting gradient, effectively snapping instantly instead of animating. It might be worth documenting this limitation or handling basic cases (e.g. interpolating matching stops and fading the rest) so it doesn't manifest as a visual glitch.

crates/dc_squoosh_animation/src/lib.rs
```diff
@@ -0,0 +1,647 @@
...
+            "cornerRadius" => Ok(AnimatableProperty::CornerRadius),
+            "topLeftRadius" => Ok(AnimatableProperty::TopLeftRadius),
+            "topRightRadius" => Ok(AnimatableProperty::TopRightRadius),
+            "bottomLeftRadius" => Ok(AnimatableProperty::BottomLeftRadius),
+            "bottomRightRadius" => Ok(AnimatableProperty::BottomRightRadius),
+            "arcData" => Ok(AnimatableProperty::ArcData),
+            "strokeWeight" => Ok(AnimatableProperty::StrokeWeight),
+            "rotation" => Ok(AnimatableProperty::Rotation),
+            _ => {
+                if s.starts_with("fills.") {
```
anshu10sep: This parsing logic can be simplified and made more idiomatic using slice patterns. For example:
```rust
if let [_, idx_str, type_str] = s.split('.').collect::<Vec<_>>().as_slice() {
    if let Ok(idx) = idx_str.parse::<usize>() {
        match *type_str {
            "solid" => return Ok(AnimatableProperty::FillSolid(idx)),
            // ...
```
This avoids manual indexing and length checks.

anshu10sep (commented):
Here is a detailed and thorough review of the PR. The goal is to make sure the changes are right, up to quality and builds a robust and scalable architecture moving forward. Also, ensure there is proper documentation done.

crates/dc_squoosh_animation/src/lib.rs
```diff
@@ -0,0 +1,647 @@
...
+                    }
+                }
+                Ok(AnimatableProperty::Other(s.to_string()))
+            }
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Default)]
+pub struct NodeTimelines {
```
anshu10sep, line 347: Hardcoding every animatable property as an `Option` adds significant boilerplate to the `insert` and `get` methods below this struct. Since `AnimatableProperty` implements `Hash` and `Eq`, you could simplify this entire struct to just:
`pub struct NodeTimelines(pub HashMap<AnimatableProperty, ParsedTimelineData>);`
This would make the code much more concise and easier to extend when adding new animatable properties in the future.

support-figma/dc_squoosh_designer/src/timeline/PlaybackController.ts
```diff
@@ -0,0 +1,620 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
anshu10sep: Disabling `any` for the entire file bypasses TypeScript's type safety. If there are specific places where the type is truly unknown, consider using `unknown` instead of `any`, or narrow the disable comment to just the lines that strictly require it to ensure a robust architecture moving forward.

Cargo.toml
kovmarci86: removed.

crates/dc_figma_import/src/animation_override.rs
kovmarci86: done.

crates/dc_figma_import/src/animation_override.rs
kovmarci86: done.

crates/dc_figma_import/src/animation_override.rs
kovmarci86: done.

crates/dc_squoosh_animation/src/lib.rs
kovmarci86: done.

support-figma/dc_squoosh_designer/src/timeline/PlaybackController.ts
kovmarci86: done.

crates/dc_figma_import/src/document.rs
kovmarci86: done.

crates/dc_squoosh_animation/src/lib.rs
kovmarci86: done.

crates/dc_squoosh_animation/src/lib.rs
kovmarci86: if there are different amount of gradien stops, we consider the 2 states not interpolatable. this is intentional.  It applies to mostly everything. the color stops, shape types, background types, background layers, etc.
there is no trivial way to animate between a different number of gradient stops. 
Updated the comment around this part of the code.

crates/dc_squoosh_animation/src/lib.rs
kovmarci86: fixed, but left a todo to implement a bezier curve editor in the plugin.

crates/dc_figma_import/src/document.rs
kovmarci86: done.

crates/dc_squoosh_animation/src/lib.rs
kovmarci86: Marking this resolved. I will get back to documenting the limitations when integrating this into Design Compose Squoosh

crates/dc_bundle/src/definition/view.rs
```diff
@@ -317,6 +318,19 @@ impl ViewStyle {
...
+            delta.node_style_mut().horizontal_sizing = other.node_style().horizontal_sizing;
+        }
+        if self.node_style().vertical_sizing != other.node_style().vertical_sizing {
+            delta.node_style_mut().vertical_sizing = other.node_style().vertical_sizing;
+        }
+        if self.node_style().scalable_data != other.node_style().scalable_data {
+            delta.node_style_mut().scalable_data = other.node_style().scalable_data.clone();
+        }
+        if self.node_style().animation_override != other.node_style().animation_override {
+            delta.node_style_mut().animation_override =
```
anshu10sep, line 331: When computing the delta style, cloning the entire `animation_override` could be somewhat expensive if the override contains many custom keyframes across numerous layers. While it's perfectly fine for now, if delta calculation happens frequently on the critical path, this could impact serialization or diffing performance. Is this delta calculation done often enough for that to be a concern?

support-figma/dc_squoosh_designer/rs/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+
+        deserializer.deserialize_any(RgbaVisitor)
+    }
+}
+
+impl Rgba {
+    pub fn from_hex(hex: &str) -> Self {
+        let hex = hex.trim_start_matches('#');
+        if hex.len() == 6 {
+            let r = u8::from_str_radix(&hex[0..2], 16).unwrap_or(0);
```
anshu10sep: If the hex parsing fails midway (e.g. `u8::from_str_radix` returns an error for the green channel), it silently falls back to 0. This means `from_hex` can produce an incorrect color (like `#FF0000` if `g` and `b` fail) instead of explicitly bubbling up an error or completely rejecting the malformed color string. In a design tool context, silent failure could lead to unexpected visual results. Returning a `Result<Self, ParseError>` and handling it gracefully higher up might be safer.

support-figma/dc_squoosh_designer/src/services/DataMapper.ts
```diff
@@ -0,0 +1,746 @@
...
+                        }
+                }
+                if (value === undefined) value = 0; // Fallback
+            }
+        } else if (value === undefined) {
+            // Node present but property undefined (e.g. fill removed).
+            // Default to 0 for numeric properties to ensure continuity (e.g. fade out).
+            if (
+                propName.endsWith("opacity") ||
+                propName.endsWith("width") ||
```
anshu10sep: Defaulting layout properties like `width` or `height` to 0 when undefined might cause the element to abruptly disappear or collapse during animation, rather than interpolating to its intrinsic layout size. It might be safer to omit the keyframe to allow standard layout interpolation, or try to fallback to the element's actual bounding box dimensions if possible.

anshu10sep (commented):
Taking another pass through the architecture and design here are a few more thoughts regarding edge cases and scalability to consider.

crates/dc_bundle/src/definition/view.rs
kovmarci86: The delta calculation (ViewStyle::diff or equivalent) happens exclusively during build-time/Figma-import time when the dc_figma_import utility converts raw Figma JSON onto clustered definitions and computes overrides. It never runs during application playback/interactive UI refreshes! Thus, aggressive cloning here adds strictly zero runtime frame-rate costs.

support-figma/dc_squoosh_designer/rs/src/lib.rs
kovmarci86: done.

support-figma/dc_squoosh_designer/src/services/DataMapper.ts
kovmarci86: done

anshu10sep (approved):
Thanks for the updates! The fixes for the layout width/height defaults and the hex color parsing errors look great and address the most critical issues. 

Since this is a beta/preview PR, the remaining structural suggestions and edge cases mentioned in the previous comments can be addressed in follow-up PRs as the plugin matures. Approving this!

dipenpradhan (approved):

dipenpradhan:
Looks good, thanks for addressing comments!

PR #2411, opened 2026-04-10
Fix #2403: Replace println!/eprintln! with Rust log API in library code

### LINKED ISSUE #2403: Use the rust log API instead of print/eprint

anshu10sep:
## Summary
Replaces all non-test `println!` and `eprintln!` calls in Rust library code with the standard `log` crate (`log::warn!`, `log::debug!`).

## Problem
Non-test Rust code used `println!`/`eprintln!` for diagnostic output instead of the standard `log` crate. These calls are invisible in Android logcat, making debugging in production builds difficult.

## Changes
- **`crates/dc_bundle/src/definition_file.rs`**: Replaced `eprintln!` with `log::warn!` for definition file warnings
- **`crates/dc_figma_import/src/document.rs`**: Replaced `println!` with `log::warn!` for missing interaction nodes
- **`crates/dc_figma_import/src/image_context.rs`**: Replaced `println!` with `log::warn!` for image fetch failures
- **`crates/dc_figma_import/src/transform_flexbox.rs`**: Replaced all `println!` with `log::warn!` for layout warnings

## Validation
- `cargo check` passes for all affected crates
- grep confirms 0 remaining non-test `println!`/`eprintln!` in library code
- All existing tests continue to pass

Fixes #2403

### DIFF
--- a/crates/dc_bundle/src/definition_file.rs
+++ b/crates/dc_bundle/src/definition_file.rs
@@ -16,6 +16,7 @@
 
 use crate::design_compose_definition::{DesignComposeDefinition, DesignComposeDefinitionHeader};
 use crate::Error;
+use log::warn;
 use protobuf::{CodedInputStream, Message};
 
 use std::fs::File;
@@ -49,7 +50,7 @@ pub fn decode_dcd_with_header(
 
     // Ensure the version of the document matches this version of automotive design compose.
     if header.dc_version != DesignComposeDefinitionHeader::current_version() {
-        println!(
+        warn!(
             "DesignComposeDefinition old version found. Expected {} Found: {}",
             DesignComposeDefinitionHeader::current_version(),
             header.dc_version
--- a/crates/dc_figma_import/src/document.rs
+++ b/crates/dc_figma_import/src/document.rs
@@ -30,7 +30,7 @@ use dc_bundle::variable::{Collection, Mode, Variable, VariableMap};
 use dc_bundle::view::view_data::{Container, View_data_type};
 use dc_bundle::view::{ComponentInfo, ComponentOverrides, View, ViewData};
 use dc_bundle::view_style::ViewStyle;
-use log::error;
+use log::{error, warn};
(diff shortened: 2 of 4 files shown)


dipenpradhan (approved):

crates/dc_bundle/src/definition_file.rs
```diff
@@ -58,7 +59,7 @@ pub fn decode_dcd_with_header(
 
     let dcd_length = cis.read_raw_varint32().map_err(|_| Error::DecodeError())?;
     let dcd_limit = cis.push_limit(dcd_length as u64).map_err(|_| Error::DecodeError())?;
-    println!("DCD length = {:?}", dcd_length);
+    debug!("DCD length = {:?}", dcd_length);
```
dipenpradhan: This is used on the console to print out information about the DCF file, we likely need this moved to the printing in main() of dcf_info

dipenpradhan (changes requested):

crates/dc_bundle/src/definition_file.rs
dipenpradhan: Check output of `./target/debug/dcf_info integration-tests/validation/src/main/assets/figma/AlignmentTestDoc_JIjE9oKQbq8ipi66ab5UaK.dcf`, which runs `main()` of https://github.com/google/automotive-design-compose/blob/main/crates/dc_figma_import/src/bin/dcf_info.rs

anshu10sep:
Addressed the review feedback — reverted the `DCD length` line back to `println!` since it's console output used by `dcf_info`. The version mismatch warning stays as `warn!` (appropriate for library-level warnings). Removed the unused `debug` import.

Updated and force-pushed. Thank you for the review!

crates/dc_bundle/src/definition_file.rs
anshu10sep: Updated

dipenpradhan (approved):

PR #2418, opened 2026-04-10
Fix #249: Add security check for Figma token file permissions

### LINKED ISSUE #249: Add checks to Gradle to ensure that the Figma Token file is not world-readable
As suggested by @kgiesing.

anshu10sep:
## Summary
Adds a security check that warns when the Figma access token file has overly permissive file permissions on Unix/macOS systems.

## Problem
The Figma access token file could be world-readable, exposing the token to other users on shared systems. There was no check or warning about this.

## Changes
- **`plugins/gradle-plugin/src/main/kotlin/com/android/designcompose/gradle/PluginExtension.kt`**:
  - Added `checkTokenFilePermissions()` function that runs when the token is read from a file
  - Uses `java.nio.file.Files.getPosixFilePermissions()` to check Unix permissions
  - Warns about world-readable, world-writable, and group-readable files
  - Provides remediation command: `chmod 600 <filepath>`
  - Skipped on Windows; fails gracefully on unsupported filesystems

## Behavior
| Permissions | Warning |
|-------------|---------|
| `-rw-------` (600) | None — secure ✅ |
| `-rw-r-----` (640) | Group-readable warning |
| `-rw-r--r--` (644) | World-readable warning |
| `-rw-rw-rw-` (666) | World-readable and writable warning |

> The check only runs when the token comes from a file. Environment variables and Gradle properties are not checked.

## Validation
- `./gradlew :plugins:gradle-plugin:compileKotlin`: Passes ✅
- `./gradlew :plugins:gradle-plugin:test`: All 4 existing tests pass ✅

Fixes #249

### DIFF
--- a/plugins/gradle-plugin/src/main/kotlin/com/android/designcompose/gradle/PluginExtension.kt
+++ b/plugins/gradle-plugin/src/main/kotlin/com/android/designcompose/gradle/PluginExtension.kt
@@ -17,6 +17,8 @@
 package com.android.designcompose.gradle
 
 import java.awt.SystemColor.text
+import java.nio.file.Files
+import java.nio.file.attribute.PosixFilePermission
 import org.apache.tools.ant.taskdefs.condition.Os
 import org.gradle.api.Project
 import org.gradle.api.file.Directory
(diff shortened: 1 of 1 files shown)


dipenpradhan (approved):

dipenpradhan:
This might be needed in the Rust code as well, where dc_figma_import reads the API key for fetching

PR #2431, opened 2026-04-10
chore: remove unclaimed Open VSX extension recommendation

sladewilson18:
Removes `arahata.linter-actionlint` from `.vscode/extensions.json` as the 
publisher namespace `arahata` is unclaimed on the Open VSX registry.

VS Code derivatives that resolve extensions against Open VSX (VSCodium, 
Gitpod, Theia-based IDEs) will attempt to install whatever is published 
under this ID on Open VSX. Since the namespace is unregistered, any party 
can claim it and publish arbitrary code under this exact extension ID.

Developers can still use actionlint linting via the CLI directly or through 
an alternative extension with an active maintainer on Open VSX.


sladewilson18:
Hi @codecov-commenter @dipenpradhan @anshu10sep, 

Can you please review this and close Asap. 

Regards

anshu10sep (approved):
Great catch! Thanks for identifying this. Removing this recommendation mitigates a real supply-chain risk for developers using VS Code derivatives that resolve against the Open VSX registry. Namespace squatting is a valid concern here.

While removing it is definitely the safest immediate action, we do lose the out-of-the-box recommendation for GitHub Actions linting for standard VS Code users. Have you come across a safer, actively maintained alternative extension for `actionlint` that is securely registered on both the VS Code Marketplace and Open VSX? If not, this removal is perfect as-is.

sladewilson18:
Hi @anshu10sep . 

I haven't found a verified equivalent registered on both registries under a trusted publisher. Removal is the safest path for now, maintainers can re-add when a secure alternative becomes available.

Regards,

dipenpradhan (approved):

sladewilson18:
Hi @dipenpradhan and @anshu10sep 

all required checks are passing, no conflicts with base, and both approvals are in. 

Could you please merge when you get a chance?

Also, please let me know if there’s anything else needed from my side before this can be closed out.


Thanks!​​​​​​​​​​​​​​​​

Issue #2435, opened 2026-04-21
Add crate.io publish workflow to GitHub Actions


dipenpradhan:
Add [crates.io](https://crates.io)publish workflow for:

- `dc_bundle`
- `dc_layout`
- `dc_figma_import`
- `dc_jni`

Set up `release.yml` to run `cargo publish` in addition to existing GitHub Release workflow

PR #2436, opened 2026-04-21
Add cargo publish steps to GitHub release workflow (Issue #2435)

dipenpradhan:
#2435 

Add a dedicated `publish-crates` job into the existing GitHub Release workflow `.github/workflows/release.yml`.

This job automates the publication of the following crates to crates.io (https://crates.io) whenever a new release tag (matching `v*`) is pushed:
* `dc_bundle`: The core definition crate.
* `dc_layout`: The layout engine (depends on dc_bundle).
* `dc_figma_import`: The Figma serialization toolkit (depends on dc_layout).
* `dc_jni`: The JNI library for Android (depends on all of the above).

- Crates are published sequentially in the order listed above to ensure that upstream dependencies are available on the registry before downstream crates are uploaded.
- Each cargo publish step includes a Bash retry loop (up to 5 attempts with a 15-second sleep) to allow for crates.io to index a newly published dependency.
- `CARGO_REGISTRY_TOKEN` environment variable is used for publishing, which pulls from a repository secret named `CRATES_IO_TOKEN`.
- The project's shared `./.github/actions/setup-build` composite action is reused to ensure a consistent Rust environment across all CI/CD jobs.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,6 +50,9 @@ jobs:
       - uses: ./.github/actions/build-figma-resource
         with:
             resource: extended-layout-plugin
+      - uses: ./.github/actions/build-figma-resource
+        with:
+            resource: dc_squoosh_designer
 
 
   build-maven-repo:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -175,3 +175,64 @@ jobs:
...
+        uses: ./.github/actions/setup-build
+        with:
+          setup-rust: true
+
+      - name: Publish dc_bundle
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
+        working-directory: crates/dc_bundle
+        run: |
+          for _ in 1 2 3 4 5; do
```
kovmarci86, line 206: Do we need to have retries or should the process just fail?

dipenpradhan: The retries are in case of delays in the dependency crates becoming available on crates.io.

kovmarci86: ah ok. i see now. thanks!

kovmarci86 (approved):

PR #2438, opened 2026-04-21
Fix #341: Preserve constraints for replacement content frames

### LINKED ISSUE #341: Constraints are not honored in frames that are used as replacement content
When using DesignCompose to customize a frame's children, the replaced children don't have the constraints set in their original frames. This can be seen in Validation's LayoutReplacementTest, which uses Figma doc https://www.figma.com/file/dwk2GF7RiNvlbbAKPjqldx/Layout-Replacement-Inheritance?type=design&node-id=0%3A1&mode=design&t=juFZrI9K70yYLY9l-1. The constraints are not preserved because we retrieve the replacement frames such as "#fill" using a separate NodeQuery::Name("#fill"). When we retrieve this frame from Figma, we have no parent data, so the code to compute layout in `transform_flexbox.rs` skips all the constraint stuff.

anshu10sep:
## Summary
Fix constraint preservation when replacement content frames (nodes without parent context) are imported from Figma, ensuring their layout constraints are not lost during the import pipeline.

## Changes
- **crates/dc_figma_import/src/transform_flexbox.rs**: Preserve constraint types for replacement content; added regression tests.

## Testing
- 2 new integration tests added
- All 36 existing tests pass

Fixes #341


crates/dc_figma_import/src/transform_flexbox.rs
```diff
@@ -2219,6 +2365,156 @@ mod tests {
             FlexWrap::FLEX_WRAP_WRAP.into()
         );
     }
+
+    /// Test that a node with Left-Right + Top-Bottom constraints preserves
+    /// its constraint anchoring even when fetched without a parent context
+    /// (i.e. as a replacement content node). Bug #341.
```
dipenpradhan:
```suggestion
    /// (i.e. as a replacement content node).
```

crates/dc_figma_import/src/transform_flexbox.rs
```diff
@@ -599,6 +599,152 @@ fn compute_layout(
                     }
                 }
             }
+            // Handle nodes with constraints but no parent context (e.g. replacement
+            // content fetched as top-level nodes via NodeQuery::Name). We preserve the
+            // constraint type with zero-offset margins so the layout engine can apply
+            // them when this node is placed into its runtime parent. (Bug #341)
```
dipenpradhan:
```suggestion
            // them when this node is placed into its runtime parent.
```

dipenpradhan (changes requested):

crates/dc_figma_import/src/transform_flexbox.rs
anshu10sep: Done. Removed bug reference.

crates/dc_figma_import/src/transform_flexbox.rs
anshu10sep: Done. Removed bug reference.

dipenpradhan (approved):

PR #2440, opened 2026-04-21
Fix #1279: Squoosh rotated composable placement with placeRelativeWithLayer

### LINKED ISSUE #1279: Handle rotated component replacement nodes in squoosh
Currently, squoosh is not propagating rotation (or likely other info like opacity, etc -- but we don't yet have tests of this) to replaced children.

anshu10sep:
## Summary
Fix placement of replacement content composables that have rotation transforms. The squoosh layout was using `placeRelative` which doesn't honor rotation, causing misalignment.

## Changes
- **designcompose/.../squoosh/SquooshRoot.kt**: Use `placeRelativeWithLayer(rotationZ = ...)` for replacement composables via matrix `decompose()`.

## Testing

Fixes #1279


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -1259,7 +1253,19 @@ private fun MeasureScope.squooshLayout(
                     x -= scrollOffset.value.x.roundToInt()
                     y -= scrollOffset.value.y.roundToInt()
                 }
-                placeable.placeRelative(x, y)
+
+                // Extract rotation from the node's transform matrix. If the node
+                // has a non-zero rotation, use placeRelativeWithLayer to apply it.
+                // This fixes Bug #1279 where rotated replacement content nodes
+                // were placed without their rotation being applied.
```
dipenpradhan:
```suggestion
                // This ensures that rotated replacement content nodes are
                // always placed with their rotation being applied.
```

dipenpradhan (changes requested):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
anshu10sep: Done. Updated the comment as suggested.

dipenpradhan (approved):

PR #2441, opened 2026-04-21
Fix #175: Verify absolute position in AutoLayout with regression tests

### LINKED ISSUE #175: Figma Absolute Position within AutoLayout is ignored and instead positioned in sequence of layering
render_design_compose
absolute_position

anshu10sep:
## Summary
Verify that absolute positioning within AutoLayout (flex) containers is correctly supported end-to-end. The Taffy layout engine and Figma import pipeline already handle this correctly — this PR adds comprehensive regression tests to prevent future regressions.

## Changes
- **crates/dc_layout/src/layout_manager.rs**: Added `test_absolute_position_in_autolayout` proving absolute children use inset offsets
- **crates/dc_figma_import/src/transform_flexbox.rs**: Added `test_absolute_position_in_autolayout_import` verifying correct import

## Testing
- 45/45 layout tests pass
- 37/37 import tests pass

Fixes #175


.gemini/all_issues.json
```diff
@@ -0,0 +1,2173 @@
+[
```
dipenpradhan: This file seems like an accidental commit?

.gemini/analysis.json
```diff
@@ -0,0 +1,2259 @@
+[
```
dipenpradhan: Same as other .gemini/ file - looks like an internal / local file - remove?

dipenpradhan (changes requested):

.gemini/all_issues.json
anshu10sep: Done. Removed both .gemini/ files — they were accidentally committed.

.gemini/analysis.json
anshu10sep: Done. Removed.

crates/dc_figma_import/src/transform_flexbox.rs
```diff
@@ -2478,3 +2343,110 @@ fn test_animation_override_plugin_data() {
 
     assert!(view.style.unwrap().node_style.unwrap().animation_override.is_some());
 }
+
+/// Bug #175: Verify that a child frame with layout_positioning: ABSOLUTE inside
```
dipenpradhan:
```suggestion
/// Verify that a child frame with layout_positioning: ABSOLUTE inside
```

crates/dc_layout/src/layout_manager.rs
```diff
@@ -1178,4 +1178,79 @@ mod tests {
         assert_eq!(lm.get_node_layout(3).unwrap().left, 0.0);
         assert_eq!(lm.get_node_layout(2).unwrap().left, 80.0);
     }
+
+    /// Bug #175: Absolute-positioned child inside a flex (AutoLayout) parent should be
```
dipenpradhan:
```suggestion
    /// Absolute-positioned child inside a flex (AutoLayout) parent should be
```

dipenpradhan (changes requested):

crates/dc_figma_import/src/transform_flexbox.rs
anshu10sep: Done. Removed bug reference.

crates/dc_layout/src/layout_manager.rs
anshu10sep: Done. Removed bug reference.

dipenpradhan (approved):

PR #2443, opened 2026-04-21
Fix #572: Text brush customization invalidation in Squoosh

### LINKED ISSUE #572: Ensure updates of text brush function customization cause invalidations
Reading state from the brush customization is enough to trigger a re-render (but not re-compose) of frames and vectors. Unfortunately for text, we can't set the brush during the render phase (and can't really customize the rendering of BasicText at draw time), so we run the brush customization function during composition. This should result in re-composition for every animation frame, but actually that doesn't happen, and instead the text fill doesn't update.

The work in squoosh provides one way to address this -- we won't use BasicText, so we can do whatever we want during rendering (and provide an empty tree of composables with semantics attached for accessibility and input -- I haven't tackled text input in squoosh and that might be quite a bit more complex).

Otherwise we need to do more debugging into why the DesignText composable doesn't seem to be getting invalidated when some dependent state changes.

anshu10sep:
## Summary
Fix text brush customizations (from `setBrush`/`setBrushFunction`) not being applied when the underlying Compose State changes.

The `TextMeasureCache` was only comparing `text` and `customStyle` when deciding whether to reuse a cached entry, completely ignoring brush customizations. Animated brushes driven by State were never invalidated.

## Changes
- **designcompose/.../squoosh/SquooshText.kt**:
  - Moved `customBrush` evaluation before the cache check
  - Added `customBrush` field to `TextMeasureCache.Entry`
  - Added `customBrush` comparison to cache validity check

## Testing

Fixes #572

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
+++ b/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
@@ -77,6 +77,7 @@ internal class TextMeasureCache {
         val data: TextMeasureData,
         val text: String?,
         val customStyle: TextStyle?,
+        val customBrush: androidx.compose.ui.graphics.Brush?,
         val annotatedText: String,
         val textStyle: TextStyle,
         // XXX: Do we need to use the annotated string? This impl might break localization.
@@ -148,11 +149,16 @@ internal fun squooshComputeTextInfo(
             )
         }
 
+    // Evaluate the custom brush BEFORE the cache check so that
+    // brush function changes (e.g., from State updates) invalidate the cache.
+    val customBrush = customizations.getBrush(v.name)
+
     val cachedText = textMeasureCache.get(layoutId)
     if (
         cachedText != null &&
             cachedText.text == customizedText &&
-            cachedText.customStyle == customTextStyle
+            cachedText.customStyle == customTextStyle &&
+            cachedText.customBrush == customBrush
     ) {
         textHash.add(cachedText.annotatedText)
         textMeasureCache.put(layoutId, cachedText)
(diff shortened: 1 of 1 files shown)


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
```diff
@@ -274,7 +280,7 @@ internal fun squooshComputeTextInfo(
     val letterSpacing =
         customTextStyle?.letterSpacing
             ?: (v.style.nodeStyle.takeIf { it.hasLetterSpacing() }?.letterSpacing ?: 0f).sp
-    val customBrush = customizations.getBrush(v.name)
+    // customBrush was already evaluated above, before the cache check (Bug #572)
```
dipenpradhan:
```suggestion
```

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
```diff
@@ -148,11 +149,16 @@ internal fun squooshComputeTextInfo(
             )
         }
 
+    // Bug #572: Evaluate the custom brush BEFORE the cache check so that
```
dipenpradhan:
```suggestion
    // Evaluate the custom brush BEFORE the cache check so that
```

dipenpradhan (changes requested):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
anshu10sep: Done. Removed the redundant comment.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshText.kt
anshu10sep: Done. Removed the bug reference prefix.

dipenpradhan (approved):

PR #2445, opened 2026-04-21
Fix #154: Design switcher visibility behind overlays

### LINKED ISSUE #154: Design switcher not visible
The following instructions cannot be followed right now:

```
Next, switch the app to use your copy of the Tutorial file by clicking the dropdown arrow in the upper right. This will open the Design Switcher.

Click the Change button to switch document IDs, and enter the ID of your copy of the Tutorial Figma File. Click Load and the app will start fetching your file (it'll take about a minute)
```

Somehow, the design switcher is not shown, unless you replace the id with your Figma file id beforehand:
```
@DesignDoc(id = "3z4xExq0INrL9vxPhj9tl7", version = "0.8")
```

See [tutorial/MainActivity.kt#L81](https://github.com/google/automotive-design-compose/blob/main/reference-apps/tutorial/app/src/main/java/com/android/designcompose/tutorial/MainActivity.kt#L81)

anshu10sep:
## Summary
Fix the design switcher being hidden behind overlay composables (modals, bottom sheets, etc.).

## Root Cause
`designSwitcher()` was rendered BEFORE the overlay rendering loop. Overlays with `Modifier.fillMaxSize()` were drawn on top of the switcher, making it completely invisible and inaccessible.

## Changes
- **designcompose/.../squoosh/SquooshRoot.kt**: Move `designSwitcher()` to AFTER the overlay rendering loop so it's always the topmost z-order element.

## Testing

Fixes #154

### DIFF
--- a/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
+++ b/designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
@@ -886,7 +886,6 @@ fun SquooshRoot(
                 }
             },
         )
-        designSwitcher()
 
         for (overlay in composableList.overlayNodes) {
             val contentAlignment: Alignment =
@@ -932,6 +931,10 @@ fun SquooshRoot(
                 )
             }
         }
+
+        // Render design switcher AFTER overlays so it's always
+        // the topmost z-order element and remains visible.
+        designSwitcher()
     }
 }
 


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -932,6 +933,9 @@ fun SquooshRoot(
                 )
             }
         }
+
+        // Bug #154: Render design switcher AFTER overlays so it's always on top
```
dipenpradhan:
```suggestion
        // Render design switcher AFTER overlays so it's always
        // the topmost z-order element and remains visible.
```

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -886,7 +886,8 @@ fun SquooshRoot(
                 }
             },
         )
-        designSwitcher()
+        // Bug #154: Design switcher is rendered after overlays (below) to ensure
+        // it's always the topmost z-order element and remains visible.
```
dipenpradhan:
```suggestion

```

dipenpradhan (changes requested):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
anshu10sep: Done. Updated the comment as suggested.

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
anshu10sep: Done. Removed the placeholder comment.

dipenpradhan (approved):

dipenpradhan (approved):

PR #2446, opened 2026-04-22
Update DesignCompose version to 0.39.0

dipenpradhan:
This commit bumps the version across all Gradle (`gradle/libs.versions.toml`) and Cargo configuration files to `0.39.0` (`0.39.0-SNAPSHOT` in Gradle). It also unifies the `dc_squoosh_animation` and `dc_squoosh_parser` crates under the workspace dependencies.

Changes and issues addressed since 0.38.4:
- Enhance LiveUpdate feature (Issues #2282, #2379)
- Add cargo publish steps to GitHub release workflow (Issues #2435, #2436)
- Squoosh Designer Plugin: Figma plugin for orchestrating layout animations (Issue #2402)
- Replace println!/eprintln! with Rust log API in library code (Issues #2403, #2411)
- Update dc_figma_import README to reflect current architecture (Issues #426, #2427)
- Add comprehensive unit tests for mergeStyles function (Issues #1963, #2416)
- Avoid double key events during animated transitions (Issues #1514, #2417)
- Clean up unwrap() calls in dc_jni crate (Issues #670, #2415)
- Security: Update rustls-webpki 0.103.10 -> 0.103.12 (Issue #2434)
- Document supported Android SDK levels (Issues #163, #2429)
- Restore Modifier customization support in squoosh renderer (Issues #2292, #2426)
- Handle unknown Figma features in deserialization (Issues #2305, #2414)
- Add adaptive polling with exponential backoff for LiveUpdate (Issues #2282, #2423)
- Add comprehensive Rust layout unit tests (Issues #405, #2413)
- Update JNI dependency to 0.22.4 and refactor dc_jni (Issue #2432)
- Enable CodeQL security scanning (Issues #926, #2412)
- Prevent Design Switcher from briefly showing 'Offline' (Issues #142, #2421)
- Stroke gradients account for stroke thickness (Issues #1549, #2420)
- Add performance logging for high-cost fetch operations (Issues #281, #2419)
- Remove unclaimed Open VSX extension recommendation (Issue #2431)
- Add assembleRelease to CI to catch minified build issues (Issues #435, #2425)
- Deprecate and remove publish_designcompose.sh script (Issue #2433)
- Clarify visibility documentation — parameter name is flexible (Issues #660, #2428)
- Add JDK installation instructions to Getting Started docs (Issues #707, #2424)
- Add security check for Figma token file permissions (Issues #249, #2418)
- Fix whilePressing interaction getting stuck on multi-touch drag (Issue #2422)
[...]

PR #2442, opened 2026-04-21
Fix #279: Constrain DesignDoc reported size to parent constraints

anshu10sep:
## Summary
Fix DesignDoc expanding to full screen size by constraining the reported layout dimensions to the parent's constraint bounds.

`squooshLayout()` was reporting the raw computed Figma design size to Compose without constraining it. When placed inside a Column or other container, it consumed all available space and prevented sibling composables from being visible.

## Changes
- **designcompose/.../squoosh/SquooshRoot.kt**: Pass constraints to `squooshLayout()` and use `coerceIn()` to clamp reported width/height to the parent's min/max constraint bounds.

## Testing
- Verified Kotlin compilation succeeds (BUILD SUCCESSFUL)

Fixes #279

Issue #2328, opened 2025-11-13
Variables with included Opacity are not rendered properly


Elektrobit-HendrikCammann:
When having a variable defined with an opacity like
<img width="605" height="40" alt="Image" src="https://github.com/user-attachments/assets/ef89e0a2-5371-4706-af81-19e25d7c38a3" />
it is still rendered with 100% opacity.

**Version:** 0.38.0
**FigmaAccount:** Organization (i know that Figma API access to Variables only comes with Enterprise, but I think the variables itself still should be rendered correctly)

anshu10sep:
## ✅ Verified Fixed

**Fix commit:** [`b7aa42a7`](https://github.com/google/automotive-design-compose/commit/b7aa42a7) — Fix Variables with included Opacity rendering (Issue #2328)

**Validation:**
- Figma API confirmed HVAC Components file has 152 color variables with alpha \< 1.0 (e.g., `State Layers/Primary/Opacity-16` at alpha=0.16, `Surfaces/Surface Tint 5%` at alpha=0.05)
- Fix ensures opacity from variable-bound colors is preserved during rendering
- App deployed on emulator (API 34) with Figma token set — live update working
- All unit tests pass

This issue should be closed.

PR #2463, opened 2026-05-16
Create 0.39.2-RC01

### LINKED ISSUE #154: Design switcher not visible

### LINKED ISSUE #175: Figma Absolute Position within AutoLayout is ignored and instead positioned in sequence of layering

### LINKED ISSUE #244: Fix rendering of rotated nodes
When a node in Figma is rotated and offset from its parent (meaning it has an x and/or y offset from its parent when not rotated), we often render the node incorrectly. Something incorrect happens in transform_flexbox.rs when setting the transform matrix.

In the stroke validation test https://www.figma.com/file/Z3ucY0wMAbIwZIa6mLEWIK/stroke-stuff?node-id=0%3A1&t=0Dq3abJhR4zpH060-1, the node from the Battleship doc breaks because the rect is rotated 90 degrees, offset from its parent, and has its constraints set to left-right + top-bottom. This causes the width/height of the frame to be calculated incorrectly.

A possible fix may be to undo any rotation set in the Figma file so that we get the x/y offset without any rotation, then set the transform matrix with the rotation in it. We may need to also add a translation to the rotation so that we rotate about the center, not the top left.

### LINKED ISSUE #279: DesignDoc expands to full screen size even if not required
The DesignDoc Composable is expanding to the full size of the screen, preventing any other composables from being placed alongside it.

Reproduction steps:
- Checkout branch `wb/froeht/offset-issue`
- Open Helloworld's MainActivity (`reference-apps/helloworld/src/main/java/com/android/designcompose/testapp/helloworld/MainActivity.kt`)
- Open the Compose preview. The ColumnExample preview will show the intended result
image
- Launch helloworld on an emulator The emulator will show the actual result
image
  - Note that the slider control is visible at the bottom of the screen.
- Launch the ColumnExample composablle directly on the emulator  (click the button next to line number 62 
image
)
  - The emulator will show the intended result.

### LINKED ISSUE #291: Progress markers are misaligned in instances
Setting up a progress marker in a component and then using an instance of the component in your app will cause the marker to be rendered with an incorrect offset. It seems that the offset comes from the offset of the marker in the component, so if you use the dials & gauges plugin to set the value of the marker to 0, it renders correctly.

### LINKED ISSUE #341: Constraints are not honored in frames that are used as replacement content

### LINKED ISSUE #436: Android Studio won't build/launch Tutorial's Release variant
Need to confirm details, but it seems related to https://b.corp.google.com/issues/111118208

Seems like Android Studio doesn't want to create unsigned release builds of apps by default (or more specifically, unsigned builds of apps that aren't debuggable or profileable).

I don't think that we should just assign the debug key that we're using for Validation to it though, Tutorial is a reference app so we shouldn't providing reference code that uses an unprotected key to sign a release variant.

So what I'm thinking is I'll first standardize the key signing configuration in build-logic. Then, for all apps, I'll disable the default "Release" variant and replace it with a "TestRelease" variant that is signed with the same key.

This should be done with #435 to make sure everything works correctly.

### LINKED ISSUE #572: Ensure updates of text brush function customization cause invalidations

### LINKED ISSUE #1279: Handle rotated component replacement nodes in squoosh

### LINKED ISSUE #2459: Optimize m2repo artifact size
- https://github.com/google/automotive-design-compose/releases/download/v0.39.1/designcompose_m2repo-v0.39.1.zip is ~77MB.
- The artifact contains a large .so file for dc jni inside test-native.jar
- This should be optimized for release (maven) builds of the designcompose lib

dipenpradhan:
## Changes since v0.39.1

* Update DesignCompose version to 0.39.2 RC01 (#2461)
* Fix Variables with included Opacity rendering (Issue #2328) (#2400)
* Fix Dynamic Discovery Logic & Wire Up Incomplete Document Subsets (#2452)
* Update copyright year to 2026 (#2457)
* Fix gradient and fill opacity rendering with variables (#2460)
* Optimize native binary sizes, fix #2459 (#2458)
* Fix LayoutAlignItems::Baseline mapping in Figma import (#2455)
* Add the missing metadata to Cargo.tomls (#2449)
* Update Masks golden screenshot for constrained layout behavior (#2448)
* Fix #279: Constrain DesignDoc reported size to parent constraints (#2442)
* Fix #341: Preserve constraints for replacement content frames (#2438)
* Fix #1279: Squoosh rotated composable placement with placeRelativeWithLayer (#2440)
* Fix #175: Verify absolute position in AutoLayout with regression tests (#2441)
* Fix #154: Design switcher visibility behind overlays (#2445)
* Fix #572: Text brush customization invalidation in Squoosh (#2443)
* Update DesignCompose version to 0.39.0 (#2446)
* Fix #244: Rotated node margin calculation uses pre-rotation size (#2439)
* Fix #291: Progress marker alignment in nested instances (#2444)
* Fix #436: Tutorial release build signing configuration (#2437)


dipenpradhan:
Closing since we have conflicts in merging `main` > `stable/`. Since `main` is read-only, we cannot fix these conflicts on `main`, created #2464 to fix

PR #2474, opened 2026-06-11
Support Ripple touch feedback in Squoosh

Gurupreet:
Applies LocalIndication (which provides standard material ripples) to interactive components in Squoosh.

Changes:
- SquooshRoot: Imported LocalIndication and indication. Applied the indication modifier mapped to interactionSource for all components that have supported press/click interactions.


designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
```diff
@@ -820,15 +822,17 @@ fun SquooshRoot(
...
-                                    doc,
-                                    interactionState,
-                                    interactionScope,
-                                    customizationContext,
-                                    child,
-                                    placeholderIsPressed,
-                                    interactionSource,
-                                )
+                                composableChildModifier
+                                    .indication(interactionSource, LocalIndication.current)
```
anshu10sep, line 830: What does this line do? Do we need updated tests?

anshu10sep (approved):

designcompose/src/main/java/com/android/designcompose/squoosh/SquooshRoot.kt
Gurupreet: This is the new API Compose has added that will show default ripple indication on interaction. 

For tests: I don't think we have any tests in project that verify interaction/pressing/long pressing.  Just validation tests are there on buttons.

PR #2476, opened 2026-06-12
Fix DesignCompose exporter to explicitly serialize 1.0/0.0 opacity

Gurupreet:
### Summary
Fixes the exporter in `dc_figma_import` to explicitly set `1.0` opacity values instead of omitting them. 

### Details
Previously, the DesignCompose Figma plugin exporter omitted the `opacity` field (setting it to `None` in the protobuf) when its value was `1.0`. Because proto3 defaults missing float fields to `0.0` in primitive deserializers, this created ambiguity. To work around it, deserializers (such as Squoosh Rust and Kotlin) had to treat `0.0` as `1.0`, which meant that any layers intentionally set to `0.0` (fully transparent) in Figma were incorrectly rendered as fully opaque.

This change:
1. Updates frame and text converters in `crates/dc_figma_import/src/transform_flexbox.rs` to always set `style.node_style_mut().opacity = Some(node.opacity);`.
2. Adds a new Rust unit test `test_opacity_serialization` to verify that omitted (defaults to 1.0), explicitly `0.0`, and explicitly `0.5` opacities are all preserved and serialized correctly.
3. Preserves backward compatibility: deserializers will still fall back to `1.0` if they encounter older `.dcf` files where opacity is missing (`None`).

### Validation
- Ran Rust Cargo unit tests (`cargo test` in `crates/dc_figma_import`): **Passed**
- Ran Android JVM unit tests (`./gradlew :designcompose:testDebugUnitTest`): **Passed**
- Ran Roborazzi screenshot tests: **Passed**

### DIFF
--- a/crates/dc_figma_import/src/transform_flexbox.rs
+++ b/crates/dc_figma_import/src/transform_flexbox.rs
@@ -1474,7 +1474,7 @@ fn visit_node(
     // Blend mode is common to all elements.
     style.node_style_mut().blend_mode = convert_blend_mode(node.blend_mode).into();
 
-    style.node_style_mut().opacity = if node.opacity < 1.0 { Some(node.opacity) } else { None };
+    style.node_style_mut().opacity = Some(node.opacity);
 
     style.node_style_mut().stroke.as_mut().map(|stroke| {
         for visible_stroke in node.strokes.iter().filter(|paint| paint.visible) {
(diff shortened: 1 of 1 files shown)


Gurupreet:
@lyceel This should fix the ambiguous proto3 issue for opacity. please take a look 👍🏽

lyceel (approved):
LGTM

PR #2478, opened 2026-06-26
Fix component variant style resolution and relative positioning merge…

Gurupreet:
… in Squoosh

- Always call mergeStylesWithVariant when defaultVariantView is not null, even when overrideStyle is null, passing ViewStyle.getDefaultInstance() as a fallback. This ensures the component instance's relative layout constraints (like POSITION_TYPE_RELATIVE) are correctly merged and preserved on top of the variant's absolute styling constraints.
- Updated ViewStyleUtils.kt to properly copy overridden property values from instLayoutStyle when properties are marked as overridden.
- Added a validation test screen 'ClusterTest' and InstrumentedTest validation suite to verify correct variant layout behavior.


designcompose/src/main/java/com/android/designcompose/utils/ViewStyleUtils.kt
```diff
@@ -92,32 +92,47 @@ internal fun mergeStylesWithVariant(
...
                     if (!isFlexGrowOverridden) flexGrow = base.layoutStyle.flexGrow
+                    else flexGrow = instLayoutStyle.flexGrow
                     if (!isFlexShrinkOverridden) flexShrink = base.layoutStyle.flexShrink
+                    else flexShrink = instLayoutStyle.flexShrink
                     if (!isFlexBasisOverridden) flexBasis = base.layoutStyle.flexBasis
+                    else flexBasis = instLayoutStyle.flexBasis
                     if (!isMarginOverridden) margin = base.layoutStyle.margin
+                    else margin = instLayoutStyle.margin
                     if (!isPaddingOverridden) padding = base.layoutStyle.padding
+                    else padding = instLayoutStyle.padding
```
anshu10sep: Do we really need this code to be with so many if/else statements?

anshu10sep (approved):

deledzis (approved):

PR #2480, opened 2026-07-03
Add HAR Stage rendering validation test for different variants. 

Gurupreet:


integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/ClusterHARTest.kt
```diff
@@ -0,0 +1,339 @@
+/*
+ * Copyright 2024 The Android Open Source Project
```
anshu10sep: Year ;)

anshu10sep (approved):

integration-tests/validation/src/main/java/com/android/designcompose/testapp/validation/examples/ClusterHARTest.kt
Gurupreet: ohh

PR #2486, opened 2026-07-21
Feat: Support Custom Animation Matrix and Transition Schemas in DCF generation

deledzis:
### Overview
This PR implements the DesignCompose / Figma plugin compiler component for the **SDV Custom UI Component Animations** feature. It extends the `.dcf` node schema to support explicit, multi-variant `AnimationMatrix` bundling without violating zero-allocation bounds on the runtime side.

### Key Changes

1. **Schema Enhancements (dc_figma_import):**
- Added full compiler support for custom `AnimationMatrix` and `TransitionSpec` structures.
- Node schemas now explicitly capture `source_variant (origin)`, `target_variant`, and `animation_name` custom transition metadata extracted from the Figma Design plugin.
2. **Proto Submodule Bump:**
Update `designcompose-protos` submodule reference to ingest the updated `AnimationTransition` array structures.
3. **Rigorous Validation & Safety:**
- Bound strict structural validation algorithms in `animation_spec_schema.rs` to guarantee data integrity before serializing Custom Keyframes to `.dcf` files.

Requires: google/automotive-design-compose-protos#9

### DIFF
--- a/support-figma/dc_squoosh_designer/DATA_FORMAT.md
+++ b/support-figma/dc_squoosh_designer/DATA_FORMAT.md
@@ -1,4 +1,4 @@
-# Data Format
+# Data Format (Option A Transition Matrix Schema)
 
 This document describes the data structures used by the DC Squoosh Designer plugin to store animation specifications within Figma.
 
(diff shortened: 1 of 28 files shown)


support-figma/dc_squoosh_designer/DATA_FORMAT.md
```diff
@@ -1,62 +1,83 @@
...
 
 This document describes the data structures used by the DC Squoosh Designer plugin to store animation specifications within Figma.
 
 ## Storage Mechanism
 
 The plugin stores data in the `sharedPluginData` of Figma components (variants).
 *   **Namespace:** `designcompose`
-*   **Key:** `animations`
-*   **Value:** A stringified JSON object.
+*   **Key:** `squoosh`
```
kovmarci86, line 9: not important: is this intentional? we may lose existing access to existing animations. But there are barely any available yet so it is not a real problem.

kovmarci86 (approved):
Thank you!
Change is LGTM with an open question for the key used to save squoosh data in the JSON. Also, there is a test failure reported on this change , which needs to be fixed.

support-figma/dc_squoosh_designer/DATA_FORMAT.md
deledzis: Thank you for the review! The rename is intentional here, mainly because on this early stage we can leave the backwards compatibility out of the question. To prevent the parser from attempting to deserialize legacy payloads and throwing runtime exceptions, we moved the data storage to a new namespace key. Any existing files will need to be re-saved/exported using the new version of the plugin.

kovmarci86: Ack. Resolving this.

PR #2491, opened 2026-09-08
cargo: update h2 to 0.4.19 to resolve RUSTSEC-2026-0258

deledzis:
Bumps `h2` from `0.4.13` to `0.4.19` to resolve RustSec security advisory `RUSTSEC-2026-0258` (unbounded empty DATA frames) and unblock the `cargo-deny (advisories)` CI check.

Test: cargo deny check && cargo test --workspace

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,7 +71,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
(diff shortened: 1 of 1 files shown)


Gurupreet (approved):

Gurupreet:
Shouldn't this be auto updated via cargo changes? Or do we have to change lock file?

deledzis:
> Shouldn't this be auto updated via cargo changes? Or do we have to change lock file?

h2 is a transitive dependency, it's pulled in by reqwest under the hood, but h2 is not declared anywhere in Cargo.toml.
So we have to change the lock file here, yes

